PR #45, opened 2021-06-21
Implement span exporting limits

mateuszrzeszutek:


splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -143,7 +144,13 @@ SpanExporter buildExporter(ConnectionUtil connectionUtil) {
             // we'll do our best to hang on to the spans with the wrapping BufferingExporter.
             ZipkinSpanExporter.baseLogger.setLevel(Level.SEVERE);
         }
-        return new BufferingExporter(connectionUtil, ZipkinSpanExporter.builder().setEndpoint(endpoint).build());
+        ZipkinSpanExporter zipkinSpanExporter = ZipkinSpanExporter.builder().setEndpoint(endpoint).build();
+        SpanExporter throttlingExporter = ThrottlingExporter.newBuilder(zipkinSpanExporter)
+                .categorizeByAttribute(SplunkRum.COMPONENT_KEY)
+                .maxSpansInWindow(100)
+                .windowSize(Duration.ofSeconds(30))
```
mateuszrzeszutek, line 151: @jkwatson should I add these to `Config`? I feel kind of bad hard-coding them here...

jkwatson: @johnbley is this something iOS has made configurable?

jkwatson:
What do you think about adding a button to the demo app which will exercise this functionality?

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
johnbley: Not currently; I don't expect it would be necessary.

mateuszrzeszutek:
> What do you think about adding a button to the demo app which will exercise this functionality?

Good point - done. I added a "spam button" to the second fragment

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
jkwatson: let's leave them out for now. If someone needs it, we can always add it later.

jkwatson (approved):
Thanks! I think this is slightly more sophisticated than what's in iOS right now, but I think that's totally fine.

PR #47, opened 2021-06-22
make sure we can handle a null component when throttling.

jkwatson:
I noticed that we were handling a "null" `component` value, but it seemed like a good idea to have a test for it so we know we haven't broken it if we decide to change our map implementation to one that doesn't support null keys.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ThrottlingExporter.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ThrottlingExporter.java
@@ -39,6 +39,7 @@ class ThrottlingExporter implements SpanExporter {
     private final long windowSizeInNanos;
     private final int maxSpansInWindow;
     // note: no need to make this thread-safe since it will only ever be called from the BatchSpanProcessor worker thread.
+    // the implementation here needs to support null keys, or we'd need to use a default component value.
     private final Map<String, Window> categoryToWindow = new HashMap<>();
 
     private ThrottlingExporter(Builder builder) {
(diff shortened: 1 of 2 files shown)


breedx-splk (approved):
LGTM!

splunk-otel-android/src/test/java/com/splunk/rum/ThrottlingExporterTest.java
```diff
@@ -130,7 +135,11 @@ public void shouldCountDifferentComponentsSeparately() {
                 spanData.get(2),
                 spanData.get(3),
                 // idx=4 will be skipped because it's the 3rd component=ui span in the last 15 secs
-                spanData.get(5)
+                spanData.get(5),
+                spanData.get(6),
+                spanData.get(7),
+                // idx = 8 will be skipped because it's the 4th no-component span in the 2-span, 15s window
```
jkwatson, line 141:
```suggestion
                // idx = 8 will be skipped because it's the 3rd no-component span in the 2-span, 15s window
```

splunk-otel-android/src/main/java/com/splunk/rum/ThrottlingExporter.java
```diff
@@ -89,7 +90,12 @@ boolean aboveLimit(SpanData spanData) {
 
             // remove oldest entries until the window shrinks to the configured size
             while (true) {
-                long first = timestamps.peekFirst();
+                Long first = timestamps.peekFirst();
+                //this shouldn't happen, due to the single-threaded nature of things here, but
+                //just to be on the safe side, don't blow up if something has cleared out the window.
```
mateuszrzeszutek, line 95: 👍

splunk-otel-android/src/test/java/com/splunk/rum/ThrottlingExporterTest.java
```diff
@@ -200,8 +209,12 @@ public void shouldDelegateShutdownCall() {
     }
 
     private static SpanData span(String component, Instant endTime) {
+        return span(component, endTime, "");
+    }
+
+    private static SpanData span(String component, Instant endTime, String additionalName) {
         return TestSpanData.builder()
-                .setName("test")
+                .setName("test" + additionalName)
```
mateuszrzeszutek, line 217: Is that `additionalName` actually used anywhere?

jkwatson: Oh shoot. I added that while debugging a test and forgot to remove it. I'll remove it.

PR #55, opened 2021-06-29
Test publishing

breedx-splk:
Ok, this is purely for helping along the process.  The Dockerfile shouldn't exist, it just make repeating some things slightly more bearable.

Some notes:

## Env setup

Assuming you have already created a signing key and know the ID of it (`gpg --list-key`), you need to set up the key and password in the environment.

```
$ export GPG_SECRET_KEY=$(gpg --armor --export-secret-key <your-key-id>)
$ IFS= read -rsp 'Please the key password: ' GPG_PASSWORD
```
now you have those in your local environment.  

## Build container

```
$ docker build .
```

Yup, that's it. Make a note of the container ID listed at the bottom of the output.

## Running the signing task

Don't expand these vars, just let your shell do it:

```
docker run -it --rm \
    -e ORG_GRADLE_PROJECT_signingKey="$GPG_SECRET_KEY" \
    -e ORG_GRADLE_PROJECT_signingPassword=${GPG_PASSWORD}
```

#### try more stuff

To try more things, interactively, you can add `/bin/bash` on the end of the above docker command. Once you get that shell, run `./do_signing.sh`

# What's not working

So the signing task wants to sign publication _components_. Right now, the maven publication doesn't seem to know about the android _components_ or what it produces. Until that is set up (I have no idea how to do that), the signing won't work. See the `publications` block in the `publishing` block in the `splunk-otel-android` submodule.

Additional docs: https://docs.gradle.org/current/userguide/publishing_customization.html#sec:adding-variants-to-existing-components

Additional docs also: https://docs.gradle.org/current/userguide/signing_plugin.html

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+
+FROM openjdk:11.0.11-9-jdk
+
+WORKDIR /app
+
+RUN apt-get --quiet update --yes
+RUN apt-get --quiet install --yes wget tar unzip lib32stdc++6 lib32z1
+RUN wget --quiet --output-document=android-sdk.zip https://dl.google.com/android/repository/commandlinetools-linux-7302050_latest.zip
+RUN unzip -d android-sdk-linux android-sdk.zip
+RUN echo y | android-sdk-linux/cmdline-tools/bin/sdkmanager --sdk_root=. "platforms;android-30" >/dev/null
+RUN echo y | android-sdk-linux/cmdline-tools/bin/sdkmanager --sdk_root=. "platform-tools" >/dev/null
+RUN echo y | android-sdk-linux/cmdline-tools/bin/sdkmanager --sdk_root=. "build-tools;30.0.3" >/dev/null
+RUN export ANDROID_SDK_ROOT=/app
+RUN export PATH=$PATH:$PWD/platform-tools/
+
+RUN android-sdk-linux/cmdline-tools/bin/sdkmanager --sdk_root=. --licenses
+RUN git clone https://github.com/signalfx/splunk-otel-android.git
+
+WORKDIR /app/splunk-otel-android
+
+RUN git checkout -t origin/test_publishing
+RUN touch local.properties
+
+RUN ANDROID_SDK_ROOT=/app ./gradlew tasks
+
+COPY build.gradle /app/splunk-otel-android
+COPY splunk-otel-android/build.gradle /app/splunk-otel-android/splunk-otel-android
+
+RUN echo "" >> gradle.properties
+RUN echo "org.gradle.jvmargs=-Xmx4g -XX:MaxMetaspaceSize=1024m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8" >> gradle.properties
+RUN echo "gradle.daemon=false" >> gradle.properties
+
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -6,20 +6,25 @@ buildscript {
...
-        classpath 'com.vanniktech:gradle-maven-publish-plugin:0.16.0'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
 }
 
+plugins {
+  id 'signing'
+  id 'maven-publish'
```
jkwatson, line 17: is this needed here? even if we're only publishing/signing the splunk-otel-java module?

build.gradle
```diff
@@ -6,20 +6,25 @@ buildscript {
...
 allprojects {
     repositories {
         google()
         mavenCentral()
     }
+
 }
 
-task clean(type: Delete) {
+task cleanx(type: Delete) {
```
jkwatson, line 28: ?

splunk-otel-android/build.gradle
```diff
@@ -64,4 +65,34 @@ dependencies {
...
+publishing {
+  repositories {
+    maven {
+      name = "GitHubPackages"
+      url = uri("https://maven.pkg.github.com/signalfx/splunk-otel-android")
+      credentials {
+        username = System.getenv("GITHUB_ACTOR_FIXMEXXX")
+        password = System.getenv("GITHUB_TOKEN_FIXMEXXX")
+      }
+    }
```
jkwatson, line 79: why did you add this bit?

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+RUN ANDROID_SDK_ROOT=/app ./gradlew -x test build
+
+RUN echo 'export ANDROID_SDK_ROOT=/app' > do_signing.sh
+# RUN echo './gradlew -x test -PsigningKeyId=${GPG_KEY_ID} -PsigningInMemoryKey=$GPG_SECRET_KEY -PsigningInMemoryKeyPassword=$GPG_PASSWORD' signMavenPublication >> ./do_signing.sh
+
+RUN echo './gradlew -x test signMavenPublication' >> ./do_signing.sh
+RUN chmod 755 /app/splunk-otel-android/do_signing.sh
+
+
+CMD './do_signing.sh'
```
jkwatson, line 42: curious about this rigamarole. Any reason why we can't just directly call gradle here?

build.gradle
jkwatson: I got it. there's already a `clean` task from one of the plugins.

build.gradle
breedx-splk: probably not

build.gradle
breedx-splk: had to disable that because a `clean` task already exists.

splunk-otel-android/build.gradle
breedx-splk: copypasta!!

Dockerfile
breedx-splk: probably can. the script just allowed me to try some other things, like echoing environment vars etc. to make sure things were set up properly.

breedx-splk:
It's publishing! 🎉

PR #58, opened 2021-06-30
Add javadoc and sources jars to the publications

jkwatson:
This also tightens up the dependencies a bit.


breedx-splk (approved):
L.G.T.M.

PR #59, opened 2021-06-30
A few updates from the GDI specification changes.

jkwatson:
Waiting on the spec pr to merge before this is complete.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/Config.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/Config.java
@@ -27,7 +27,7 @@ import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
  */
 public class Config {
 
-    private final String beaconUrl;
+    private final String beaconEndpoint;
     private final String rumAuthToken;
     private final boolean debugEnabled;
     private final String applicationName;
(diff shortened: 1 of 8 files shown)


mateuszrzeszutek (approved):

jkwatson:
I guess we should hold off on this until the spec changes get merged. I'll put in a separate PR to update the sources jar classifier.

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
```diff
@@ -36,27 +37,33 @@
...
-        return beaconUrl;
+    public String getBeaconEndpoint() {
+        return beaconEndpoint;
     }
 
     /**
      * The configured RUM auth token for the library.
      */
-    public String getRumAuthToken() {
+    public String getRumAuth() {
```
mateuszrzeszutek, line 66: Should it be `getRumAuthToken()`?

jkwatson: I think it should actually be `getRumAccessToken()`. ;)

jkwatson: fixed.

mateuszrzeszutek (approved):

PR #69, opened 2021-07-02
Use the monotonic clock for initialization timings.

jkwatson:
It probably doesn't matter much, as I suspect that the system clock won't get jiggled during the very brief initialization time, but this is more correct.


laurit:
> It probably doesn't matter much, as I suspect that the system clock won't get jiggled during the very brief initialization time, but this is more correct.

Can system clock ever get jiggled, doesn't it just call `System.currentTimeMillis()`? I thought that this change is to get nanosecond precision for timings.

laurit (approved):

jkwatson:
> > It probably doesn't matter much, as I suspect that the system clock won't get jiggled during the very brief initialization time, but this is more correct.
> 
> 
> 
> Can system clock ever get jiggled, doesn't it just call `System.currentTimeMillis()`? I thought that this change is to get nanosecond precision for timings.

System.currentTimeMillis() is not monotonic. If the system clock gets adjusted, it can go backwards. It's not reliable for timings for this reason. System.nanotime() is monotonic, and so is perfect for timings, but not for getting wall clock values. If you combine them, you can get the best of both worlds, and nano precision.

PR #70, opened 2021-07-02
add a blob of config info to the init span

jkwatson:
/cc @flands

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
@@ -92,7 +92,7 @@ class RumInitializer {
             new NetworkMonitor(connectionUtil).addConnectivityListener(tracer);
         }
 
-        recordInitializationSpan(startTimeNanos, initializationEvents, tracer);
+        recordInitializationSpan(startTimeNanos, initializationEvents, tracer, config);
 
         return new SplunkRum(openTelemetrySdk, sessionId);
     }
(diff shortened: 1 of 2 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -107,11 +107,17 @@ private String detectRumVersion() {
...
+    private static void recordInitializationSpan(long startTimeNanos, List<InitializationEvent> initializationEvents, Tracer tracer, Config config) {
         Span span = tracer.spanBuilder("SplunkRum.initialize")
                 .setStartTimestamp(startTimeNanos, TimeUnit.NANOSECONDS)
                 .setAttribute(SplunkRum.COMPONENT_KEY, SplunkRum.COMPONENT_APPSTART)
                 .startSpan();
+
+        String configSettings = "[debug:" + config.isDebugEnabled() + "," +
+                "crashReporting:" + config.isCrashReportingEnabled() + "," +
+                "networkMonitor:" + config.isNetworkMonitorEnabled() + "]";
+        span.setAttribute("configSettings", configSettings);
```
mateuszrzeszutek, line 119: Nit: should we use a more otel-like attribute name?
```suggestion
        span.setAttribute("config_settings", configSettings);
```

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
jkwatson: no idea. :) Should we add something to the gdi-spec about this?

jkwatson: In the meantime, I'm happy to use snake_case on this.

PR #78, opened 2021-07-13
Implement very basic ANR detection and reporting.

jkwatson:
And add a button to the sample app to demo.

### DIFF
--- a/splunk-otel-android/src/test/java/com/splunk/rum/AnrWatcherTest.java
+++ b/splunk-otel-android/src/test/java/com/splunk/rum/AnrWatcherTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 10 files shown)


splunk-otel-android/src/test/java/com/splunk/rum/AnrWatcherTest.java
```diff
@@ -0,0 +1,101 @@
...
+        StackTraceElement[] stackTrace = new StackTraceElement[0];
+        when(mainThread.getStackTrace()).thenReturn(stackTrace);
+
+        AnrWatcher anrWatcher = new AnrWatcher(handler, mainThread, () -> splunkRum);
+        for (int i = 0; i < 5; i++) {
+            when(handler.post(isA(Runnable.class))).thenReturn(true);
+            anrWatcher.run();
+        }
+        verify(splunkRum).recordAnr(stackTrace);
+    }
```
mateuszrzeszutek, line 100: How about a test case that verifies that only one ANR is reported if the main thread does not respond for 6-9 seconds?

mateuszrzeszutek (approved):

splunk-otel-android/src/test/java/com/splunk/rum/AnrWatcherTest.java
jkwatson: Good idea. I'll add it.

mateuszrzeszutek (approved):

PR #79, opened 2021-07-14
Add a button to generate an ANR with tight loop on the main thread

jkwatson:
And fix a bug that made the ANR detection be at 10s, rather than 5s.


splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -112,7 +112,7 @@ private void initializeAnrReporting(Looper mainLooper) {
         Handler uiHandler = new Handler(mainLooper);
         AnrWatcher anrWatcher = new AnrWatcher(uiHandler, mainThread, SplunkRum::getInstance);
         Executors.newScheduledThreadPool(1)
-                .scheduleWithFixedDelay(anrWatcher, 1, 1, TimeUnit.SECONDS);
+                .scheduleAtFixedRate(anrWatcher, 1, 1, TimeUnit.SECONDS);
```
breedx-splk, line 115: 👍🏻

breedx-splk (approved):

PR #85, opened 2021-07-19
Remove the WorkflowTimer due to issues with cross-thread scope management

jkwatson:
There were really two issues here that I realized were dangerous:
1) By tying the Span & the Scope together, but not having any way to control how the WorkflowTimer might be passed between threads, we were potentially moving the Scope between threads and having it be closed on the wrong thread.
2) By making the WorkflowTimer Autocloseable, we were enabling the thing that we try to avoid in otel: https://github.com/open-telemetry/opentelemetry-java/blob/main/docs/rationale.md#span-not-closeable

So, this PR turns the workflow timer into a simple otel Span instance, with the appropriate attribute set on it for the user.


sample-app/src/main/java/com/splunk/android/sample/FirstFragment.java
```diff
@@ -84,11 +85,20 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
...
+            Span workflow = splunkRum.startWorkflow("Custom Workflow");
+            backgrounder.submit(() -> {
+                makeCall("https://ssidhu.o11ystore.com/");
+                workflow.end();
+            });
+        });
+        binding.httpMeBad.setOnClickListener(v -> {
+            Span workflow = splunkRum.startWorkflow("Workflow With Error");
+            backgrounder.submit(() -> {
+                makeCall("https://asdlfkjasd.asdfkjasdf.ifi");
```
breedx-splk, line 97: If the `makeCall` throws an exception, we will never make it to the `workflow.end()` below -- is that ok, or would it potentially "leak" a span? Do blocks like this need to be done in try...finally?

breedx-splk (approved):
Seems ok to me. Had a question about closing/leaking spans that might not get ended.

sample-app/src/main/java/com/splunk/android/sample/FirstFragment.java
jkwatson: generally, yes. Right now the `makeCall` catches anything that might be thrown, so it's probably ok for the sample app?  I agree it's not ideal, and this whole thing needs a bit of a rework to really be considered "good code".

There isn't really any span "leakage" in otel unless you hold a reference to it, since nothing in the SDK keeps a reference itself.

PR #87, opened 2021-07-20
Update/simplify the okhttp instrumentation based on new capabilities …

jkwatson:
…in the OpenTelemetry library instrumentation.

### DIFF
--- a/splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
+++ b/splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        RumResponseAttributesExtractor attributesExtractor = new RumResponseAttributesExtractor(headerParser);
+        AttributesBuilder attributesBuilder = Attributes.builder();
+        attributesExtractor.onEnd(attributesBuilder, fakeRequest, response);
+        attributesExtractor.onStart(attributesBuilder, fakeRequest);
+        Attributes attributes = attributesBuilder.build();
+
+        assertEquals("http", attributes.get(SplunkRum.COMPONENT_KEY));
+        assertNull(attributes.get(OkHttpRumInterceptor.LINK_TRACE_ID_KEY));
+        assertNull(attributes.get(OkHttpRumInterceptor.LINK_SPAN_ID_KEY));
```
breedx-splk, line 97: Nit: Might be worth also verifying that the content length is still recorded even without this special header?

breedx-splk (approved):
Looks good. One small thought about the last test.

breedx-splk:
Also these new APIs are 🔥

splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
jkwatson: added. good call

PR #89, opened 2021-07-22
Add a Chrome custom tab to the app and a timer to time how long the user visits it.

jkwatson:


kubawach (approved):
Looks good to me. Although it took me some time to read thru Android docs (no experience with it) and understand why the hell there is no "close event" for the tab object ;)

jkwatson:
Yes. This pr was the result of experimenting with what we had access to with the custom tab. The answer is: basically nothing. :(

PR #92, opened 2021-07-28
Update crash spans' names and component

mateuszrzeszutek:
In accordance with the latest RUM backend changes

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/CrashReporter.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/CrashReporter.java
@@ -50,12 +50,13 @@ class CrashReporter {
         public void uncaughtException(@NonNull Thread t, @NonNull Throwable e) {
             StringWriter writer = new StringWriter();
             e.printStackTrace(new PrintWriter(writer));
-            tracer.spanBuilder("crash.report")
+            String exceptionType = e.getClass().getSimpleName();
+            tracer.spanBuilder(exceptionType)
                     .setAttribute(SemanticAttributes.THREAD_ID, t.getId())
                     .setAttribute(SemanticAttributes.THREAD_NAME, t.getName())
                     .setAttribute(SemanticAttributes.EXCEPTION_STACKTRACE, writer.toString())
                     .setAttribute(SemanticAttributes.EXCEPTION_ESCAPED, true)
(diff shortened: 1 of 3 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/CrashReporter.java
```diff
@@ -50,12 +50,13 @@ static void initializeCrashReporting(Tracer tracer, OpenTelemetrySdk openTelemet
...
             e.printStackTrace(new PrintWriter(writer));
-            tracer.spanBuilder("crash.report")
+            String exceptionType = e.getClass().getSimpleName();
+            tracer.spanBuilder(exceptionType)
                     .setAttribute(SemanticAttributes.THREAD_ID, t.getId())
                     .setAttribute(SemanticAttributes.THREAD_NAME, t.getName())
                     .setAttribute(SemanticAttributes.EXCEPTION_STACKTRACE, writer.toString())
                     .setAttribute(SemanticAttributes.EXCEPTION_ESCAPED, true)
-                    .setAttribute(SplunkRum.COMPONENT_KEY, SplunkRum.COMPONENT_ERROR)
+                    .setAttribute(SplunkRum.COMPONENT_KEY, SplunkRum.COMPONENT_CRASH)
```
breedx-splk, line 59: I notice that `COMPONENT_ERROR` is still used for the ANRs...is that intentional?

breedx-splk (approved):
Looks good. Seems like maybe the only remaining usages of `COMPONENT_ERROR` are in the ANR code, so we should confirm that it's still wanted there and needed, otherwise remove it (future PR).

mateuszrzeszutek:
> Looks good. Seems like maybe the only remaining usages of COMPONENT_ERROR are in the ANR code, so we should confirm that it's still wanted there and needed, otherwise remove it (future PR).

I'm going to submit a separate PR that adds setting `component=error` to `SplunkRum#addRumException()`, because that's what the RUM team expects (I think). But I'll wait for John with merging that PR, I'd like to get his opinion on this.

splunk-otel-android/src/main/java/com/splunk/rum/CrashReporter.java
mateuszrzeszutek: Hmm, I guess so? ANRs are not necessarily fatal, unrecoverable errors, it's just the application being horribly slow.
(At the very least, that's my understanding - and I don't know that much about Android 😅 )

PR #96, opened 2021-07-29
Update SplunkRum#addRumException() according to RUM BE requirements

mateuszrzeszutek:
I've noticed that Android RUM differs in 2 things from what RUM backend expects:

* missing `component=error` attribute;
* span name is different than `exception.type` attribute; in fact, it may be completely custom one.

I've added 2 new methods that do not allow setting custom exception span name and deprecated the old one - @jkwatson please take a look at this, whether this makes sense or not.

### DIFF
--- a/sample-app/src/main/java/com/splunk/android/sample/MainActivity.java
+++ b/sample-app/src/main/java/com/splunk/android/sample/MainActivity.java
@@ -74,8 +74,9 @@ public class MainActivity extends AppCompatActivity {
         //noinspection SimplifiableIfStatement
         if (id == R.id.action_settings) {
             SplunkRum.getInstance()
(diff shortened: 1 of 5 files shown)


jkwatson:
Was this a request from the BE team? IIRC, there was some reluctance to make things generally show up as error spans, but I don't remember the details.

jkwatson (approved):
Looks good! I'm fine with not having a custom `name` if the backend/frontend doesn't want it.

jkwatson:
Should we also update the usage(s) in the sample app to not use the deprecated method?

mateuszrzeszutek:
> Should we also update the usage(s) in the sample app to not use the deprecated method?

Oh, definitely, I missed that one usage.

PR #98, opened 2021-08-04
Only record the last.screen.name during screen transitions.

jkwatson:


splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
```diff
@@ -68,10 +68,6 @@ public void onStart(Context parentContext, ReadWriteSpan span) {
 
         String currentScreen = visibleScreenTracker.getCurrentlyVisibleScreen();
         span.setAttribute(SplunkRum.SCREEN_NAME_KEY, currentScreen);
-        String previouslyVisibleScreen = visibleScreenTracker.getPreviouslyVisibleScreen();
-        if (previouslyVisibleScreen != null) {
-            span.setAttribute(SplunkRum.LAST_SCREEN_NAME_KEY, previouslyVisibleScreen);
-        }
```
mateuszrzeszutek, line 74: Shouldn't we update the test class and remove the 
```java
when(visibleScreenTracker.getPreviouslyVisibleScreen()).thenReturn(null);
```
stubbing?

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
jkwatson: fixed

PR #100, opened 2021-08-04
Add the ability to update the global attributes.

jkwatson:


splunk-otel-android/src/main/java/com/splunk/rum/Config.java
```diff
@@ -111,6 +119,12 @@ public boolean isAnrDetectionEnabled() {
         return anrDetectionEnabled;
     }
 
+    void updateGlobalAttributes(Consumer<AttributesBuilder> updater) {
+        AttributesBuilder builder = globalAttributes.get().toBuilder();
+        updater.accept(builder);
+        globalAttributes.set(builder.build());
```
mateuszrzeszutek, line 125: Nit: I assume we're not worried about a concurrent update scenario here?

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
jkwatson: I'm not, but it's a good callout that we should at least have some javadoc stating.

jkwatson: done.

PR #102, opened 2021-08-05
Hook up an OkHttp Dispatcher to propagate context into calls.

jkwatson:
For the demo app, it's nice to demonstrate how to get span-causality for http calls.


mateuszrzeszutek (approved):

jkwatson:
Just as an FYI... even though this is a safe usage in the demo app, it may not be safe in general, due the way that okhttp queues calls under high load. Mis-parenting of spans probably isn't a huge deal in Android, but I'm going to be working on some better approaches to doing this in opentelemetry-java-instrumentation and will probably use this example as the way to experiment with those better approaches.

PR #105, opened 2021-08-11
Add API requirements to the docs and reorganize slightly

jkwatson:


breedx-splk (approved):

README.md
```diff
@@ -155,8 +155,8 @@ In order to build and run the sample application, you will need to configure a `
 in the root of the project. It will need to have two properties configured:
 
 ```properties
-rum.access.token=<a valid Splunk RUM access token>
-rum.beacon.url=https://rum-ingest.<realm>.signalfx.com/v1/rum
+rum.realm=<realm>
+rum.access.token=<a valid Splunk RUM access token for the realm>
```
breedx-splk, line 159: Is the URL no longer configurable?

jkwatson: it totally is, but realm is much easier so I put that in here for the most common use-case for the demo app

jkwatson: well, the demo app only uses the realm, to be more clear, so this has to have the realm here. :)

breedx-splk: Doesn't this couple the implementation to the url scheme of the ingest endpoint? What if they move it?

jkwatson: If they do that, every device and browser won't be able to send data. If they change it, we'll update the library for the new scheme.

jkwatson: And this library already supports a custom endpoint. Realm is just a convenience so people don't need to worry about the precise URL if they don't want to.

mateuszrzeszutek (approved):

README.md
breedx-splk: Sorry to be pedantic, but then you also need to have defined behavior in the event that the user supplies both realm and URL (and when they conflict)? More complicated than just a URL IMO.

jkwatson: I believe it is defined in the Config.Builder javadoc. Or at least, it's alluded to pretty strongly that you would use one or the other and not both.

PR #108, opened 2021-08-12
Remove the screen name from the span name for lifecycle spans.

jkwatson:


sample-app/src/main/java/com/splunk/android/sample/FirstFragment.java
```diff
@@ -120,9 +121,11 @@ public void onFailure(@NonNull Call call, @NonNull IOException e) {
...
                 @Override
                 public void onResponse(@NonNull Call call, @NonNull Response response) {
-                    int responseCode = response.code();
-                    httpResponse.postValue("" + responseCode);
-                    workflow.end();
+                    try (ResponseBody body = response.body()) {
+                        int responseCode = response.code();
+                        httpResponse.postValue("" + responseCode);
+                        workflow.end();
+                    }
```
jkwatson, line 128: I also snuck this change in since okhttp was yelling at me in the logs while I was testing things.

breedx-splk (approved):
Seems ok to me. ✅

PR #113, opened 2021-08-17
Create a new span to represent the full AppStart sequence.

jkwatson:
Looks like this in the UI: 

image

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.rum;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+
+class AppStartupTimer {
+    private final AtomicLong firstPossibleTimestamp = new AtomicLong(System.currentTimeMillis());
+    private final AtomicReference<Span> overallAppStartSpan = new AtomicReference<>();
+
+    Span start(Tracer tracer) {
+        //guard against a double-start and just return what's already in flight.
+        if (overallAppStartSpan.get() != null) {
+            return overallAppStartSpan.get();
+        }
+        final Span appStart = tracer.spanBuilder("AppStart")
+                .setStartTimestamp(firstPossibleTimestamp.get(), TimeUnit.MILLISECONDS)
+                .setAttribute(SplunkRum.COMPONENT_KEY, SplunkRum.COMPONENT_APPSTART)
+                .setAttribute(SplunkRum.START_TYPE_KEY, "cold")
+                .startSpan();
+        overallAppStartSpan.set(appStart);
+        return appStart;
+    }
+
+    void end() {
+        Span span = overallAppStartSpan.get();
+        if (span != null) {
+            span.end();
+            overallAppStartSpan.set(null);
+        }
+    }
+
+    Span getStartupSpan() {
+        return overallAppStartSpan.get();
+    }
+}
--- a/splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
@@ -16,6 +16,8 @@
 
 package com.splunk.rum;
 
+import androidx.annotation.Nullable;
+
 import java.util.concurrent.TimeUnit;
 
 import io.opentelemetry.api.trace.Span;
@@ -46,6 +48,7 @@ class AppStartupTimer {
         }
     }
 
+    @Nullable
     Span getStartupSpan() {
         return overallAppStartSpan;
     }
--- a/splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
@@ -17,39 +17,36 @@
 package com.splunk.rum;
 
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.AtomicReference;
 
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.Tracer;
 
 class AppStartupTimer {
-    private final AtomicLong firstPossibleTimestamp = new AtomicLong(System.currentTimeMillis());
-    private final AtomicReference<Span> overallAppStartSpan = new AtomicReference<>();
+    private final long firstPossibleTimestamp = System.currentTimeMillis();
+    private volatile Span overallAppStartSpan = null;
 
     Span start(Tracer tracer) {
         //guard against a double-start and just return what's already in flight.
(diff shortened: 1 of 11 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/NamedTrackableTracer.java
```diff
@@ -119,13 +135,14 @@ public void endSpanForActivityResumed() {
 
     @Override
     public void endActiveSpan() {
+        SplunkRum.getStartupTimer().end();
```
jkwatson, line 138: this is a little bit of a bummer. I'd like to figure out a way to know that we're in the right kind of span when we do this, rather than blindly ending it always.

jkwatson: I added a null check on the startup span, but it's still a bit ugly to blindly ending the startup span here, when there should only ever be one.

splunk-otel-android/src/main/java/com/splunk/rum/SplunkRum.java
```diff
@@ -41,12 +41,15 @@
  * Entrypoint for Splunk's Android RUM (Real User Monitoring) support.
  */
 public class SplunkRum {
+    private static AppStartupTimer startupTimer = new AppStartupTimer();
```
mateuszrzeszutek, line 44: Nit: `final`?

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -125,11 +128,15 @@ private String detectRumVersion() {
...
-    private static void recordInitializationSpan(long startTimeNanos, List<InitializationEvent> initializationEvents, Tracer tracer, Config config) {
+    private void recordInitializationSpans(long startTimeNanos, List<InitializationEvent> initializationEvents, Tracer tracer, Config config) {
+        Span overallAppStart = startupTimer.start(tracer);
+        System.out.println("overallAppStart = " + overallAppStart);
         Span span = tracer.spanBuilder("SplunkRum.initialize")
+                .setParent(Context.current().with(overallAppStart))
                 .setStartTimestamp(startTimeNanos, TimeUnit.NANOSECONDS)
                 .setAttribute(SplunkRum.COMPONENT_KEY, SplunkRum.COMPONENT_APPSTART)
                 .startSpan();
+        System.out.println("span = " + span);
```
mateuszrzeszutek, line 139:
```suggestion
```

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+
+class AppStartupTimer {
+    private final AtomicLong firstPossibleTimestamp = new AtomicLong(System.currentTimeMillis());
```
mateuszrzeszutek, line 27: What's the reason for an `AtomicLong` here? Are we updating this value somewhere?

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
```diff
@@ -0,0 +1,55 @@
...
+    private final AtomicReference<Span> overallAppStartSpan = new AtomicReference<>();
```
mateuszrzeszutek, line 28: Nit: it probably could be just a `volatile` field, you're not using CAS anywhere in this class.

splunk-otel-android/src/main/java/com/splunk/rum/NamedTrackableTracer.java
```diff
@@ -119,13 +135,17 @@ public void endSpanForActivityResumed() {
 
     @Override
     public void endActiveSpan() {
+        Span startupSpan = SplunkRum.getStartupTimer().getStartupSpan();
+        if (startupSpan != null) {
+            startupSpan.end();
+        }
```
mateuszrzeszutek, line 141: Why not simply call `SplunkRum.getStartupTimer().end()`? Is there any reason behind duplicating a part of that method?

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -125,11 +128,15 @@ private String detectRumVersion() {
         return "unknown";
     }
 
-    private static void recordInitializationSpan(long startTimeNanos, List<InitializationEvent> initializationEvents, Tracer tracer, Config config) {
+    private void recordInitializationSpans(long startTimeNanos, List<InitializationEvent> initializationEvents, Tracer tracer, Config config) {
+        Span overallAppStart = startupTimer.start(tracer);
+        System.out.println("overallAppStart = " + overallAppStart);
```
mateuszrzeszutek, line 133:
```suggestion
```

splunk-otel-android/src/main/java/com/splunk/rum/NamedTrackableTracer.java
mateuszrzeszutek: Also, would it make sense to inject the `AppStartupTimer` via the constructor instead of using the global?

splunk-otel-android/src/main/java/com/splunk/rum/NamedTrackableTracer.java
mateuszrzeszutek: > I'd like to figure out a way to know that we're in the right kind of span when we do this, rather than blindly ending it always.

I wonder if the current problem might be caused by the fact that `NamedTrackableTracer`'s supposed to be completely independent of the lifecycle phases, and yet at the same time it must keep track of whether the current execution is an application start, whether it's a cold or hot one. Just thinking out loud, but what if we had separate instances (and implementations) for each lifecycle? When you `endActiveSpan()` the next time you create one a different implementation of the tracer might get chosen, depending on what lifecycle callback gets called.

jkwatson: That's a possibility. It would lead to more allocations, rather than just one per Activity/Fragment, but it's definitely worth a try. I don't want to try it in this PR, but I like the suggestion.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
jkwatson: 🤦🏽

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
jkwatson: good point! it was initially being updated, but that got changed early in the experimentation. will fix.

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
jkwatson: TBH, I don't know how volatile works on Android, and I tend to favor atomics over bare volatiles, but I'll do a bit of research on whether volatile has the same semantics as on the standard JVM.

jkwatson: cool. looks like as of Android 4, it does work the same as it is supposed to in the JMM:  https://stackoverflow.com/questions/6973667/dalvik-vm-java-memory-model-concurrent-programming-on-android

splunk-otel-android/src/main/java/com/splunk/rum/NamedTrackableTracer.java
jkwatson: > Why not simply call `SplunkRum.getStartupTimer().end()`? Is there any reason behind duplicating a part of that method?

heh. I actually had this initially, and thought it was even more "blindly doing the dumb thing", and add the explicit check. Honestly, I think both are bad options. ;)

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
jkwatson: I think I'm juggling too many things at once. Thanks for keeping me honest on this stuff!

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/com/splunk/rum/NamedTrackableTracer.java
jkwatson: ok, changed to injecting the startup timer. Next step (next PR) will to break the NamedTrackableTracer into 2 classes, one for activities, and one for fragments.

PR #120, opened 2021-08-20
Hack the zipkin encoder to preserve span name casing.

jkwatson:
The Zipkin Span.Builder lowercases all span names, without an option to disable this behavior. To work around this, we add an extra attribute with the actual span name, and then munge the output of the zipkin V2 Span Encoder implementation to replace the proper span name from the attribute value.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.rum;
+
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+import zipkin2.Span;
+import zipkin2.codec.BytesEncoder;
+import zipkin2.codec.Encoding;
+import zipkin2.internal.JsonCodec;
+import zipkin2.internal.V2SpanWriter;
+import zipkin2.internal.WriteBuffer;
+
+class CustomZipkinEncoder implements BytesEncoder<Span> {
+
+    private final WriteBuffer.Writer<Span> writer = new V2SpanWriter();
+
+    public Encoding encoding() {
+        return Encoding.JSON;
+    }
+
+    public int sizeInBytes(Span span) {
+        return this.writer.sizeInBytes(span);
+    }
+
+    public byte[] encode(Span span) {
+        String properSpanName = span.tags().get(RumAttributeAppender.SPLUNK_OPERATION_KEY.getKey());
+
+        //note: this can be optimized, if necessary. Let's keep it simple for now.
+        byte[] rawBytes = JsonCodec.write(this.writer, span);
+        String renamedResult = new String(rawBytes).replace(
+                "\"name\":\"" + span.name() + "\"",
+                "\"name\":\"" + properSpanName + "\"");
+        return renamedResult.getBytes(StandardCharsets.UTF_8);
+    }
+
+    public byte[] encodeList(List<Span> spans) {
+        //note: this doesn't appear to be called in our current code paths, so let's leave it be.
+        return JsonCodec.writeList(this.writer, spans);
+    }
+
+}
--- a/splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
@@ -36,14 +36,17 @@ class CustomZipkinEncoder implements BytesEncoder<Span> {
 
     private final WriteBuffer.Writer<Span> writer = new V2SpanWriter();
 
+    @Override
     public Encoding encoding() {
         return Encoding.JSON;
     }
 
+    @Override
     public int sizeInBytes(Span span) {
         return this.writer.sizeInBytes(span);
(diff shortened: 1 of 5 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+
+import zipkin2.Span;
+import zipkin2.codec.BytesEncoder;
+import zipkin2.codec.Encoding;
+import zipkin2.internal.JsonCodec;
+import zipkin2.internal.V2SpanWriter;
+import zipkin2.internal.WriteBuffer;
+
+class CustomZipkinEncoder implements BytesEncoder<Span> {
```
dtagatac-splunk, line 29: Can we annotate this class to explain its purpose, and possibly link to the upstream bug/feature request that we are working around?

dtagatac-splunk:
Upstream issue where zipkin maintainers have decided not to support mixed case: https://github.com/openzipkin/zipkin/issues/2438

splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
jkwatson: Added a bunch of javadoc. Unfortunately, the zipkin authors have said they will not change this behavior, so they don't consider it a bug...it's just the way the zipkin library works.

splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
```diff
@@ -0,0 +1,63 @@
...
+                "\"name\":\"" + properSpanName + "\"");
+        return renamedResult.getBytes(StandardCharsets.UTF_8);
+    }
+
+    public byte[] encodeList(List<Span> spans) {
+        //note: this doesn't appear to be called in our current code paths, so let's leave it be.
+        return JsonCodec.writeList(this.writer, spans);
+    }
+
+}
```
breedx-splk, line 63: Can we please get `@Override` on the interface methods?

splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
```diff
@@ -0,0 +1,63 @@
...
+        return this.writer.sizeInBytes(span);
+    }
+
+    public byte[] encode(Span span) {
+        String properSpanName = span.tags().get(RumAttributeAppender.SPLUNK_OPERATION_KEY.getKey());
+
+        //note: this can be optimized, if necessary. Let's keep it simple for now.
+        byte[] rawBytes = JsonCodec.write(this.writer, span);
+        String renamedResult = new String(rawBytes).replace(
+                "\"name\":\"" + span.name() + "\"",
```
breedx-splk, line 53: This is probably fine. It _feels_ a tad fragile, because whitespace after the colon breaks this, but assuming the other encoder compacts stuff and won't change, that's probably fine.

splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
```diff
@@ -39,6 +39,7 @@
     //note: these will be in the otel semantic conventions as of spec v1.6.0
     static final AttributeKey<String> NETWORK_TYPE_KEY = stringKey("host.connection.type");
     static final AttributeKey<String> NETWORK_SUBTYPE_KEY = stringKey("host.connection.subtype");
+    static final AttributeKey<String> SPLUNK_OPERATION_KEY = stringKey("_splunk_operation");
```
breedx-splk, line 42: I didn't see anything that would remove or filter out this attribute before export. Are we ok with it making its way into ingest?

breedx-splk (changes requested):
Looks good enough for a hack!  🪓 
One nitpick thinger.

splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
jkwatson: of course. No idea why IDEA didn't do that for me. 🤔

splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
jkwatson: yep. it does, and the test confirms it. :)

splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
jkwatson: yeah... right now we'll get an extra attribute, which isn't harmful (aside from a bit of extra data). A further refinement would be to figure out how to elide this attribute in the encoder, but I think that would require writing the entire `Writer<Span>` implementation or doing some really dirty hacks. Because this ends up in a JSON array of tags, we'd have to be clever about removing commas conditionally, etc.

splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
jkwatson: but, absolutely. This is pretty much terrible in all ways. The other option is to write a new `Writer<Span>` to handle all of this, or pull in the Zipkin one into our source and start hacking on it. I didn't go that way for now because I want to figure out what the OSS process implications are for including foreign sourcecode in our distribution (not a legal issue, because zipkin is Apache 2.0, but a Splunk process issue).

splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
jkwatson: done

breedx-splk (approved):
Cool beans.

splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
```diff
@@ -56,6 +57,8 @@
 
     @Override
     public void onStart(Context parentContext, ReadWriteSpan span) {
+        //set this custom attribute in order to let the CustomZipEncoder use it for the span name on the wire.
```
jkwatson, line 60:
```suggestion
        //set this custom attribute in order to let the CustomZipkinEncoder use it for the span name on the wire.
```

PR #127, opened 2021-08-23
turn on minification for the sample app

jkwatson:
This will hopefully surface potential proguard issues before they get to customers.


breedx-splk (approved):
Seems ok. Is proguard minification the standard on Android?

jkwatson:
> Seems ok. Is proguard minification the standard on Android?

It's at least the built-in.

PR #135, opened 2021-08-27
Introduce a class to hold standard RUM attributes

jkwatson:
Useful until they are included in the OTel semantic conventions.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/StandardAttributes.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/StandardAttributes.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/StandardAttributes.java
```diff
@@ -0,0 +1,33 @@
...
+package com.splunk.rum;
+
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+
+/**
+ * This class hold {@link AttributeKey}s for standard RUM-related attributes that are not in the
+ * OpenTelemetry {@link io.opentelemetry.semconv.trace.attributes.SemanticAttributes} definitions.
+ */
+public class StandardAttributes {
```
mateuszrzeszutek, line 26: Nit: make it final and add a private default constructor?

mateuszrzeszutek (approved):

PR #136, opened 2021-08-31
Disable the network monitor events and the ANR detector when the app …

jkwatson:
…is put in the background.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
@@ -26,6 +26,7 @@ import androidx.fragment.app.FragmentActivity;
 import androidx.fragment.app.FragmentManager;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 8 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
```diff
@@ -38,11 +39,15 @@
     private final Tracer tracer;
     private final VisibleScreenTracker visibleScreenTracker;
     private final AppStartupTimer startupTimer;
+    private final List<AppStateListener> appStateListeners;
+    //we count the number of activities that have been "started" and not yet "stopped" here to figure out when the app goes into the background.
+    private int numberOfOpenActivities = 0;
```
mateuszrzeszutek, line 44: Can this be a "normal" integer? No need for thread safety?

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
jkwatson: It can be, since this is only ever accessed from the UI thread (which is where the callbacks run).

PR #137, opened 2021-08-31
Provide a means of redacting data from export

mateuszrzeszutek:
Implements:
* Rejecting spans from export by name and/or attribute
* Removing/modifying attribute value in a span before export

Modifying span name before export could be easily implemented if needed. I didn't implement modifying span events, it'd be a bit harder to do since events have name + attributes at the very least and we don't expose SDK types like `EventData` to the user of the library.
The fact that we don't expose SDK types is also the reason why I built the `SpanFilterBuilder` class and didn't just use a simple `SpanData -> SpanData?` function.

### DIFF
--- a/sample-app/src/main/java/com/splunk/android/sample/SecondFragment.java
+++ b/sample-app/src/main/java/com/splunk/android/sample/SecondFragment.java
@@ -39,6 +39,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
+import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Scope;
@@ -93,9 +94,12 @@ public class SecondFragment extends Fragment {
                 span.end();
             }
         });
-        binding.buttonToWebview.setOnClickListener(v ->
-                NavHostFragment.findNavController(SecondFragment.this)
-                        .navigate(R.id.action_SecondFragment_to_webViewFragment));
+        binding.buttonToWebview.setOnClickListener(v -> {
+            SplunkRum.getInstance().addRumEvent("this span will be ignored", Attributes.empty());
+
+            NavHostFragment.findNavController(SecondFragment.this)
+                    .navigate(R.id.action_SecondFragment_to_webViewFragment);
+        });
 
         binding.buttonToCustomTab.setOnClickListener(v -> {
             String url = "https://ssidhu.o11ystore.com/";
--- a/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
@@ -177,11 +177,12 @@ class RumInitializer {
         }
         SpanExporter zipkinSpanExporter = getCoreSpanExporter(endpoint);
 
-        return ThrottlingExporter.newBuilder(new BufferingExporter(connectionUtil, zipkinSpanExporter))
+        ThrottlingExporter throttlingExporter = ThrottlingExporter.newBuilder(new BufferingExporter(connectionUtil, zipkinSpanExporter))
                 .categorizeByAttribute(SplunkRum.COMPONENT_KEY)
                 .maxSpansInWindow(100)
                 .windowSize(Duration.ofSeconds(30))
                 .build();
+        return config.decorateWithSpanFilter(throttlingExporter);
     }
 
     //visible for testing
--- a/splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
@@ -31,8 +31,8 @@ import io.opentelemetry.sdk.trace.export.SpanExporter;
 public final class SpanFilterBuilder {
 
     private Predicate<String> rejectSpanNamesPredicate = spanName -> false;
-    private final Map<AttributeKey<String>, Predicate<String>> rejectSpanAttributesPredicates = new HashMap<>();
-    private final Map<AttributeKey<String>, Function<String, String>> spanAttributeReplacements = new HashMap<>();
+    private final Map<AttributeKey<?>, Predicate<?>> rejectSpanAttributesPredicates = new HashMap<>();
+    private final Map<AttributeKey<?>, Function<?, ?>> spanAttributeReplacements = new HashMap<>();
 
     SpanFilterBuilder() {
     }
@@ -62,18 +62,18 @@ public final class SpanFilterBuilder {
      *                                with matching value should be rejected.
      * @return {@code this}.
      */
-    public SpanFilterBuilder rejectSpansByAttributeValue(
-            AttributeKey<String> attributeKey,
-            Predicate<String> attributeValuePredicate) {
+    public <T> SpanFilterBuilder rejectSpansByAttributeValue(
+            AttributeKey<T> attributeKey,
+            Predicate<? super T> attributeValuePredicate) {
 
         rejectSpanAttributesPredicates.compute(attributeKey,
                 (k, oldValue) -> oldValue == null
                         ? attributeValuePredicate
-                        : oldValue.or(attributeValuePredicate));
+                        : ((Predicate<T>) oldValue).or(attributeValuePredicate));
         return this;
     }
 
-    public SpanFilterBuilder removeSpanAttribute(AttributeKey<String> attributeKey) {
+    public <T> SpanFilterBuilder removeSpanAttribute(AttributeKey<T> attributeKey) {
         return removeSpanAttribute(attributeKey, value -> true);
     }
 
@@ -88,9 +88,9 @@ public final class SpanFilterBuilder {
      *                                value should be removed from the span.
      * @return {@code this}.
      */
-    public SpanFilterBuilder removeSpanAttribute(
-            AttributeKey<String> attributeKey,
-            Predicate<String> attributeValuePredicate) {
+    public <T> SpanFilterBuilder removeSpanAttribute(
+            AttributeKey<T> attributeKey,
+            Predicate<? super T> attributeValuePredicate) {
 
         return replaceSpanAttribute(attributeKey, old -> attributeValuePredicate.test(old) ? null : old);
     }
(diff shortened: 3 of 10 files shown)


sample-app/src/main/java/com/splunk/android/sample/SecondFragment.java
```diff
@@ -93,9 +94,12 @@ public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
                 span.end();
             }
         });
-        binding.buttonToWebview.setOnClickListener(v ->
-                NavHostFragment.findNavController(SecondFragment.this)
-                        .navigate(R.id.action_SecondFragment_to_webViewFragment));
+        binding.buttonToWebview.setOnClickListener(v -> {
+            SplunkRum.getInstance().addRumEvent("this span will be ignored", Attributes.empty());
```
jkwatson, line 98: doesn't this have to just be "ignored" to match the span filter defined in the config? Or should we update the span filter to have a `contains("ignored")`?

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+            Predicate<String> attributeValuePredicate) {
+
+        rejectSpanAttributesPredicates.compute(attributeKey,
+                (k, oldValue) -> oldValue == null
+                        ? attributeValuePredicate
+                        : oldValue.or(attributeValuePredicate));
+        return this;
+    }
+
+    public SpanFilterBuilder removeSpanAttribute(AttributeKey<String> attributeKey) {
```
jkwatson, line 76: this is missing javadoc.   But more importantly, are we sure that `<String>` is good enough here? Why not any type?

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+     * {@code attributeValuePredicate} will be removed from the span before it is exported.
+     *
+     * @param attributeKey            An attribute key to match.
+     * @param attributeValuePredicate A function that returns true if an attribute with matching
+     *                                value should be removed from the span.
+     * @return {@code this}.
+     */
+    public SpanFilterBuilder removeSpanAttribute(
+            AttributeKey<String> attributeKey,
+            Predicate<String> attributeValuePredicate) {
```
jkwatson, line 93: could we make this use a parameterized type, rather than forcing `<String>` ?

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+     * from the span.
+     *
+     * @param attributeKey           An attribute key to match.
+     * @param attributeValueModifier A function that receives the old attribute value and returns
+     *                               the new one.
+     * @return {@code this}.
+     */
+    public SpanFilterBuilder replaceSpanAttribute(
+            AttributeKey<String> attributeKey,
+            Function<String, String> attributeValueModifier) {
```
jkwatson, line 113: same question about parameterizing the type here

sample-app/src/main/java/com/splunk/android/sample/SecondFragment.java
mateuszrzeszutek: 🙈

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
mateuszrzeszutek: Added javadoc

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
mateuszrzeszutek: Yeah, parameterized types work, changed in all methods.
(Java generics are super painful to deal with when you have to use `?` wildcards though...)

jkwatson: agreed. it's definitely not easy to figure out.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -177,11 +177,12 @@ SpanExporter buildExporter(ConnectionUtil connectionUtil) {
         }
         SpanExporter zipkinSpanExporter = getCoreSpanExporter(endpoint);
 
-        return ThrottlingExporter.newBuilder(new BufferingExporter(connectionUtil, zipkinSpanExporter))
+        ThrottlingExporter throttlingExporter = ThrottlingExporter.newBuilder(new BufferingExporter(connectionUtil, zipkinSpanExporter))
                 .categorizeByAttribute(SplunkRum.COMPONENT_KEY)
                 .maxSpansInWindow(100)
                 .windowSize(Duration.ofSeconds(30))
                 .build();
+        return config.decorateWithSpanFilter(throttlingExporter);
```
jkwatson, line 185: 👍🏽

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+     */
+    public SpanFilterBuilder rejectSpansByName(Predicate<String> spanNamePredicate) {
+        rejectSpanNamesPredicate = rejectSpanNamesPredicate.or(spanNamePredicate);
+        return this;
+    }
+
+    /**
+     * Remove matching spans from the exporter pipeline.
+     * <p>
+     * Span that contains an attribute with key {@code attributeKey} and value matching the
```
jkwatson, line 57:
```suggestion
     * Any Span that contains an attribute with key {@code attributeKey} and value matching the
```

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+    private final Map<AttributeKey<?>, Predicate<?>> rejectSpanAttributesPredicates = new HashMap<>();
+    private final Map<AttributeKey<?>, Function<?, ?>> spanAttributeReplacements = new HashMap<>();
+
+    SpanFilterBuilder() {
+    }
+
+    /**
+     * Remove matching spans from the exporter pipeline.
+     * <p>
+     * Span names that match the {@code spanNamePredicate} will not be exported.
```
jkwatson, line 43:
```suggestion
     * Spans with names that match the {@code spanNamePredicate} will not be exported.
```

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+                (k, oldValue) -> oldValue == null
+                        ? attributeValuePredicate
+                        : ((Predicate<T>) oldValue).or(attributeValuePredicate));
+        return this;
+    }
+
+    /**
+     * Modify span data before it enters the exporter pipeline.
+     * <p>
+     * An attribute with key {@code attributeKey} and  will be removed from the span before it is exported.
```
jkwatson, line 79:
```suggestion
     * Any attribute with key {@code attributeKey} and  will be removed from the span before it is exported.
```

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+     * @return {@code this}.
+     */
+    public <T> SpanFilterBuilder removeSpanAttribute(AttributeKey<T> attributeKey) {
+        return removeSpanAttribute(attributeKey, value -> true);
+    }
+
+    /**
+     * Modify span data before it enters the exporter pipeline.
+     * <p>
+     * An attribute with key {@code attributeKey} and value matching the
```
jkwatson, line 91:
```suggestion
     * Any attribute with key {@code attributeKey} and value matching the
```

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+            AttributeKey<T> attributeKey,
+            Predicate<? super T> attributeValuePredicate) {
+
+        return replaceSpanAttribute(attributeKey, old -> attributeValuePredicate.test(old) ? null : old);
+    }
+
+    /**
+     * Modify span data before it enters the exporter pipeline.
+     * <p>
+     * Value of an attribute with key {@code attributeKey} will be passed to the
```
jkwatson, line 109:
```suggestion
     * The value of any attribute with key {@code attributeKey} will be passed to the
```

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+     * {@code attributeValueModifier} function. Value returned by the function will replace the
```
jkwatson, line 110:
```suggestion
     * {@code attributeValueModifier} function. the value returned by the function will replace the
```

splunk-otel-android/src/test/java/com/splunk/rum/SpanFilterTest.java
```diff
@@ -0,0 +1,226 @@
...
+        assertEquals("second", exportedSpans.get(1).getName());
+        assertEquals(Attributes.of(ATTRIBUTE, "not test", OTHER_ATTRIBUTE, "test"), exportedSpans.get(1).getAttributes());
+    }
+
+    @Test
+    public void shouldReplaceSpanAttributes() {
+        // given
+        SpanExporter underTest = new SpanFilterBuilder()
+                .replaceSpanAttribute(ATTRIBUTE, value -> value + "!!!")
+                .replaceSpanAttribute(ATTRIBUTE, value -> value + "1")
```
jkwatson, line 146: whoa. this is a weird one. I hadn't thought of combinations like this. Should be fine, though, I think.

splunk-otel-android/src/test/java/com/splunk/rum/SpanFilterTest.java
```diff
@@ -0,0 +1,226 @@
...
+                .replaceSpanAttribute(LONG_ATTRIBUTE, value -> value + 1)
+                // make sure that attribute types are taken into account
+                .replaceSpanAttribute(stringKey("long_attribute"), value -> "abc")
```
jkwatson, line 149: I've actually wondered if this might end up being more confusing than simply matching on the key's string value. I honestly don't know if users will grok this difference, but we can always loosen it up later if we get complaints or confusion.

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+     * original value. When the modifier function returns {@code null} the attribute will be removed
```
jkwatson, line 111: do we have a test for the `null` case?

mateuszrzeszutek: Indirectly - the `removeSpanAttribute()` method passes a function that returns null when the predicate passes.

splunk-otel-android/src/test/java/com/splunk/rum/SpanFilterTest.java
mateuszrzeszutek: We can probably add overloads (like all those `AttributesBuilder#put()` methods) if needed.

splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
jkwatson: I think it's worth adding a simple direct test to verify that we don't break this with an internal change in the future.

jkwatson (approved):

PR #143, opened 2021-09-06
Provide a means of filtering spans based on name

mateuszrzeszutek:
It almost seems too easy with the existing `SpanFilterBuilder` framework...

Do we need a `rejectSpansByName(Pattern...)` variant? Or is calling `rejectSpansByName()` multiple times acceptable?

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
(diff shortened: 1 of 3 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -37,6 +38,19 @@
...
+    /**
+     * Remove matching spans from the exporter pipeline.
+     * <p>
+     * Spans with names that match the passed {@code pattern} will not be exported.
+     *
+     * @param pattern A regular expression pattern that matches names of spans that should be
+     *                rejected.
+     * @return {@code this}.
+     */
+    public SpanFilterBuilder rejectSpansByName(Pattern pattern) {
```
jkwatson, line 50: I think we should hold off on adding this method until we get feedback that it's definitely necessary. It's so easy for the end user to implement this themselves using the `rejectSpansByName(Predicate<String>)` method that I think we can start that just that one.

mateuszrzeszutek: Got it 👍 
Let's close it for now and wait for more feedback on span filtering then.

PR #144, opened 2021-09-06
Provide a way to remove/skip http client spans based on request URL

mateuszrzeszutek:


splunk-otel-android/src/main/java/com/splunk/rum/SpanFilterBuilder.java
```diff
@@ -103,6 +138,25 @@ public SpanFilterBuilder rejectSpansByName(Predicate<String> spanNamePredicate)
...
+     * modified in a way that replaces all subsequences that match {@code pattern} with the
+     * {@code replacement} string. The behavior of this method is consistent with the
+     * {@link Matcher#replaceAll(String)} method.
+     *
+     * @param pattern     A regular expression pattern that matches {@code http.url} attribute
+     *                    values that should undergo replacement.
+     * @param replacement The replacement string.
+     * @return {@code this}.
+     */
+    public <T> SpanFilterBuilder replaceHttpUrl(Pattern pattern, String replacement) {
```
jkwatson, line 154: The `pattern` and the `replacement` here need to be quite carefully crafted to work. I think we should at *least* put an example in the javadoc of what would work, otherwise I think that users will have a hard time figuring it out without a lot of digging.

jkwatson:
I go back and forth on whether we should be providing these 3 helpers, or whether we should see what users do with the existing method and see where we can help them do it more easily when/if they run into difficulty. The bigger the API we ship, the more API we have to support.

Are these easy enough for Android developers to implement for themselves that we could live without them for now?

mateuszrzeszutek:
> Are these easy enough for Android developers to implement for themselves that we could live without them for now?

I'd say they're pretty easy, they're just a few lines of code - and it's mostly trivial code; anyone who has ever used `Pattern` could implement them. We can close this PR for now and wait a bit for the feedback - and reopen it if there's a need for more specific filtering capabilities.

jkwatson:
> > Are these easy enough for Android developers to implement for themselves that we could live without them for now?
> 
> I'd say they're pretty easy, they're just a few lines of code - and it's mostly trivial code; anyone who has ever used `Pattern` could implement them. We can close this PR for now and wait a bit for the feedback - and reopen it if there's a need for more specific filtering capabilities.

One thing that this PR does provide, though, is removing the requirement that user know the semantic convention attribute where the url is stored. I wonder if it would be helpful to put a reference to it in the StandardAttributes class, so it's easier to find.

mateuszrzeszutek:
> I wonder if it would be helpful to put a reference to it in the StandardAttributes class, so it's easier to find.

I think it would, the users of this lib wouldn't have to include an additional dependency on `opentelemetry-semconv` to use that attribute. Should we do just that instead of adding dedicated `http.url` methods?

jkwatson:
> > I wonder if it would be helpful to put a reference to it in the StandardAttributes class, so it's easier to find.
> 
> I think it would, the users of this lib wouldn't have to include an additional dependency on `opentelemetry-semconv` to use that attribute. Should we do just that instead of adding dedicated `http.url` methods?

yeah, let's start with that and get some user feedback before anything more significant.

mateuszrzeszutek:
Got it - I'll close this PR then and open a new one just for `http.url`

PR #145, opened 2021-09-07
Fix ConnectionUtil/ConnectionManager NPE

mateuszrzeszutek:


splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
```diff
@@ -35,35 +34,40 @@
     static final CurrentNetwork NO_NETWORK = new CurrentNetwork(NetworkState.NO_NETWORK_AVAILABLE, null);
     static final CurrentNetwork UNKNOWN_NETWORK = new CurrentNetwork(NetworkState.TRANSPORT_UNKNOWN, null);
 
-    private final ConnectionMonitor connectionMonitor;
+    private final Supplier<NetworkRequest> createNetworkMonitoringRequest;
     private final NetworkDetector networkDetector;
+    private final ConnectivityManager connectivityManager;
 
-    private final AtomicReference<CurrentNetwork> currentNetwork = new AtomicReference<>();
+    private volatile CurrentNetwork currentNetwork;
```
mateuszrzeszutek, line 41: We don't use CAS anywhere, so I replaced an `AtomicReference` with a `volatile` field

splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
```diff
@@ -35,35 +34,40 @@
...
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
-            connectivityManager.registerDefaultNetworkCallback(connectionMonitor);
-        } else {
-            NetworkRequest networkRequest = createNetworkMonitoringRequest.get();
-            connectivityManager.registerNetworkCallback(networkRequest, connectionMonitor);
-        }
+        this.createNetworkMonitoringRequest = createNetworkMonitoringRequest;
         this.networkDetector = networkDetector;
+        this.connectivityManager = connectivityManager;
         refreshNetworkStatus();
```
jkwatson, line 53: I think this call should be moved down to `start`. See my implementation in #146  :)

splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
```diff
@@ -35,35 +34,40 @@
...
     }
 
     CurrentNetwork refreshNetworkStatus() {
         CurrentNetwork activeNetwork = networkDetector.detectCurrentNetwork();
-        currentNetwork.set(activeNetwork);
+        currentNetwork = activeNetwork;
         return activeNetwork;
     }
 
+    void start() {
```
jkwatson, line 62: maybe rename to "startMonitoring"

splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
jkwatson: also note that you don't need to keep references to them if they just get passed into the `start` method with my approach.

splunk-otel-android/src/test/java/com/splunk/rum/ConnectionUtilTest.java
```diff
@@ -115,4 +121,40 @@ public void quiznos() {
         monitorCaptor.getValue().onLost(null);
         assertEquals(2, notified.get());
     }
+
+    @Test
+    @Config(sdk = Build.VERSION_CODES.LOLLIPOP)
```
jkwatson, line 126: I discovered that you can give this an array of `sdk` entries to collapse to a single test.

mateuszrzeszutek: That won't work in this case unfortunately: https://github.com/signalfx/splunk-otel-android/pull/146#discussion_r704265207

jkwatson (approved):
Ship it!

PR #146, opened 2021-09-07
Fix an NPE that could occur in the NetworkMonitor.

jkwatson:
This was because of badly organized initialization order that leaked an instance before the constructor was done executing.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
@@ -16,7 +16,6 @@
 
 package com.splunk.rum;
 
-import android.content.Context;
 import android.net.ConnectivityManager;
 import android.net.Network;
 import android.net.NetworkCapabilities;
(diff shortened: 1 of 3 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
```diff
@@ -40,22 +39,19 @@
...
 
-    ConnectionUtil(Context context) {
-        this(ConnectionUtil::createNetworkMonitoringRequest, NetworkDetector.create(context), (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE));
-    }
-
-    //for testing, since building the NetworkRequest fails in junit due to .. Android.
-    ConnectionUtil(Supplier<NetworkRequest> createNetworkMonitoringRequest, NetworkDetector networkDetector, ConnectivityManager connectivityManager) {
+    ConnectionUtil(NetworkDetector networkDetector) {
+        this.networkDetector = networkDetector;
         this.connectionMonitor = new ConnectionMonitor();
```
breedx-splk, line 44: This could be moved to the declaration.

breedx-splk (approved):
Definitely seems better not to do work in the constructor. :) There's still a minor DI violation with the ConnectionMonitor, but not a huge deal.

splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
jkwatson: well... then we'd be leaking ourselves even *before* the constructor, since `ConnectionMonitor` is not a static class.

mateuszrzeszutek: Or you could move the `ConnectionStateListener` field to `ConnectionUtil` and not store `ConnectionMonitor` in a field at all - simply create a new instance in `startMonitoring()` and pass it to the `ConnectivityManager`. You'll avoid the dependency cycle between those two classes this way.

splunk-otel-android/src/test/java/com/splunk/rum/ConnectionUtilTest.java
```diff
@@ -115,4 +121,22 @@ public void quiznos() {
...
+        NetworkRequest networkRequest = mock(NetworkRequest.class);
+        NetworkDetector networkDetector = mock(NetworkDetector.class);
+        ConnectivityManager connectivityManager = mock(ConnectivityManager.class);
+
+        doAnswer(invocation -> {
+            ConnectivityManager.NetworkCallback callback = invocation.getArgument(1);
+            callback.onAvailable(mock(Network.class));
+            return null;
+        }).when(connectivityManager)
+                .registerNetworkCallback(eq(networkRequest), any(ConnectivityManager.NetworkCallback.class));
```
mateuszrzeszutek, line 137: This test doesn't verify the correct behavior on Q though: the `registerDefaultNetworkCallback()` method is called instead of `registerNetworkCallback()`, so this whole stubbing won't execute on Q and the test will always pass.

jkwatson: Drat you are correct

jkwatson:
closing as solved in #145

PR #148, opened 2021-09-09
Create a span on sessionId change

mateuszrzeszutek:


splunk-otel-android/src/main/java/com/splunk/rum/SessionIdChangeTracer.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final Tracer tracer;
+
+    SessionIdChangeTracer(Tracer tracer) {
+        this.tracer = tracer;
+    }
+
+    @Override
+    public void onChange(String oldSessionId, String newSessionId) {
+        tracer.spanBuilder("sessionId.change")
```
mateuszrzeszutek, line 36: Is there any guidance for the span name for this? I used `sessionId.change` because it's similar to `network.change`

splunk-otel-android/src/main/java/com/splunk/rum/SessionIdChangeTracer.java
```diff
@@ -0,0 +1,43 @@
...
+                // TODO: new component for sessionId changes?
+                .setAttribute(SplunkRum.COMPONENT_KEY, "session")
```
mateuszrzeszutek, line 38: Should we have a separate component type for this?

splunk-otel-android/src/main/java/com/splunk/rum/SessionIdChangeTracer.java
```diff
@@ -0,0 +1,43 @@
...
+package com.splunk.rum;
+
+import static io.opentelemetry.api.common.AttributeKey.stringKey;
+
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.trace.Tracer;
+
+final class SessionIdChangeTracer implements SessionIdChangeListener {
+
+    static final AttributeKey<String> PREVIOUS_SESSION_ID_KEY = stringKey("splunk.previous.rumSessionId");
```
mateuszrzeszutek, line 26: Ideally I'd like to name it `splunk.rum.previous_session_id`, but it wouldn't match `splunk.rumSessionId` (which _ideally_ should be named `splunk.rum.session_id`...) so I just put together something like this. I'm not really convinced this is a good name though.

splunk-otel-android/src/main/java/com/splunk/rum/SessionId.java
```diff
@@ -55,12 +56,19 @@ String getSessionId() {
             //if this returns false, then another thread updated the value already.
```
mateuszrzeszutek, line 56: There's a 
```java
if (currentValue == null || sessionExpired()) {
```
condition a couple of lines above - @jkwatson is it really possible for `currentValue` to be null? The value gets initialized in the constructor, it shouldn't ever return null

jkwatson: I agree. That is probably some extra guarding that isn't actually necessary.

splunk-otel-android/src/main/java/com/splunk/rum/SessionIdChangeTracer.java
jkwatson: Since Android is the first to implement this feature, I think we can apply some leverage to the backend team to use what we like, and ask iOS/web to conform if they agree.

splunk-otel-android/src/main/java/com/splunk/rum/SessionIdChangeTracer.java
jkwatson: I might leave it out until we get clarity from the backend/frontend about what they want to see for this.

splunk-otel-android/src/test/java/com/splunk/rum/SessionIdChangeTracerTest.java
```diff
@@ -0,0 +1,59 @@
...
+    private SessionIdChangeListener underTest;
+
+    @Before
+    public void setup() {
+        Tracer tracer = otelTesting.getOpenTelemetry().getTracer("testTracer");
+        underTest = new SessionIdChangeTracer(tracer);
+    }
+
+    @Test
+    public void shouldEmitSessionIdChangeSpan() {
```
jkwatson, line 46: can you add a test to verify that nothing bad happens if there is no listener hooked up?

jkwatson: whoops. this comment should be on the SessionIdTest.

jkwatson: never mind. I don't think there is anything new needed.

jkwatson (approved):
Let's go with this, and we'll adjust if the naming changes.

PR #149, opened 2021-09-09
Expose http.url attribute in StandardAttributes

mateuszrzeszutek:


splunk-otel-android/src/main/java/com/splunk/rum/StandardAttributes.java
```diff
@@ -31,6 +32,14 @@
      */
     public static final AttributeKey<String> APP_VERSION = AttributeKey.stringKey("app.version");
 
+    /**
+     * Full HTTP request URL in the form {@code scheme://host[:port]/path?query[#fragment]}. Useful
```
jkwatson, line 36:
```suggestion
     * Full HTTP client request URL in the form {@code scheme://host[:port]/path?query[#fragment]}. Useful
```

jkwatson (approved):

PR #156, opened 2021-09-21
add a CI build for merges to main

jkwatson:


mateuszrzeszutek (approved):

.github/workflows/main.yaml
```diff
@@ -0,0 +1,20 @@
+name: "Continuous Build"
+
+on:
```
mateuszrzeszutek, line 3: You might also want to add `workflow_dispatch` to be able to run the build manually:
```suggestion
on:
  workflow_dispatch:
```

.github/workflows/main.yaml
```diff
@@ -0,0 +1,20 @@
...
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
mateuszrzeszutek, line 10: Wait, shouldn't we use concrete versions?
```suggestion
    runs-on: ubuntu-20.04
```

jkwatson: hm. we're using latest for our pr builds....this was just a copy of that. I'll update both.

mateuszrzeszutek (approved):

PR #158, opened 2021-09-24
Provide a method on SplunkRum for integrating with Browser RUM javascript

jkwatson:


johnbley:
LGTM - I agree with the choice of names - `integrateWithBrowserRum` seems fine, and I particularly like (from a browser point of view) the redundancy in `SplunkRumNative.getNativeSessionId` since there are so many kinds of session IDs floating around.

mateuszrzeszutek (approved):

PR #160, opened 2021-09-27
Add try/catch around trying to access the connectivityManager APIs.

jkwatson:
See https://issuetracker.google.com/issues/175055271 for the root Android bug.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/PostApi29NetworkDetector.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/PostApi29NetworkDetector.java
@@ -16,6 +16,9 @@
 
 package com.splunk.rum;
 
+import static com.splunk.rum.ConnectionUtil.NO_NETWORK;
+import static com.splunk.rum.ConnectionUtil.UNKNOWN_NETWORK;
+
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.content.Context;
@@ -27,9 +30,6 @@ import android.telephony.TelephonyManager;
 
 import androidx.annotation.RequiresApi;
 import androidx.core.app.ActivityCompat;
-
-import static com.splunk.rum.ConnectionUtil.NO_NETWORK;
-import static com.splunk.rum.ConnectionUtil.UNKNOWN_NETWORK;
 
 @RequiresApi(api = Build.VERSION_CODES.Q)
 class PostApi29NetworkDetector implements NetworkDetector {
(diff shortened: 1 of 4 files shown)


splunk-otel-android/src/test/java/com/splunk/rum/PostApi29NetworkDetectorTest.java
```diff
@@ -51,6 +51,21 @@ public void none() {
...
 
+    @Test
+    public void exception() {
+        ConnectivityManager connectivityManager = mock(ConnectivityManager.class);
+        TelephonyManager telephonyManager = mock(TelephonyManager.class);
+        Context context = mock(Context.class);
+
+        Network network = mock(Network.class);
+        when(connectivityManager.getActiveNetwork()).thenReturn(network);
+        when(connectivityManager.getNetworkCapabilities(network)).thenThrow(new SecurityException("bug"));
```
breedx-splk, line 62: 👍🏻

splunk-otel-android/src/test/java/com/splunk/rum/SimpleNetworkDetectorTest.java
```diff
@@ -50,6 +50,19 @@ public void none() {
         assertEquals(new CurrentNetwork(NetworkState.NO_NETWORK_AVAILABLE, null), currentNetwork);
     }
 
+    @Test
+    public void exception() {
+        ConnectivityManager connectivityManager =
+                mock(ConnectivityManager.class);
+
+        when(connectivityManager.getActiveNetworkInfo()).thenThrow(new SecurityException("bug"));
```
breedx-splk, line 58: 👍🏻

splunk-otel-android/src/main/java/com/splunk/rum/SimpleNetworkDetector.java
```diff
@@ -31,20 +31,26 @@
...
+            NetworkInfo activeNetwork = connectivityManager.getActiveNetworkInfo(); // Deprecated in API 29
+            if (activeNetwork == null) {
+                return NO_NETWORK;
+            }
+            switch (activeNetwork.getType()) {
+                case ConnectivityManager.TYPE_MOBILE:  // Deprecated in API 28
+                    return new CurrentNetwork(NetworkState.TRANSPORT_CELLULAR, activeNetwork.getSubtypeName());
+                case ConnectivityManager.TYPE_WIFI:  // Deprecated in API 28
+                    return new CurrentNetwork(NetworkState.TRANSPORT_WIFI, activeNetwork.getSubtypeName());
+                case ConnectivityManager.TYPE_VPN:
```
breedx-splk, line 44: Outside the scope of this specific change, but static imports on these would be nice for readability.

splunk-otel-android/src/main/java/com/splunk/rum/PostApi29NetworkDetector.java
```diff
@@ -46,24 +46,30 @@
...
-        if (capabilities.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR)) {
-            // If the app has the permission, use it to get a subtype.
-            if (hasPermission(Manifest.permission.READ_PHONE_STATE)) {
-                subType = getDataNetworkTypeName(telephonyManager.getDataNetworkType());
+        try {
+            NetworkCapabilities capabilities = connectivityManager.getNetworkCapabilities(connectivityManager.getActiveNetwork());
+            if (capabilities == null) {
+                return NO_NETWORK;
+            }
+            String subType = null;
```
breedx-splk, line 54: Unrelated to this change, but looks like the `subType` should be moved into the `if` scope block.

breedx-splk (approved):
Looks good to me. I read in a book one time that a method should only do one thing, and that exception handling is also a thing.  I think that means you might have cleaner code if you factored out the `try` body to a method...but dealer's choice.

splunk-otel-android/src/main/java/com/splunk/rum/SimpleNetworkDetector.java
jkwatson: I think when I wrote this originally, I wanted to make sure it was clear which of these things were Android's, and which were ours, and keeping the class name prefixes made that a little more clear to me.

jkwatson:
> Looks good to me. I read in a book one time that a method should only do one thing, and that exception handling is also a thing. I think that means you might have cleaner code if you factored out the `try` body to a method...but dealer's choice.

Heh. That method would have to be named something silly, though, wouldn't it? Like "doDetectCurrentNetwork"? or "detectCurrentNetworkUnlessAndroidIsBuggy"? :)

jkwatson:
> > Looks good to me. I read in a book one time that a method should only do one thing, and that exception handling is also a thing. I think that means you might have cleaner code if you factored out the `try` body to a method...but dealer's choice.
> 
> Heh. That method would have to be named something silly, though, wouldn't it? Like "doDetectCurrentNetwork"? or "detectCurrentNetworkUnlessAndroidIsBuggy"? :)

I guess the other option (probably better) would be to move the exception handling up a level into the `ConnectionUtil` call.  I can make that change if you'd like.

splunk-otel-android/src/main/java/com/splunk/rum/PostApi29NetworkDetector.java
```diff
@@ -16,6 +16,9 @@
 
 package com.splunk.rum;
 
+import static com.splunk.rum.ConnectionUtil.NO_NETWORK;
+import static com.splunk.rum.ConnectionUtil.UNKNOWN_NETWORK;
```
jkwatson, line 20: no idea why IDEA decided to reorder these imports. 🤷🏽

breedx-splk:
> > > Looks good to me. I read in a book one time that a method should only do one thing, and that exception handling is also a thing. I think that means you might have cleaner code if you factored out the `try` body to a method...but dealer's choice.
> > 
> > 
> > Heh. That method would have to be named something silly, though, wouldn't it? Like "doDetectCurrentNetwork"? or "detectCurrentNetworkUnlessAndroidIsBuggy"? :)
> 
> I guess the other option (probably better) would be to move the exception handling up a level into the `ConnectionUtil` call. I can make that change if you'd like.

I think it's an improvement for sure. I would have gotten rid of the local variable there tho.

jkwatson:
> I think it's an improvement for sure. I would have gotten rid of the local variable there tho.
oh yeah, that's totally a good change. I had missed altogether that it wasn't even necessary.

PR #162, opened 2021-09-28
Add a try/catch around registering network callbacks.

jkwatson:
Looks like this is another thing that can fail with the same Android bug (https://issuetracker.google.com/issues/175055271 for reference)

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
@@ -48,6 +48,7 @@ class ConnectionUtil {
             registerNetworkCallbacks(createNetworkMonitoringRequest, connectivityManager);
         } catch (Exception e) {
             //if this fails, we'll go without network change events.
+            Log.w(SplunkRum.LOG_TAG, "Failed to register network callbacks. Automatic network monitoring is disabled.", e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
```diff
@@ -44,6 +44,14 @@
 
     void startMonitoring(Supplier<NetworkRequest> createNetworkMonitoringRequest, ConnectivityManager connectivityManager) {
         refreshNetworkStatus();
+        try {
+            registerNetworkCallbacks(createNetworkMonitoringRequest, connectivityManager);
+        } catch (Exception e) {
+            //if this fails, we'll go without network change events.
```
breedx-splk, line 50: Worth logging or making an event or anything?

breedx-splk (approved):
LGTM, one question about logging, but not a big deal.

splunk-otel-android/src/main/java/com/splunk/rum/ConnectionUtil.java
jkwatson: I'm honestly not sure. This is a rare issue, and it happens before the library is initialized, and I've never seen it locally..so...long story short, I don't know what might be useful, and where to put it.

jkwatson: I added a warning log. Hopefully if anyone sees it, they will let us know

PR #167, opened 2021-10-05
Update to the latest sdk version

jkwatson:
And make sure version-specific tests are running on more versions


breedx-splk (approved):
I will miss quiznos but LGTM

PR #198, opened 2021-12-01
Apply GDI spec 1.2 to Android RUM

mateuszrzeszutek:
I'm not adding the badge yet because of the [`os.type`/`os.name`](https://github.com/signalfx/gdi-specification/pull/155) bug in the spec; but aside from that this repo implements the 1.2 spec.


mateuszrzeszutek:
@jkwatson please take a look at this PR - does RUM backend depend on the version attribute naming?

jkwatson:
We don't index on that key, so as far as I know, that's a fine thing to change.

breedx-splk (approved):
Story checks out. 😉

PR #209, opened 2022-01-18
Volley HurlStack instrumentation

Enkelian:
First part of [this](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3996) otel issue.

Due to the incompatibility of android and java gradle plugins, Volley instrumentation needs to be placed in this repo for now.


jkwatson:
I think this should be a separate library that users can add, rather than adding this dependency to all of our customers apps.

mateuszrzeszutek:
> I think this should be a separate library that users can add, rather than adding this dependency to all of our customers apps.

That was the first idea, to put this instrumentation into one of the OTel repos - but the gradle setup turned out to be so painful that we've decided to put it here for now, as the quicker option. (we still want to upstream that, only later)

Also, suppose that it is in a separate library - don't we want to add volley-instrumenting methods to `SplunkRum`? Like we've did with OkHTTP.

jkwatson:
> That was the first idea, to put this instrumentation into one of the OTel repos - but the gradle setup turned out to be so painful that we've decided to put it here for now, as the quicker option. (we still want to upstream that, only later)

It could be a separate library published out of this repo, though.

> 
> Also, suppose that it is in a separate library - don't we want to add volley-instrumenting methods to `SplunkRum`? Like we did with OkHTTP.

That's a good question. My gut tells me that okhttp is probably 100x more popular than volley at this point. We could always pull it in to the main API if it gains in popularity. Would we lose functionality if we kept it separate? Could that functionality be provided some other way?

mateuszrzeszutek:
> It could be a separate library published out of this repo, though.

That was also an idea - I thought it would be simpler to just include it in the current lib (we won't have to publish anything new).
WDYT? Should we publish (and document) a separate library just for volley? Or keep a `compileOnly` dependency on volley and put the instrumentation code into the main package?

> Would we lose functionality if we kept it separate? Could that functionality be provided some other way?

No, I don't think that we lose anything. The only thing that differs here is the API - using `SplunkRum` vs using library instrumentation classes. Using `SplunkRum` has that advantage of it being the only class that you need to know to instrument your app (on the other hand, the number of its public methods keeps growing...)

jkwatson:
Certainly making it a separate library will make it easier to upstream once we have a place to put android-specific instrumentation. And, we can always make it a hard dependency later. My vote is definitely to start as a separate bit of "library" instrumentation and only integrate after we get some feedback from users.

splunk-otel-android/src/main/java/com/splunk/rum/volley/TracingHurlStack.java
```diff
@@ -0,0 +1,96 @@
...
+            response = super.executeRequest(request, requestWrapper.getAdditionalHeaders());
+        } catch (Throwable t) {
+            throwable = t;
+            throw t;
+        } finally {
+            if (scope != null) {
+                scope.close();
+                instrumenter.end(context, currentRequestWrapper.get(), response, throwable);
+            }
+        }
```
laurit, line 84: You probably should also clear `currentRequestWrapper` in the finally block

splunk-otel-android/src/main/java/com/splunk/rum/volley/VolleyHttpClientAttributesExtractor.java
```diff
@@ -0,0 +1,154 @@
...
+            case 4:
+                return "HEAD";
+            case 5:
+                return "OPTIONS";
+            case 6:
+                return "TRACE";
+            case 7:
+                return "PATCH";
+            default:
+                return "";
```
laurit, line 81: I believe you should be returning null when method is unknown

splunk-otel-android/src/main/java/com/splunk/rum/volley/VolleyHttpClientAttributesExtractor.java
```diff
@@ -0,0 +1,154 @@
...
+    protected String flavor(RequestWrapper requestWrapper, @Nullable HttpResponse response) {
+        return null;
+    }
+
+    @Override
+    protected String method(RequestWrapper requestWrapper) {
+        Request<?> request = requestWrapper.getRequest();
+        switch (request.getMethod()) {
+            case -1:
+                try {
```
laurit, line 55: I would probably just return `null` in this case. It feels like calling `getPostBody` could have side effects, perhaps it is too risky to call?

splunk-otel-android/src/main/java/com/splunk/rum/volley/VolleyHttpClientAttributesExtractor.java
```diff
@@ -0,0 +1,154 @@
...
+                    if (request.getPostBody() != null) {
+                        return "POST";
+                    } else {
+                        return "GET";
+                    }
+                } catch (AuthFailureError authFailureError) {
+                    return "GET_OR_POST";
```
laurit, line 62: I believe this method may return only valid http method names.

splunk-otel-android/src/main/java/com/splunk/rum/volley/VolleyNetClientAttributesExtractor.java
```diff
@@ -0,0 +1,47 @@
...
+final class VolleyNetClientAttributesExtractor extends NetClientAttributesExtractor<RequestWrapper, HttpResponse> {
+
+    @Override
+    public String transport(RequestWrapper requestWrapper, @Nullable HttpResponse httpResponse) {
+        return SemanticAttributes.NetTransportValues.IP_TCP;
+    }
+
+    @Override
+    public String peerName(RequestWrapper requestWrapper, @Nullable HttpResponse httpResponse) {
+        return requestWrapper.getRequest().getUrl();
```
laurit, line 35: probably should return host here

splunk-otel-android/src/main/java/com/splunk/rum/volley/VolleyNetClientAttributesExtractor.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public Integer peerPort(RequestWrapper requestWrapper, @Nullable HttpResponse httpResponse) {
+        return requestWrapper.getUrl().getPort();
+    }
+
+    @Override
+    public String peerIp(RequestWrapper requestWrapper, @Nullable HttpResponse httpResponse) {
+        return requestWrapper.getUrl().getHost();
```
laurit, line 45: this should return null

splunk-otel-android/src/test/java/com/splunk/rum/volley/TracingHurlStackTest.java
```diff
@@ -0,0 +1,274 @@
...
+        SpanData firstSpan = spans.get(0);
+        testsAttributes(firstSpan, url, 200);
+
+        SpanData secondSpan = spans.get(1);
+        testsAttributes(secondSpan, url, 200);
+    }
+
+    //TODO: concurrent tests
+
+    private void testsAttributes(SpanData span, String url, int status) {
```
mateuszrzeszutek, line 229: Nit: how about `verifyAttributes` or `assertAttributes`?

splunk-otel-android/build.gradle
```diff
@@ -42,6 +42,7 @@ android {
 dependencies {
     implementation 'androidx.appcompat:appcompat:1.3.1'
     implementation 'androidx.navigation:navigation-fragment:2.3.5'
+    implementation 'com.android.volley:volley:1.2.0'
```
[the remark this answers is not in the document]
Enkelian: `implementation` is not necessary, changing to `compileOnly`, thanks!

splunk-otel-android/build.gradle
```diff
@@ -55,12 +56,17 @@ dependencies {
     api 'io.opentelemetry:opentelemetry-api'
     api 'com.squareup.okhttp3:okhttp:4.9.3'
 
+    implementation platform("io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:1.10.0-alpha")
+    implementation 'io.opentelemetry.javaagent:opentelemetry-javaagent-instrumentation-api'
```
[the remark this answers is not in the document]
Enkelian: Forgot to remove this, thanks for catching it!

splunk-otel-android/src/main/java/com/splunk/rum/volley/VolleyHttpClientAttributesExtractor.java
```diff
@@ -0,0 +1,154 @@
...
+                }
+            case 0:
```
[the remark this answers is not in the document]
Enkelian: Sure!

splunk-otel-android/src/main/java/com/splunk/rum/volley/TracingHurlStack.java
```diff
@@ -0,0 +1,96 @@
...
+
+        return response;
+
+    }
+
+    @Override
+    protected HttpURLConnection createConnection(URL url) throws IOException {
+        RequestWrapper requestWrapper = currentRequestWrapper.get();
```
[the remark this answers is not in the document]
Enkelian: It shouldn't be - `createConnection` is called only inside parent's `executeRequest` (through a private method - `openConnection`), so `currentRequestWrapper.set()` is always called before that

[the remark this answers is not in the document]
mateuszrzeszutek: Nit: can you add a comment with that explanation just above this line?

splunk-otel-android/src/main/java/com/splunk/rum/volley/VolleyHttpClientAttributesExtractor.java
```diff
@@ -0,0 +1,144 @@
...
+        return header != null ? Collections.singletonList(header) : Collections.emptyList();
+
+    }
+
+    @Nullable
+    @Override
+    protected Long requestContentLength(RequestWrapper requestWrapper, @Nullable HttpResponse response) {
+        Request<?> request = requestWrapper.getRequest();
+        try {
+            return request.getBody() != null ? (long) requestWrapper.getRequest().getBody().length : null;
```
mateuszrzeszutek, line 97: No need to go through the wrapper again
```suggestion
            return request.getBody() != null ? (long) request.getBody().length : null;
```

splunk-otel-android/src/test/java/com/splunk/rum/volley/TracingHurlStackTest.java
```diff
@@ -0,0 +1,274 @@
...
+        SdkTracerProvider sdkTracerProvider = SdkTracerProvider.builder()
+                .addSpanProcessor(SimpleSpanProcessor.create(exporter))
+                .build();
+
+        GlobalOpenTelemetry.resetForTest();
+
+        OpenTelemetry otel = OpenTelemetrySdk.builder()
+                .setTracerProvider(sdkTracerProvider)
+                .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
+                .buildAndRegisterGlobal();
```
mateuszrzeszutek, line 89: You could just call `build()` and remove all the global usages from this tests - it's not used in the main `splunk-otel-android` lib at all, so the volley lib should not use it too.

splunk-otel-android/src/test/java/com/splunk/rum/volley/TracingHurlStackTest.java
```diff
@@ -0,0 +1,274 @@
...
+    //TODO: concurrent tests
+
+    private void verifyAttributes(SpanData span, String url, int status) {
+        assertThat(span.getName()).isEqualTo("HTTP GET");
+
+        Attributes spanAttributes = span.getAttributes();
+        assertThat(spanAttributes.get(SemanticAttributes.HTTP_STATUS_CODE)).isEqualTo(status);
+        assertThat(spanAttributes.get(SemanticAttributes.NET_PEER_PORT)).isEqualTo(server.getPort());
+        assertThat(spanAttributes.get(SemanticAttributes.NET_PEER_NAME)).isEqualTo(server.getHostName());
+        assertThat(spanAttributes.get(SemanticAttributes.HTTP_URL)).isEqualTo(url);
```
mateuszrzeszutek, line 236: Also
```suggestion
        assertThat(spanAttributes.get(SemanticAttributes.HTTP_URL)).isEqualTo(url);
        assertThat(spanAttributes.get(SemanticAttributes.HTTP_METHOD)).isEqualTo("GET");
```

splunk-otel-android/src/test/java/com/splunk/rum/volley/TracingHurlStackTest.java
```diff
@@ -0,0 +1,274 @@
...
+        assertThat(spans).hasSize(1);
+
+        SpanData span = spans.get(0);
+        assertThat(span.getName()).isEqualTo("HTTP GET");
+
+        assertThat(span.getStatus()).isEqualTo(StatusData.error());
+
+        EventData event = span.getEvents().stream().filter(e -> e.getName().equals(SemanticAttributes.EXCEPTION_EVENT_NAME)).findFirst().orElse(null);
+
+        assertThat(event).isNotNull();
```
mateuszrzeszutek, line 191: You could probably use some AssertJ idiom to assert on that:
```suggestion
        assertThat(span.getEvents())
            .hasSize(1)
            .allSatisfy(e -> e.getName().equals(SemanticAttributes.EXCEPTION_EVENT_NAME))
```

splunk-otel-android/src/test/java/com/splunk/rum/volley/TracingHurlStackTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+        assertThatThrownBy(() -> response.get(3, TimeUnit.SECONDS)).hasCauseInstanceOf(TimeoutError.class);
+
+        assertThat(server.getRequestCount()).isEqualTo(0); //server received no requests
+
+        List<SpanData> spans = exporter.getFinishedSpanItems();
+        assertThat(spans).hasSize(1);
+
+        SpanData span = spans.get(0);
+        assertThat(span.getName()).isEqualTo("HTTP GET");
```
mateuszrzeszutek, line 185: What about other span attributes? E.g. URL -- do they get filled in?

mateuszrzeszutek:
@jkwatson the story does not mention that, but do we need to add the same RUM-specific features as in the OkHttp instrumentation? Parsing `Server-Timing` header, adding error type & message as span attributes, etc.
(Of course, implemented as a separate story/task/unit of work)

splunk-otel-android/src/test/java/com/splunk/rum/volley/TracingHurlStackTest.java
Enkelian: Yeah, they should definitely be checked here as well - I'll add a `verifyAttributes()` call

jkwatson:
> @jkwatson the story does not mention that, but do we need to add the same RUM-specific features as in the OkHttp instrumentation? Parsing `Server-Timing` header, adding error type & message as span attributes, etc. (Of course, implemented as a separate story/task/unit of work)

I think this is a question for PM, not for me. :) My opinion is: of course, because the integration here needs to work for our customers.

splunk-otel-android-volley/build.gradle.kts
```diff
@@ -0,0 +1,166 @@
...
+    compileOnly("com.android.volley:volley:1.2.0")
+
+    api(platform("io.opentelemetry:opentelemetry-bom:1.10.0"))
+    implementation("io.opentelemetry:opentelemetry-sdk")
+
+    implementation(platform("io.opentelemetry:opentelemetry-bom-alpha:1.10.0-alpha"))
+    implementation("io.opentelemetry:opentelemetry-semconv")
+
+    implementation(platform("io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:1.10.0-alpha"))
+    implementation("io.opentelemetry.javaagent:opentelemetry-javaagent-instrumentation-api")
```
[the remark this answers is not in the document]
mateuszrzeszutek: Hmm, you should be using `io.opentelemetry.instrumentation:opentelemetry-instrumentation-api` instead

splunk-otel-android-volley/src/main/java/com/splunk/rum/volley/VolleyTracingBuilder.java
```diff
@@ -0,0 +1,97 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.http.CapturedHttpHeaders;
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpSpanNameExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpSpanStatusExtractor;
+
+/**
+ * A builder for {@link VolleyTracing}.
+ */
+public final class VolleyTracingBuilder {
+
+    private static final String INSTRUMENTATION_NAME = "io.opentelemetry.volley";
```
mateuszrzeszutek, line 39: Oh, I totally missed that one -- it probably should be something like
```suggestion
    private static final String INSTRUMENTATION_NAME = "com.splunk.android.volley";
```
cause it's not OTel instrumentation; it's our (Splunk) custom instrumentation.

splunk-otel-android-volley/src/test/java/com/splunk/rum/volley/TracingHurlStackTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+        //setup OpenTelemetry
+        exporter = InMemorySpanExporter.create();
+
+        SdkTracerProvider sdkTracerProvider = SdkTracerProvider.builder()
+                .addSpanProcessor(SimpleSpanProcessor.create(exporter))
+                .build();
+
+        GlobalOpenTelemetry.resetForTest();
+
```
mateuszrzeszutek, line 85: Shouldn't be needed anymore
```suggestion
```

mateuszrzeszutek (approved):
LGTM 👍

PR #210, opened 2022-01-20
Convert gradle files to kotlin

mateuszrzeszutek:
Because Groovy's just too painful to work with


sample-app/proguard-rules.pro
```diff
@@ -1,6 +1,6 @@
 # Add project specific ProGuard rules here.
 # You can control the set of applied configuration files using the
-# proguardFiles setting in build.gradle.
+# proguardFiles setting in build.gradle.kts.kts.
```
jkwatson, line 3:
```suggestion
# proguardFiles setting in build.gradle.kts.
```

breedx-splk (approved):
It also just makes sense for an android project. Nice job!

PR #221, opened 2022-01-27
Enable gradle enterprise

mateuszrzeszutek:


breedx-splk:
@mateuszrzeszutek what should we do with this?

mateuszrzeszutek:
Hmmmm, let's close it. I totally forgot I even tried to do that.

PR #226, opened 2022-02-02
Implement session timeout due to inactivity

mateuszrzeszutek:


splunk-otel-android/src/main/java/com/splunk/rum/SessionId.java
```diff
@@ -60,20 +64,39 @@ String getSessionId() {
...
+        if (timeoutStartNanos == NO_TIMEOUT) {
+            // we haven't started the inactivity timeout yet
+            return false;
+        }
+        long elapsedTime = clock.nanoTime() - timeoutStartNanos;
+        return elapsedTime >= SESSION_TIMEOUT_NANOS;
+    }
+
+    private void resetTimeout() {
+        timeoutStartNanos = NO_TIMEOUT;
```
jkwatson, line 93: it seems like it would be simplest just to keep track of a "last accessed time" that updates whenever anyone asks for a sessionId, and just check for how old it is, rather than having this reset logic in addition.

mateuszrzeszutek: It is possible for an app to be in the foreground and not send any spans for >15 minutes, isn't it? In that case I'd still need another variable (a boolean) that'd be set whenever the app gets backgrounded.

jkwatson: I'm not sure how you're handling that now, though? Seems like you're only checking for timeout when the sessionId is accessed already. My idea would just remove one extra bit to check.

mateuszrzeszutek: Hm, but the session is not supposed to time out when the app is active, in the foreground - in that case it's supposed to last full 4 hours (even if the app doesn't send any span for 15 minutes). The timeout should only be applied when the app gets backgrounded.

jkwatson: I don't know how we can possibly track that an app is "active" if no activity is being observed by the instrumentation. The best we can do (as far as I know) is to just track "liveness" by the generation of spans. If no traceable activity happens for 15 minutes, could we not consider that as a new session?

jkwatson: If an app developer wants to have instrumentation-idle apps to have extended sessions, we could recommend that they generate some sort of "ping" event every few minutes to keep it alive, I guess?

mateuszrzeszutek: @jkwatson PTAL at this PR now - I've talked to Seerut about the background/foreground timeout requirements, and reimplemented the whole thing.

jkwatson: ah, cool. As long as our foreground/background detection is reliable, then, 👍🏽

splunk-otel-android/src/main/java/com/splunk/rum/SessionId.java
```diff
@@ -50,30 +55,39 @@ private static String createNewId() {
...
+        // sessionId change listener needs tobe called after bumping the timer because it may create a new span
+        if (sessionIdChanged && sessionIdChangeListener != null) {
+            sessionIdChangeListener.onChange(oldValue, currentValue);
+        }
+
+        return currentValue;
     }
 
     private boolean sessionExpired() {
+        // TODO: it probably should use nanoTime(); now() javadoc explicitly states that it's not meant to be used to compute duration
```
breedx-splk, line 82: Do we want this TODO to be addressed in this PR?

mateuszrzeszutek:
> As I read it, a backgrounded app that doesn't change state and doesn't do anything would technically keep its state open until it finally changes state or decides to do something with the session id.

Yep -- and in either case, the next span generation will call `getSessionId()` which in turn will trigger the timeout mechanism.

> Do we need to proactively close the session after a given time and report that?

By "proactively close the session" you mean forcing the sessionId to regenerate after every 15 minutes of no activity in the background? Isn't just waiting for the application to do something and then reacting a bit simpler?

PR #228, opened 2022-02-07
Centralize publishing configuration

Enkelian:


buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,12 @@
...
+    `kotlin-dsl`
+}
+
+repositories {
+    mavenCentral()
+    google()
+}
+
+dependencies {
+    implementation("com.android.tools.build:gradle:7.1.0")
```
mateuszrzeszutek, line 11:
```suggestion
    implementation("com.android.tools.build:gradle:7.1.1")
```

splunk-otel-android-volley/build.gradle.kts
```diff
@@ -1,10 +1,8 @@
-import java.net.URI
 import java.time.Duration
 
 plugins {
     id("com.android.library")
-    id("maven-publish")
-    id("signing")
+    id("splunk.publish-conventions")
```
mateuszrzeszutek, line 5: Nit: WDYT about calling it `splunk.android-library-conventions`? This plugin does a bit more than just setting up artifact publishing

buildSrc/src/main/kotlin/splunk.publish-conventions.gradle.kts
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+    publications {
+        register<MavenPublication>("maven") {
+            groupId = "com.splunk"
+            artifactId = base.archivesName.get()
+
+            pom {
+                name.set("Splunk Otel Android")
+                description.set("A library for instrumenting Android applications for Splunk RUM")
```
mateuszrzeszutek, line 29: Shouldn't we set these in the libraries' `build.gradle.kts` files? I think the volley lib could have a bit different description

mateuszrzeszutek (commented):
Left couple of minor comments, overall it LGTM 👍

buildSrc/src/main/kotlin/splunk.publish-conventions.gradle.kts
Enkelian: Done, although I wonder if I approached it correctly with setting the name through an `extra`. WDYT?

mateuszrzeszutek: I think it's fine; the upstream repo handles a similar use case in a similar way: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/conventions/src/main/kotlin/otel.javaagent-instrumentation.gradle.kts#L8

mateuszrzeszutek (approved):

PR #236, opened 2022-02-15
Slow rendering detector

breedx-splk:
This sets up a new component: SlowRenderingDetector. This is an interface, of which there are two implementations -- a noop version and the real version. The noop version is wired in if the platform doesn't support the API class we use or if the polling mechanism is set to 0. The real implementation will poll the `FrameMetricsAggregator` every second (by default) and look for renders that are slower than 16ms (slow render) and 700ms (frozen render).

These are then aggregated per interval (1s default) and ingested as zero-duration spans (which zipkin thinks are 1us spans). The span names are `slowRenders` or `frozenRenders` and the spans contain an attribute named "count" which is a sum of the renders that met the times shown above.

The demo app was also updated to have a self-painting component in the second fragment to show this. Its slow mode can be toggled on/off with the button.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
@@ -40,14 +40,16 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
     private final VisibleScreenTracker visibleScreenTracker;
     private final AppStartupTimer startupTimer;
     private final List<AppStateListener> appStateListeners;
+    private SlowRenderingDetector slowRenderingDetector;
     //we count the number of activities that have been "started" and not yet "stopped" here to figure out when the app goes into the background.
     private int numberOfOpenActivities = 0;
 
-    ActivityCallbacks(Tracer tracer, VisibleScreenTracker visibleScreenTracker, AppStartupTimer startupTimer, List<AppStateListener> appStateListeners) {
+    ActivityCallbacks(Tracer tracer, VisibleScreenTracker visibleScreenTracker, AppStartupTimer startupTimer, List<AppStateListener> appStateListeners, SlowRenderingDetector slowRenderingDetector) {
         this.tracer = tracer;
         this.visibleScreenTracker = visibleScreenTracker;
         this.startupTimer = startupTimer;
         this.appStateListeners = appStateListeners;
+        this.slowRenderingDetector = slowRenderingDetector;
     }
 
     @Override
@@ -77,6 +79,7 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
         getTracer(activity)
                 .initiateRestartSpanIfNecessary(tracersByActivityClassName.size() > 1)
                 .addEvent("activityPreStarted");
+        slowRenderingDetector.add(activity);
     }
 
     @Override
@@ -149,6 +152,7 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
             }
         }
         addEvent(activity, "activityStopped");
+        slowRenderingDetector.stop(activity);
     }
 
     @Override
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
@@ -40,7 +40,7 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
     private final VisibleScreenTracker visibleScreenTracker;
     private final AppStartupTimer startupTimer;
     private final List<AppStateListener> appStateListeners;
-    private SlowRenderingDetector slowRenderingDetector;
+    private final SlowRenderingDetector slowRenderingDetector;
     //we count the number of activities that have been "started" and not yet "stopped" here to figure out when the app goes into the background.
     private int numberOfOpenActivities = 0;
 
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
@@ -83,7 +83,6 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
         getTracer(activity)
                 .initiateRestartSpanIfNecessary(tracersByActivityClassName.size() > 1)
                 .addEvent("activityPreStarted");
-        slowRenderingDetector.add(activity);
     }
 
     @Override
@@ -112,6 +111,7 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
     @Override
     public void onActivityResumed(@NonNull Activity activity) {
         addEvent(activity, "activityResumed");
+        slowRenderingDetector.add(activity);
     }
 
     @Override
@@ -134,6 +134,7 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
     @Override
     public void onActivityPaused(@NonNull Activity activity) {
         addEvent(activity, "activityPaused");
+        slowRenderingDetector.stop(activity);
     }
 
     @Override
@@ -156,7 +157,6 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
             }
         }
         addEvent(activity, "activityStopped");
-        slowRenderingDetector.stop(activity);
     }
 
     @Override
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
@@ -44,12 +44,16 @@ class ActivityCallbacks implements Application.ActivityLifecycleCallbacks {
     //we count the number of activities that have been "started" and not yet "stopped" here to figure out when the app goes into the background.
     private int numberOfOpenActivities = 0;
 
-    ActivityCallbacks(Tracer tracer, VisibleScreenTracker visibleScreenTracker, AppStartupTimer startupTimer, List<AppStateListener> appStateListeners, SlowRenderingDetector slowRenderingDetector) {
-        this.tracer = tracer;
-        this.visibleScreenTracker = visibleScreenTracker;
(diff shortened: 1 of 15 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
```diff
@@ -40,14 +40,16 @@
     private final VisibleScreenTracker visibleScreenTracker;
     private final AppStartupTimer startupTimer;
     private final List<AppStateListener> appStateListeners;
+    private SlowRenderingDetector slowRenderingDetector;
```
jkwatson, line 43: no final?

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
```diff
@@ -266,6 +280,17 @@ public Builder anrDetectionEnabled(boolean enable) {
             return this;
         }
 
+        /**
+         * Configures the rate at which frame render durations are polled. Pass zero to
+         * disable polling for slow/frozen renders.
+         * @param interval - The period that should be used for polling (in ms)
+         * @return this
+         */
+        public Builder renderDurationPollingIntervalMs(int interval){
```
jkwatson, line 289: Rather than using a zero as a magic number here in the config, why not have an explicit way to enable/disable this feature?

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -120,6 +127,20 @@ SplunkRum initialize(Supplier<ConnectionUtil> connectionUtilSupplier, Looper mai
         return new SplunkRum(openTelemetrySdk, sessionId, config);
     }
 
+    private SlowRenderingDetector buildSlowRenderDetector(Config config, Tracer tracer) {
```
jkwatson, line 130: there's some broken parallel naming here. is it a "slow render detector" or a "slow rendering detector"?

splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
```diff
@@ -149,6 +152,7 @@ public void onActivityStopped(@NonNull Activity activity) {
             }
         }
         addEvent(activity, "activityStopped");
+        slowRenderingDetector.stop(activity);
```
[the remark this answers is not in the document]
breedx-splk: That's cool. For symmetry do you think we should also change start to `onResume()` then? Looks like those are the two state transitions immediately surrounding running state.

splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
```diff
@@ -40,14 +40,16 @@
     private final VisibleScreenTracker visibleScreenTracker;
     private final AppStartupTimer startupTimer;
     private final List<AppStateListener> appStateListeners;
+    private SlowRenderingDetector slowRenderingDetector;
     //we count the number of activities that have been "started" and not yet "stopped" here to figure out when the app goes into the background.
     private int numberOfOpenActivities = 0;
 
-    ActivityCallbacks(Tracer tracer, VisibleScreenTracker visibleScreenTracker, AppStartupTimer startupTimer, List<AppStateListener> appStateListeners) {
+    ActivityCallbacks(Tracer tracer, VisibleScreenTracker visibleScreenTracker, AppStartupTimer startupTimer, List<AppStateListener> appStateListeners, SlowRenderingDetector slowRenderingDetector) {
```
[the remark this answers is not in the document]
breedx-splk: Yeah, agreed. I can pick that up here.

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
breedx-splk: I think there's a well established common understanding that zero time means "don't do this". I think that, in this case, the zero is way clearer than having to juggle a second config item and then having handle the cases where they conflict (eg. disabled but value specified, enabled but zero specified, etc).

I'm not completely opposed to changing it, but the zero feels simpler/cleaner to me.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
breedx-splk: I can make it match the class name. 👍🏻

splunk-otel-android/src/main/java/com/splunk/rum/ActivityCallbacks.java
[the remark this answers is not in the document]
jkwatson: probably, yes. 👍🏽

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
jkwatson: Some other thoughts... maybe use a `Duration` rather than assuming ms? Or maybe an int and a TimeUnit or some such? 

I think that the name of this method is going to be confusing to users and it will be even more confusing to have to realize that disabling this requires sending in a magic number. After all, '0' could mean "do it continuously" rather than "disable it". 

Another thought to think about: this configuration might want to be included in the config blob that we send up with the initialization span.

laurit: You could internally still use 0 or negative value for disabled when user facing api uses `Duration` and has an explicit disable method.

splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+
+public class SlowRenderingDetectorImpl implements SlowRenderingDetector {
+
+    public static final int SLOW_THRESHOLD_MS = 16;
```
laurit, line 25: These seem somewhat arbitrary. I suspect people will not realise that frozen doesn't mean completely hung but took 700ms to render. These values should be mentioned somewhere. Maybe even make them configurable?

breedx-splk: These values are taken directly from the android docs. https://developer.android.com/topic/performance/vitals/frozen
16ms is a threshhold for 60FPS rendering. 700ms is a google definition.

I agree that making these configurable could be useful, but I think that should be a follow-up issue/item.

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
breedx-splk: > Some other thoughts... maybe use a `Duration` rather than assuming ms? Or maybe an int and a TimeUnit or some such?

I went with milliseconds based on [this past discussion in GDI spec](https://github.com/signalfx/gdi-specification/pull/150#discussion_r748397270) and [this part of the otel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/2fb3b181ae2ee928039cda3baa8acc1eea701ce8/specification/sdk-environment-variables.md#duration).
 
> I think that the name of this method is going to be confusing to users and it will be even more confusing to have to realize that disabling this requires sending in a magic number. After all, '0' could mean "do it continuously" rather than "disable it".

You're wrong 😀 but if you feel that strongly about the magic number I will concede and introduce another boolean. Do you have a recommendation for a better method name?

> Another thought to think about: this configuration might want to be included in the config blob that we send up with the initialization span.

Seems reasonable.

laurit (approved):

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
jkwatson: > > Some other thoughts... maybe use a `Duration` rather than assuming ms? Or maybe an int and a TimeUnit or some such?
> 
> 
> 
> I went with milliseconds based on [this past discussion in GDI spec](https://github.com/signalfx/gdi-specification/pull/150#discussion_r748397270) and [this part of the otel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/2fb3b181ae2ee928039cda3baa8acc1eea701ce8/specification/sdk-environment-variables.md#duration).
> 
>  

Those are both about env vars. I think having it match the programmatic versions in the otel Java SDK would make sense.

breedx-splk: > Some other thoughts... maybe use a `Duration` rather than assuming ms? Or maybe an int and a TimeUnit or some such?

Duration is API 26, and we require 21, so no dice there right? https://developer.android.com/reference/java/time/Duration

How is the throttling exporter `windowSize` using `Duration`?

breedx-splk: Oh nevermind, I derped on the desugaring and verified that it's handled. I will make it a `Duration`.

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
```diff
@@ -266,6 +298,21 @@ public Builder anrDetectionEnabled(boolean enable) {
             return this;
         }
 
+        /**
+         * Configures the rate at which frame render durations are polled. Pass zero to
+         * disable polling for slow/frozen renders.
```
jkwatson, line 303:
```suggestion
         * Configures the rate at which frame render durations are polled. 
```

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
```diff
@@ -266,6 +298,20 @@ public Builder anrDetectionEnabled(boolean enable) {
             return this;
         }
 
+        /**
+         * Configures the rate at which frame render durations are polled.
+         * @param interval - The period that should be used for polling
+         * @return this
+         */
+        public Builder slowRenderPollingDuration(Duration interval){
```
mateuszrzeszutek, line 306: WDYT about using the `SlowRenderingDetector` name here as well? (and just using the same name for this feature everywhere in code). E.g. `slowRenderingDetectorPollingInterval()`, `slowRenderingDetectorEnabled()`

PR #244, opened 2022-03-01
Use snapshot version of otel

laurit:
Hopefully this fixes desugaring issue.


jkwatson:
I think publishing a version of our library that has a transitive dependency on a snapshot is a very bad idea. If this is only to try things out, pre-release, then 👍🏽

iNikem (approved):

ber4444:
thanks a lot, it did fix the issue

PR #251, opened 2022-03-10
Buffer spans through storage

breedx-splk:
This adds the ability to buffer all span data through the local storage. It is disabled by default but can be enabled.
When enabled, all span data are written to a series of timestamped files with a `.spans` prefix. A scheduled thread looks every 5 seconds for new data so that it can read the files (in timestamp order) and send these to ingest.

In the preexisting exporter, the span throttling mechanism is fairly complicated -- it throttles per component types per time window. With the new exporter, we keep track of the number of bytes being sent over the last 5 `send()` calls and if it exceeds 15kBps we throttle back. Throttling in this case simply means not processing any file data until our averaged bandwidth has gone back under our threshold. With this approach, larger batches _can_ spike us temporarily over the threshold, but we will not process additional files until the average bitrate has relaxed. 

With the storage buffering exporter, data is never purposefully dropped when we throttle.

There are 2 major limitations right now that will be addressed with followup actions:

* Retry strategy around send failures prior to deleting a file.
* Limiting how much storage is used.


splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -241,24 +247,65 @@ private SdkTracerProvider buildTracerProvider(
...
+        Path spanFilesPath = filesDir.resolve("spans");
+        DiskToZipkinExporter diskToZipkinExporter = DiskToZipkinExporter.builder()
+                .connectionUtil(connectionUtil)
+                .sender(sender)
+                .spanFilesPath(spanFilesPath)
+                .build();
+        diskToZipkinExporter.startPolling();
+
+        SpanExporter diskBufferingExporter = getToDiskExporter();
+        SpanExporter spanExporter = config.decorateWithSpanFilter(diskBufferingExporter);
```
mateuszrzeszutek, line 279: This line is common to both disk & memory pipelines; how about moving it up to the `buildExporter` method?

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -241,24 +247,65 @@ private SdkTracerProvider buildTracerProvider(
...
+        }
+        return buildStorageBufferingExporter(connectionUtil);
+    }
+
+    private SpanExporter buildStorageBufferingExporter(ConnectionUtil connectionUtil) {
+        Sender sender = OkHttpSender.newBuilder()
+                .endpoint(getEndpoint())
+                .build();
+        Path filesDir = Paths.get(application.getApplicationContext().getFilesDir().toURI());
+        Path spanFilesPath = filesDir.resolve("spans");
```
mateuszrzeszutek, line 270: Can we pass this path to the `getToDiskExporter` method too? Currently it's being computed both here and in `getToDiskExporter`/`ZipkinWriteToDiskExporterFactory`, let's stick to just one place.

splunk-otel-android/src/main/java/com/splunk/rum/FileUtils.java
```diff
@@ -0,0 +1,58 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+// Basic wrapper around filesystem operations, primarily for testing
+public class FileUtils {
```
mateuszrzeszutek, line 19: This probably does not need to be a part of public API
```suggestion
final class FileUtils {
```

splunk-otel-android/src/main/java/com/splunk/rum/ZipkinToDiskSender.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.time.Clock;
+import java.util.List;
+
+import zipkin2.Call;
+import zipkin2.codec.Encoding;
+import zipkin2.reporter.Sender;
+
+public class ZipkinToDiskSender extends Sender {
```
mateuszrzeszutek, line 14:
```suggestion
final class ZipkinToDiskSender extends Sender {
```

splunk-otel-android/src/main/java/com/splunk/rum/ZipkinWriteToDiskExporterFactory.java
```diff
@@ -0,0 +1,41 @@
...
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import zipkin2.reporter.Sender;
+
+/**
+ * Creates a ZipkinSpanExporter that is configured with an instance of
+ * a ZipkinToDiskSender that writes telemetry to disk.
+ */
+public class ZipkinWriteToDiskExporterFactory {
```
mateuszrzeszutek, line 19:
```suggestion
final class ZipkinWriteToDiskExporterFactory {
```

splunk-otel-android/src/main/java/com/splunk/rum/ZipkinWriteToDiskExporterFactory.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private ZipkinWriteToDiskExporterFactory(){
+    }
+
+    public static ZipkinSpanExporter create(Path path) {
```
mateuszrzeszutek, line 24:
```suggestion
    static ZipkinSpanExporter create(Path path) {
```

splunk-otel-android/src/main/java/com/splunk/rum/ZipkinToDiskSender.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private final Path path;
+    private final FileUtils fileUtils;
+    private final Clock clock;
+
+    public ZipkinToDiskSender(Path path) {
```
mateuszrzeszutek, line 20:
```suggestion
    ZipkinToDiskSender(Path path) {
```

splunk-otel-android/src/main/java/com/splunk/rum/DiskToZipkinExporter.java
```diff
@@ -0,0 +1,182 @@
...
+                break;
+            }
+        }
+        if (!sentAnything) {
+            bandwidthTracker.tick(emptyList());
+        }
+    }
+
+    private List<Path> getPendingFiles() throws IOException {
+        return fileUtils.listFiles(spanFilesPath)
```
mateuszrzeszutek, line 87: The stream returned by `Files.list` must be closed - please use try-with-resources here.

splunk-otel-android/src/main/java/com/splunk/rum/FileUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+    void writeAsLines(Path tmpFile, List<byte[]> blocksOfData) throws IOException {
+        try (FileOutputStream out = new FileOutputStream(tmpFile.toFile())) {
```
mateuszrzeszutek, line 22: You can use `Files.newOutputStream()` instead of the legacy `File` API

splunk-otel-android/src/main/java/com/splunk/rum/DiskToZipkinExporter.java
```diff
@@ -0,0 +1,182 @@
...
+
+import zipkin2.Call;
+import zipkin2.reporter.Sender;
+
+/**
+ * An exporter that pulls pre-encoded zipkin spans from storage and sends them
+ * via a sender. It is bandwidth sensitive and will throttle back if the limit
+ * is exceeded.
+ */
+public class DiskToZipkinExporter {
```
mateuszrzeszutek, line 26:
```suggestion
final class DiskToZipkinExporter {
```

splunk-otel-android/src/main/java/com/splunk/rum/DiskToZipkinExporter.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private final ScheduledExecutorService threadPool;
+    private final ConnectionUtil connectionUtil;
+    private final Sender sender;
+    private final Path spanFilesPath;
+    private final FileUtils fileUtils;
+    private final BandwidthTracker bandwidthTracker;
+    private final double bandwidthLimit;
+
+    public DiskToZipkinExporter(Builder builder) {
```
mateuszrzeszutek, line 36:
```suggestion
    DiskToZipkinExporter(Builder builder) {
```

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
```diff
@@ -38,6 +38,7 @@
 
     public static final boolean DEFAULT_ENABLE_SLOW_RENDERING_DETECTION = true;
     public static final Duration DEFAULT_SLOW_RENDER_POLLING_INTERVAL = Duration.ofSeconds(1);
+    public static final double DEFAULT_MAX_UNCOMPRESSED_BANDWIDTH = 15.0 * 1024;
```
mateuszrzeszutek, line 41: That's not configurable right now; WDYT about moving it to the `DiskToZipkinExporter` class, where it's actually used?

splunk-otel-android/src/main/java/com/splunk/rum/DiskToZipkinExporter.java
```diff
@@ -0,0 +1,182 @@
...
+    private List<byte[]> readFileCompletely(Path file) {
+        try {
+            return fileUtils.readFileCompletely(file);
+        } catch (IOException e) {
+            Log.w(LOG_TAG, "Error reading span data from file " + file, e);
+            return emptyList();
+        }
+    }
+
+    void stop() {
```
mateuszrzeszutek, line 125: Is this method called anywhere?

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Utility class to track how much bandwidth is being used by span data.
+ * It tracks raw, uncompressed spans bytes being input to a sender and is not
+ * intended to represent actual gzipped/compressed/tls bytes on the network.
+ */
+class BandwidthTracker {
+    private final static int DATAPOINTS_TO_TRACK = 6;
+    private final Clock clock;
+    private final List<Long> times = new ArrayList<>();
+    private final List<Long> sizes = new ArrayList<>();
```
mateuszrzeszutek, line 16: That's probably a bit of micro-optimization, but WDYT about using `ArrayDeque` instead? It doesn't recreate the underlying array after removing the first element, it just changes the head index.

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
```diff
@@ -0,0 +1,62 @@
...
+        if (sizes.size() < 2) return 0;
+
+        // Don't count the first ingest payload
+        double total = sizes.stream().skip(1).reduce(0L, Long::sum, Long::sum);
+
+        double timeDelta = (times.get(times.size() - 1) - times.get(0)) / 1000.0;
+        return total / timeDelta;
+    }
+
+    double latestBlockRate() {
```
mateuszrzeszutek, line 56: This method does not seem to be used in production code, is it needed?

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
```diff
@@ -0,0 +1,62 @@
...
+        sizes.add(currentSize);
+    }
+
+    double totalSustainedRate() {
+        if (sizes.size() < 2) return 0;
+
+        // Don't count the first ingest payload
+        double total = sizes.stream().skip(1).reduce(0L, Long::sum, Long::sum);
+
+        double timeDelta = (times.get(times.size() - 1) - times.get(0)) / 1000.0;
```
mateuszrzeszutek, line 52: That's in seconds, right? How about `timeDeltaInSeconds`?

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
```diff
@@ -0,0 +1,62 @@
...
+        return total / timeDelta;
```
mateuszrzeszutek, line 53: The unit of the value returned by this operation is `B/s`, is that correct? Can we document that somehow? It was not immediately obvious to me what `sustainedRate` or `bandwidthLimit` actually represent.

splunk-otel-android/src/test/java/com/splunk/rum/DiskToZipkinExporterTest.java
```diff
@@ -0,0 +1,161 @@
...
+        when(fileUtils.listFiles(spanFilesPath)).thenReturn(files);
+        when(fileUtils.isRegularFile(file1)).thenReturn(true);
+        when(fileUtils.isRegularFile(file2)).thenReturn(true);
+        when(fileUtils.isRegularFile(imposter)).thenReturn(true);
+        when(fileUtils.readFileCompletely(file1)).thenReturn(file1Spans);
+        when(fileUtils.readFileCompletely(file2)).thenReturn(file2Spans);
+    }
+
+    @Test
+    public void testHappyPathExport() throws Exception {
```
mateuszrzeszutek, line 68: Can you add some verifications that `bandwidthTracker.tick()` was called?

splunk-otel-android/src/main/java/com/splunk/rum/Config.java
```diff
@@ -173,13 +176,18 @@ SpanExporter decorateWithSpanFilter(SpanExporter exporter) {
         return spanFilterExporterDecorator.apply(exporter);
     }
 
+    public boolean isDiskBufferingDisabled() {
+        return !enableDiskBuffering;
+    }
```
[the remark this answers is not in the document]
breedx-splk: sure.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
breedx-splk: Yeah good call. There was something nagging me about this in my brain. The initialization event is also duplicated, so I've pulled that up too.

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
breedx-splk: It was really nice for logging, but since that was probably too verbose I removed it, so yeah, I guess this can go too.

splunk-otel-android/src/test/java/com/splunk/rum/DiskToZipkinExporterTest.java
breedx-splk: Good call. 👍🏻

mateuszrzeszutek:
Screen Shot 2022-03-14 at 13 12 38
Not sure if you've seen these, but there's still a few unresolved comments (it's quite easy to miss them when GH collapses them, I've missed a few of those myself)

breedx-splk:
> Not sure if you've seen these, but there's still a few unresolved comments (it's quite easy to miss them when GH collapses them, I've missed a few of those myself)

🤦🏻 I definitely did not see those. Sorry. I'll have a look soon....and thanks for all the input!

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
breedx-splk: This changed with the `Path` -> `File` stuff, but it's a good idea and I agree that having it in two spots makes it harder to change later. I'm going to consolidate it.

splunk-otel-android/src/main/java/com/splunk/rum/DiskToZipkinExporter.java
breedx-splk: The implementation had to change in order to remove usages of `Path`. I don't think it requires closing any longer.

splunk-otel-android/src/main/java/com/splunk/rum/FileUtils.java
breedx-splk: Impl changed to remove `Path` usages.

splunk-otel-android/src/main/java/com/splunk/rum/DiskToZipkinExporter.java
breedx-splk: It's not. Let me circle back to see where in the lifecycle it _should_ be called...

breedx-splk: Well, since the lifecycle of the exporter matches the lifecycle of `SplunkRum` and it doesn't look like that gets shut down...probably given this comment 
image

Do you think we should just remove the `stop()` method here then?

mateuszrzeszutek: Hm, let's remove it - if we ever need to support stopping the whole `SplunkRum` then we'll probably have to do a lot more work than just that one method.

splunk-otel-android/src/main/java/com/splunk/rum/ZipkinToDiskSender.java
```diff
@@ -0,0 +1,63 @@
...
+            fileUtils.writeAsLines(filename, encodedSpans);
+        } catch (IOException e) {
+            Log.e(SplunkRum.LOG_TAG, "Error writing spans to storage", e);
+        }
+        return Call.create(null);
+    }
+
+    private File createFilename(long now) {
+        String filePath = String.format(Locale.getDefault(), "%s%s%d.spans", path.getAbsolutePath(), File.separator, now);
+        return new File(filePath);
```
mateuszrzeszutek, line 61: Wouldn't
```suggestion
        return new File(path, now + ".spans");
```
work?

splunk-otel-android/src/main/java/com/splunk/rum/FileUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.stream.Stream;
+
+// Basic wrapper around filesystem operations, primarily for testing
+class FileUtils {
+
+    static File getSpansDirectory(Application application){
+        File filesDir = application.getApplicationContext().getFilesDir();
+        return new File(String.format(Locale.getDefault(), "%s%sspans", filesDir.getAbsolutePath(), File.separator));
```
mateuszrzeszutek, line 26:
```suggestion
        return new File(filesDir, "spans");
```

splunk-otel-android/src/main/java/com/splunk/rum/DiskToZipkinExporter.java
```diff
@@ -0,0 +1,183 @@
...
+import zipkin2.reporter.Sender;
+
+/**
+ * An exporter that pulls pre-encoded zipkin spans from storage and sends them
+ * via a sender. It is bandwidth sensitive and will throttle back if the limit
+ * is exceeded.
+ */
+class DiskToZipkinExporter {
+
+    public static final double DEFAULT_MAX_UNCOMPRESSED_BANDWIDTH = 15.0 * 1024;
```
mateuszrzeszutek, line 27:
```suggestion
    static final double DEFAULT_MAX_UNCOMPRESSED_BANDWIDTH = 15.0 * 1024;
```

splunk-otel-android/src/main/java/com/splunk/rum/DiskToZipkinExporter.java
mateuszrzeszutek: No, it looks like it's not needed anymore 👍

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
```diff
@@ -0,0 +1,61 @@
+package com.splunk.rum;
+
+import java.time.Clock;
+import java.util.ArrayList;
+import java.util.List;
+
+import kotlin.collections.ArrayDeque;
```
mateuszrzeszutek, line 7: Oh it's Kotlin's `ArrayDeque`... I thought it was a bit strange, since the Java one does not implement `List`

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
```diff
@@ -0,0 +1,61 @@
...
+     * Calculates the current average sustained throughput.
+     * @return - The currently tracked bandwidth, in bytes per second.
+     */
+    double totalSustainedRate() {
+        if (sizes.size() < 2) return 0;
+
+        // Don't count the first ingest payload
+        double total = sizes.stream().skip(1).reduce(0L, Long::sum, Long::sum);
+
+        double timeDeltaInSeconds = (times.get(times.size() - 1) - times.get(0)) / 1000.0;
```
mateuszrzeszutek, line 58: You could probably use `getLast()` and `getFirst()` here

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
```diff
@@ -0,0 +1,61 @@
...
+        this.clock = clock;
+    }
+
+    /**
+     * Call this method with encoded zipkin span data to have it
+     * tracked.
+     */
+    void tick(List<byte[]> zipkinSpanData) {
+        if (times.size() > DATAPOINTS_TO_TRACK) {
+            times.remove(0);
```
mateuszrzeszutek, line 35: How about 
```suggestion
            times.removeFirst();
```

splunk-otel-android/src/main/java/com/splunk/rum/ZipkinToDiskSender.java
breedx-splk: Yeah WAY simpler. Cool! Thanks.

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
breedx-splk: Oh yeah, I have a weakness about never looking at imports. 🙈 
Changing to the java version is nice as you point out below -- will do.

PR #256, opened 2022-03-14
Update Volley instrumentation to match latest instrumentation-api cha…

mateuszrzeszutek:
…nges


splunk-otel-android-volley/src/main/java/com/splunk/rum/VolleyTracing.java
```diff
@@ -29,13 +29,27 @@
  */
 public final class VolleyTracing {
 
+    /**
+     * Returns a new {@link VolleyTracing} configured with the given {@link SplunkRum}.
+     */
+    public static VolleyTracing create(SplunkRum splunkRum) {
```
mateuszrzeszutek, line 35: I figured these methods might make it a bit simpler to setup the whole thing

PR #260, opened 2022-03-28
Retry sending files.

breedx-splk:
In the event of a file sending failure, the file will be retried up to 10 times. This is not yet configurable through the API (erred on the side of keeping it simple until needed).

With the standard attempt rate at 5 seconds, a splunk ingest outage of even 1 minute would result in mobile RUM data loss. Should we consider bumping the max attempt count up to 500 or 1000?

In the event that all retries for a file have been exhausted, then a warning is logged and the file is deleted.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
@@ -94,7 +94,11 @@ class FileSender {
 
         public boolean incrementAndCheckMax(File file) {
             Integer count = attempts.merge(file, 1, (cur, x) -> cur + 1);
-            return count >= maxRetries;
+            boolean exceededRetries = count >= maxRetries;
+            if(exceededRetries){
+                Log.w(LOG_TAG, "Dropping data in " + file + " (max retries exceeded " + maxRetries + ")");
+            }
+            return exceededRetries;
         }
     }
 
--- a/splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
@@ -107,7 +107,7 @@ class FileSender {
         private Sender sender;
         private FileUtils fileUtils = new FileUtils();
         private BandwidthTracker bandwidthTracker;
-        public RetryTracker retryTracker;
+        private RetryTracker retryTracker;
         private int maxRetries = DEFAULT_MAX_RETRIES;
 
         Builder sender(Sender sender) {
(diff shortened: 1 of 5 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
```diff
@@ -0,0 +1,139 @@
...
+            return exceededRetries;
+        }
+    }
+
+    static class Builder {
+
+        private Sender sender;
+        private FileUtils fileUtils = new FileUtils();
+        private BandwidthTracker bandwidthTracker;
+        public RetryTracker retryTracker;
```
mateuszrzeszutek, line 110:
```suggestion
        private RetryTracker retryTracker;
```

splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        public boolean incrementAndCheckMax(File file) {
+            Integer count = attempts.merge(file, 1, (cur, x) -> cur + 1);
+            boolean exceededRetries = count >= maxRetries;
+            if(exceededRetries){
+                Log.w(LOG_TAG, "Dropping data in " + file + " (max retries exceeded " + maxRetries + ")");
+            }
+            return exceededRetries;
+        }
```
mateuszrzeszutek, line 102:
```suggestion
        void clear(File file) {
            attempts.remove(file);
        }

        boolean incrementAndCheckMax(File file) {
            Integer count = attempts.merge(file, 1, (cur, x) -> cur + 1);
            boolean exceededRetries = count >= maxRetries;
            if(exceededRetries){
                Log.w(LOG_TAG, "Dropping data in " + file + " (max retries exceeded " + maxRetries + ")");
            }
            return exceededRetries;
        }
```

mateuszrzeszutek (approved):

mateuszrzeszutek:
> With the standard attempt rate at 5 seconds, a splunk ingest outage of even 1 minute would result in mobile RUM data loss. Should we consider bumping the max attempt count up to 500 or 1000?

Another idea would be to use exponential backoff and retry 10 times, but with pauses of e.g. 15s, 30s, 1min, 2min, 4min, 5min, 5min...
The retry state map could keep a (count, timestamp) pair; the `FileSender` would simply do nothing before the pause interval passes.

breedx-splk:
> Another idea would be to use exponential backoff and retry 10 times, but with pauses of e.g. 15s, 30s, 1min, 2min, 4min, 5min, 5min... The retry state map could keep a (count, timestamp) pair; the `FileSender` would simply do nothing before the pause interval passes.

I like it. Maybe not capping at 5min. How about `Math.min(60, (5 * tries))` seconds, with a total of 20 retries. I think that then looks like 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60, 60, 60, 60, 60, 60, 60, 60, 60. That allows for 14.5 minutes of failure time before dropping.

mateuszrzeszutek:
Looks nice -- let's do it this way.

mateuszrzeszutek (approved):

PR #261, opened 2022-04-01
Add @RumScreenName annotation for customization

breedx-splk:
This introduces a new annotation `@RumScreenName` that can be used to customize the `screen.name` attribute on Fragment and Activity spans.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,8 +302,8 @@ By default, the instrumentation uses the simple class name of each `Fragment`
 and `Activity` type as the value of the `screen.name` attribute. To customize the
 screen name, use the `@RumScreenName` annotation.
 
-For example, the following Activity will show up with the `screen.name` 
-set to the value "Buttercup".
+For example, the following activity appears with the `screen.name` 
+attribute set to the value "Buttercup":
 
 ```java
 @RumScreenName("Buttercup")
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ Splunk RUM defines renders as slow or frozen following the [Android Vitals defin
 | Slow     | >16ms  | slowRenders    | count       | 
 | Frozen   | >700ms | frozenRenders  | count       |
 
-## Customizing Screen Names
+## Customizing screen names
 
 By default, instrumentation will use the simple class name of each `Fragment`
 and `Activity` type in the `screen.name` attribute. If you wish to customize this
--- a/README.md
+++ b/README.md
@@ -298,9 +298,9 @@ Splunk RUM defines renders as slow or frozen following the [Android Vitals defin
 
 ## Customizing screen names
 
-By default, instrumentation will use the simple class name of each `Fragment`
-and `Activity` type in the `screen.name` attribute. If you wish to customize this
-behavior, you may use the `@RumScreenName` annotation.
+By default, the instrumentation uses the simple class name of each `Fragment`
+and `Activity` type as the value of the `screen.name` attribute. To customize the
+screen name, use the `@RumScreenName` annotation.
 
 For example, the following Activity will show up with the `screen.name` 
 set to the value "Buttercup".
--- a/README.md
+++ b/README.md
@@ -296,6 +296,21 @@ Splunk RUM defines renders as slow or frozen following the [Android Vitals defin
 | Slow     | >16ms  | slowRenders    | count       | 
 | Frozen   | >700ms | frozenRenders  | count       |
 
+## Customizing Screen Names
+
+By default, instrumentation will use the simple class name of each `Fragment`
+and `Activity` type in the `screen.name` attribute. If you wish to customize this
+behavior, you may use the `@RumScreenName` annotation.
+
+For example, the following Activity will show up with the `screen.name` 
+set to the value "Buttercup".
+
+```java
+@RumScreenName("Buttercup")
+public class MainActivity extends Activity {
+    ...
+}
+```
 
 ## Troubleshooting
 
--- a/splunk-otel-android/src/main/java/com/splunk/rum/FragmentTracer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/FragmentTracer.java
@@ -26,7 +26,7 @@ class FragmentTracer {
     static final AttributeKey<String> FRAGMENT_NAME_KEY = AttributeKey.stringKey("fragmentName");
 
     private final String fragmentName;
-    private final String customScreenName;
+    private final String screenName;
     private final Tracer tracer;
     private final ActiveSpan activeSpan;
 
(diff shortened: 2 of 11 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/FragmentTracer.java
```diff
@@ -26,12 +26,15 @@
     static final AttributeKey<String> FRAGMENT_NAME_KEY = AttributeKey.stringKey("fragmentName");
 
     private final String fragmentName;
+    private final String customScreenName;
```
jkwatson, line 29: rather than saving the customScreenName only when it's "custom", what if you just saved the "screenName" and made it either the fragmentName or the customScreenName value, as appropriate. Then you can leave off the ternaries when using it.

splunk-otel-android/src/test/java/com/splunk/rum/ActivityTracerTest.java
```diff
@@ -160,6 +160,23 @@ public void addPreviousScreen() {
...
+        activityTracer.endActiveSpan();
+        SpanData span = getSingleSpan();
+        assertEquals("squarely", span.getAttributes().get(SplunkRum.SCREEN_NAME_KEY));
+    }
+
+    @RumScreenName("squarely")
+    static class AnnotatedActivity extends Activity {
+        @Override
+        public void onPointerCaptureChanged(boolean hasCapture) {
+        }
```
jkwatson, line 177: any particular reason to override this method here?

README.md
```diff
@@ -296,6 +296,21 @@ Splunk RUM defines renders as slow or frozen following the [Android Vitals defin
...
+By default, instrumentation will use the simple class name of each `Fragment`
+and `Activity` type in the `screen.name` attribute. If you wish to customize this
+behavior, you may use the `@RumScreenName` annotation.
+
+For example, the following Activity will show up with the `screen.name` 
+set to the value "Buttercup".
+
+```java
+@RumScreenName("Buttercup")
+public class MainActivity extends Activity {
```
mateuszrzeszutek, line 310: Nice, I like it way more than the `.setScreenName()` idea 👍

splunk-otel-android/src/main/java/com/splunk/rum/RumScreenName.java
```diff
@@ -0,0 +1,19 @@
+package com.splunk.rum;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation can be used to customize the screen.name attribute
```
mateuszrzeszutek, line 9: Nit
```suggestion
 * This annotation can be used to customize the {@code screen.name} attribute
```

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/com/splunk/rum/FragmentTracer.java
breedx-splk: This did cross my mind....I'll give it a shot.

splunk-otel-android/src/main/java/com/splunk/rum/Pre29ActivityCallbacks.java
```diff
@@ -119,12 +119,8 @@ public void onActivityDestroyed(@NonNull Activity activity) {
     }
 
     private ActivityTracer getTracer(Activity activity) {
-        ActivityTracer activityTracer = tracersByActivityClassName.get(activity.getClass().getName());
-        if (activityTracer == null) {
-            activityTracer = new ActivityTracer(activity, initialAppActivity, tracer, visibleScreenTracker, appStartupTimer);
-            tracersByActivityClassName.put(activity.getClass().getName(), activityTracer);
-        }
-        return activityTracer;
+        return tracersByActivityClassName.computeIfAbsent(activity.getClass().getName(),
```
jkwatson, line 122: for some reason, I thought that this method wasn't Android API level 21 friendly, but I might be mis-remembering.

jkwatson (approved):

splunk-otel-android/src/main/java/com/splunk/rum/Pre29ActivityCallbacks.java
breedx-splk: :rage4: 
image
It's a real bummer that there's no good way to get a warning about this in Android studio huh?

jkwatson: definitely. it's very easy to get tripped up on.

breedx-splk: Yeah. Well, I rolled those pieces back...so worse code, better compatibility. 🤷🏻

laurit: I have a memory of android studio complaining when I tried using unsupported stuff. This api https://developer.android.com/studio/write/java8-support-table seems to be desugared. Perhaps using it really isn't an issue?

README.md
```diff
@@ -296,6 +296,21 @@ Splunk RUM defines renders as slow or frozen following the [Android Vitals defin
 | Slow     | >16ms  | slowRenders    | count       | 
 | Frozen   | >700ms | frozenRenders  | count       |
 
+## Customizing Screen Names
```
theletterf, line 299:
```suggestion
## Customizing screen names
```

README.md
```diff
@@ -296,6 +296,21 @@ Splunk RUM defines renders as slow or frozen following the [Android Vitals defin
...
+
+By default, instrumentation will use the simple class name of each `Fragment`
+and `Activity` type in the `screen.name` attribute. If you wish to customize this
+behavior, you may use the `@RumScreenName` annotation.
```
theletterf, line 303:
```suggestion
By default, the instrumentation uses the simple class name of each `Fragment`
and `Activity` type as the value of the `screen.name` attribute. To customize the
screen name, use the `@RumScreenName` annotation.
```

README.md
```diff
@@ -296,6 +296,21 @@ Splunk RUM defines renders as slow or frozen following the [Android Vitals defin
...
+
+For example, the following Activity will show up with the `screen.name` 
+set to the value "Buttercup".
```
theletterf, line 306:
```suggestion
For example, the following activity appears with the `screen.name` 
attribute set to the value "Buttercup":
```

theletterf (changes requested):

theletterf (approved):

splunk-otel-android/src/main/java/com/splunk/rum/Pre29ActivityCallbacks.java
breedx-splk: Turns out that `computeIfAbsent` is in the desugaring list after all!

PR #262, opened 2022-04-04
Add session ID ratio based sampler

agoallikmaa:
Added a session ID based sampler. This is enabled by setting `sessionBasedSamplingEnabled(boolean)` to `true` and specifying the ratio via `sessionBasedSamplingRatio(double)` to the desired ratio.

Internally it reuses `TraceIdRatioBasedSampler` from `io.opentelemetry:opentelemetry-sdk-trace`, as the logic is identical to what we need (and what is used in `SessionBasedSampler` of `splunk-otel-js-web`), but passes in the current session ID to it instead of `traceId`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,10 +255,10 @@ when initializing your instance of the SplunkRum API:
   When disk buffering is enabled, this can be used to adjust the maximum amount of storage
   that will be used. Default = 25MB.
 - `sessionBasedSamplingEnabled(boolean)` :
-  Enables session ID based sampling for traces. Ratio must be set with `sessionBasedSamplingRatio`,
-  as the default ratio includes all sessions. Sampling is only applied if the spans created via 
-  `SplunkRum` APIs are root spans. If a custom parent span context was created, then the sampling 
-  flag of the parent span is respected instead.
+  Set sampling ratio for session ID based sampling. This is a probability of a session being 
+  included between between 0.0 (all dropped) and 1.0 (all included). Default is 1. Sampling is only
+  applied if the spans created via `SplunkRum` APIs are root spans. If a custom parent span context
+  was created, then the sampling flag of the parent span is respected instead.
 - `sessionBasedSamplingRatio(double)` :
   Set sampling ratio for session ID based sampling. This is a probability of a session being
   included between between 0.0 (all dropped) and 1.0 (all included). Default is 1.
--- a/README.md
+++ b/README.md
@@ -254,14 +254,11 @@ when initializing your instance of the SplunkRum API:
 - `limitDiskUsageMegabytes(int)` :
   When disk buffering is enabled, this can be used to adjust the maximum amount of storage
   that will be used. Default = 25MB.
-- `sessionBasedSamplingEnabled(boolean)` :
-  Set sampling ratio for session ID based sampling. This is a probability of a session being 
-  included between between 0.0 (all dropped) and 1.0 (all included). Default is 1. Sampling is only
+- `enableSessionBasedSampling(double)` :
+  Enable session ID based sampling and set its sampling ratio. The ratio is a probability of a
+  session being included between between 0.0 (all dropped) and 1.0 (all included). Sampling is only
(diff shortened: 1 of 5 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -238,6 +238,10 @@ private SdkTracerProvider buildTracerProvider(
                 .setResource(resource);
         initializationEvents.add(new RumInitializer.InitializationEvent("tracerProviderBuilderInitialized", timingClock.now()));
 
+        if (config.isSessionBasedSamplerEnabled()) {
+            tracerProviderBuilder.setSampler(Sampler.parentBased(new SessionIdRatioBasedSampler(config.getSessionBasedSamplerRatio(), sessionId)));
```
jkwatson, line 242: Is there any reason to use `parentBased` here? I think the only input is the session id, and that should be available for all spans, yes?

README.md
```diff
@@ -251,6 +251,14 @@ when initializing your instance of the SplunkRum API:
...
   Enables the storage-based buffering of telemetry. 
   This setting is useful when instrumenting applications that might work offline for extended periods of time.
+- `sessionBasedSamplingEnabled(boolean)` :
+  Enables session ID based sampling for traces. Ratio must be set with `sessionBasedSamplingRatio`,
+  as the default ratio includes all sessions. Sampling is only applied if the spans created via 
+  `SplunkRum` APIs are root spans. If a custom parent span context was created, then the sampling 
+  flag of the parent span is respected instead.
+- `sessionBasedSamplingRatio(double)` :
+  Set sampling ratio for session ID based sampling. This is a probability of a session being
+  included between between 0.0 (all dropped) and 1.0 (all included). Default is 1.
```
jkwatson, line 261: in this case, since the default is to not sample (i.e. a value of 1.0), that we don't need 2 settings for this. Just one method like `enableSessionBasedSampling(double)` seems like it should be clear and concise enough to cover all the bases.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
agoallikmaa: The reason is that if there is a parent span created by the user manually, which is not sampled (sampling is only applied to the tracer used internally, not any tracer created by the user), that would cause the situation where a single trace contains some spans which are sampled and some which are not, which is generally undesirable. The expected behavior is that a trace is either wholly sampled or not. Parent-based makes sure to copy the sampled bit from the parent if a parent exists.

README.md
agoallikmaa: Done. Kept the boolean field internally as comparing that during initialization seems to be cleaner than comparing a double to `1.0`.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
agoallikmaa: Removed `parentBased` around it.

README.md
```diff
@@ -251,6 +251,11 @@ when initializing your instance of the SplunkRum API:
 - `diskBufferingEnabled(boolean)` : 
   Enables the storage-based buffering of telemetry. 
   This setting is useful when instrumenting applications that might work offline for extended periods of time.
+- `enableSessionBasedSampling(double)` :
+  Enable session ID based sampling and set its sampling ratio. The ratio is a probability of a
+  session being included between between 0.0 (all dropped) and 1.0 (all included). Sampling is only
+  applied if the spans created via `SplunkRum` APIs are root spans. If a custom parent span context
+  was created, then the sampling flag of the parent span is respected instead.
```
mateuszrzeszutek, line 258: Is the part about root spans and parent spans still correct after removed the `parentBased` sampling?
```suggestion
  session being included between between 0.0 (all dropped) and 1.0 (all included).
```

agoallikmaa: Removed that part.

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

PR #263, opened 2022-04-05
Remove computeIfAbsent() call from ThrottlingExporter

mateuszrzeszutek:


laurit:
Did you verify that desugaring doesn't handle this?

mateuszrzeszutek:
Hmm, you're right, it does get desugared. Nevermind then, closing this

PR #264, opened 2022-04-06
Ensure that recorded exceptions get exported as span attributes

mateuszrzeszutek:
`ModifiedSpanData` now translates exception events to span attributes. This ended up simplifying many things, as we can now use `.recordException()` on spans and be sure that they'll be properly transformed to span attributes during export.


splunk-otel-android/src/test/java/com/splunk/rum/SplunkRumTest.java
```diff
@@ -189,22 +190,17 @@ public void addException() {
...
-                .put(SemanticAttributes.EXCEPTION_MESSAGE, "oopsie")
-                .put(SplunkRum.ERROR_MESSAGE_KEY, "oopsie")
-                .put(SemanticAttributes.EXCEPTION_TYPE, "NullPointerException")
-                .put(SplunkRum.ERROR_TYPE_KEY, "NullPointerException")
-                .build();
 
-        assertEquals(expected.asMap(), spans.get(0).getAttributes().asMap());
+        assertThat(spans.get(0))
+                .hasName("NullPointerException")
+                .hasAttributes(attributes.toBuilder().put(SplunkRum.COMPONENT_KEY, SplunkRum.COMPONENT_ERROR).build())
```
breedx-splk, line 202: Is this losing coverage on the other attributes? `EXCEPTION_MESSAGE`, `ERROR_MESSAGE_KEY`, `EXCEPTION_TYPE`, `ERROR_TYPE_KEY`?

breedx-splk (changes requested):
I good change, I like the separation of concerns that resulted from this. A couple silly asks. Thanks!

splunk-otel-android/src/test/java/com/splunk/rum/SplunkRumTest.java
mateuszrzeszutek: Hmm, not really - the code that adds them is simply moved elsewhere, and tested elsewhere.
Although, we don't have an integration test that'd test `SplunkRum` and all the exporter modifications - I think I'll add one, it sounds like a useful test to have.

breedx-splk (approved):
Killer, thanks!

PR #272, opened 2022-04-13
Add ability to limit disk usage

breedx-splk:
When through-disk buffering is enabled, we want to prevent filling. up the disk if the telemetry cannot be exported for long periods of time (like when camping or in airplane mode). This introduces a default 25MB limit for telemetry on disk, along with a configuration option to adjust this limit. 

When the limit has been reached, older telemetry will be deleted to make room for new telemetry. This effectively creates a time-based ring buffer for telemetry.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/Config.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/Config.java
@@ -16,6 +16,8 @@
 
 package com.splunk.rum;
 
+import static com.splunk.rum.DeviceSpanStorageLimiter.DEFAULT_MAX_STORAGE_USE_MB;
+
 import android.util.Log;
 
 import java.time.Duration;
@@ -48,6 +50,7 @@ public class Config {
     private final boolean slowRenderingDetectionEnabled;
     private final Duration slowRenderingDetectionPollInterval;
     private final boolean diskBufferingEnabled;
+    private final int maxUsageMegabytes;
 
     private Config(Builder builder) {
         this.beaconEndpoint = builder.beaconEndpoint;
@@ -62,6 +65,7 @@ public class Config {
         this.slowRenderingDetectionEnabled = builder.slowRenderingDetectionEnabled;
         this.spanFilterExporterDecorator = builder.spanFilterBuilder.build();
         this.diskBufferingEnabled = builder.diskBufferingEnabled;
+        this.maxUsageMegabytes = builder.maxUsageMegabytes;
     }
 
     private Attributes addDeploymentEnvironment(Builder builder) {
@@ -156,6 +160,14 @@ public class Config {
     }
 
     /**
+     * Returns the max number of megabytes that will be used to buffer telemetry data in storage.
+     * If this value is exceeded, older telemetry will be deleted until the usage is reduced.
+     */
+    public int getMaxUsageMegabytes() {
+        return maxUsageMegabytes;
+    }
+
+    /**
      * Create a new instance of the {@link Builder} class. All default configuration options will be pre-populated.
      */
     public static Builder builder() {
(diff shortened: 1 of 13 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/Config.java
```diff
@@ -155,6 +159,14 @@ public boolean isDiskBufferingEnabled() {
         return diskBufferingEnabled;
     }
 
+    /**
+     * Returns the max number of megabytes that will be used to buffer telemetry data in storage.
+     * If this value is exceeded, older telemetry will be deleted until the usage is reduced.
+     */
```
mateuszrzeszutek, line 165: Nit: could you add a similar comment to the builder method? Users of this library don't really have any reason to use the getter methods (I think), they probably only construct the `Config` and never get its properties - and, I suppose, in case of any doubts they'd check the builder javadocs first.

splunk-otel-android/src/main/java/com/splunk/rum/ZipkinToDiskSender.java
```diff
@@ -46,6 +42,10 @@ public int messageSizeInBytes(List<byte[]> encodedSpans) {
 
     @Override
     public Call<Void> sendSpans(List<byte[]> encodedSpans) {
+        if (!storageLimiter.ensureFreeSpace()) {
```
mateuszrzeszutek, line 45: Nit: you could roughly estimate the amount of space needed by summing the sizes of `encodedSpans`

splunk-otel-android/src/main/java/com/splunk/rum/DeviceSpanStorageLimiter.java
```diff
@@ -0,0 +1,92 @@
...
+     * Ensures that the storage currently used by spans has not exceeded the limit.
+     * If it does, it will delete older files until the limit is no longer exceeded.
+     *
+     * @return - true if the free space is under the limit (including when files have
+     * been deleted to return back under the limit), false if not enough space could be
+     * freed to get us back under out limit.
+     */
+    boolean ensureFreeSpace() {
+        tryFreeingSpace();
+        return path.getFreeSpace() > limitInBytes();
```
mateuszrzeszutek, line 34: I don't really get this condition -- it's comparing `the number of unallocated bytes in the partition` (free disk space if I understand it correctly) against our own usage limit (25 MB). 
What if for some reason we did not manage to remove old span files and free space in our "spans dir", but there's still 1gb more on the disk? Should we try to write the spans in this case?
Or, the disk partition is really small, like 20MB, and we've managed to free all space? Should we not write the spans, if we're under the limit?

splunk-otel-android/src/main/java/com/splunk/rum/ZipkinToDiskSender.java
breedx-splk: Sure, you could. I thought it would be simpler to just keep deleting until we're back under the limit, and I think it's ok if a single write takes us back over the limit.  This means that the limit is a "soft limit" that we try for but are not strict about. I hope this is ok because we're talking about a whopping 25MB. :)

splunk-otel-android/src/main/java/com/splunk/rum/DeviceSpanStorageLimiter.java
breedx-splk: I knew this was going to come up..... 🙃 

The idea was to do two things:
1. Delete old telemetry if we're using more than a certain amount on disk.
2. Play nice if the disk is getting full

So with #2, we are simply checking to see if the disk is getting full -- and we do this by seeing if the amount of free space is less than our max size limit. 

The downside of this is that the limit is now doing double duty -- but I definitely didn't think another config item was warranted and I didn't want to hard code something. Maybe it's confusing....I couldn't think of a great way to make it clearer (open to ideas here).

mateuszrzeszutek: I think it's fine if we just add a comment explaining the reason for this, and that it's intentional.

mateuszrzeszutek (approved):

PR #273, opened 2022-04-19
Add SplunkRUM.isInitialized

agoallikmaa:
Added `SplunkRUM.isInitialized` method and mentioned it in documentation. Since initialization as shown in the documentation makes it effectively thread-safe (unless used from threads started before that), the non-volatility of this (and `getInstance`) doesn't really need to be mentioned explicitly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,8 @@ class MyApplication extends Application {
 Examples of this process can be seen in the sample application included in this repository in
 the `sample-app` submodule.
 
+To check whether the RUM library has been initialized, call `SplunkRUM.isInitialized()`.
+
 ### Instrument WebViews using the Browser RUM agent
 
 Mobile RUM instrumentation and Browser RUM instrumentation can be used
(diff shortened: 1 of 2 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/SplunkRum.java
```diff
@@ -130,6 +130,14 @@ static SplunkRum initialize(Config config, Application application, Supplier<Con
         return INSTANCE;
     }
 
+    /**
+     *
+     * @return True if the Splunk RUM library has been successfully initialized.
```
laurit, line 135:
```suggestion
     * @return true if the Splunk RUM library has been successfully initialized.
```

laurit (approved):

README.md
```diff
@@ -171,6 +171,8 @@ class MyApplication extends Application {
 Examples of this process can be seen in the sample application included in this repository in
 the `sample-app` submodule.
 
+To check whether the RUM library has been initialized, call `SplunkRUM.isInitialized()`.
```
mateuszrzeszutek, line 174: Nit
```suggestion
To check whether the RUM library has been initialized, call `SplunkRum.isInitialized()`.
```

mateuszrzeszutek (approved):

PR #277, opened 2022-04-19
Spotless

laurit:
using `googleJavaFormat().aosp()` for java and `ktlint()` for kotlin


mateuszrzeszutek (approved):
That was a lot more that I thought it'd be

breedx-splk:
Woah.

PR #306, opened 2022-06-22
Don't lock on this

breedx-splk:
This helps reduce the potential opportunity for deadlocks (where another bit of code could also lock our instance). Instead of locking on `this` we just use a private/hidden lock object.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-#distributionSha256Sum=29e49b10984e585d8118b7d0bc452f944e386458df27371b49b4ac1dec4b7fda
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek (approved):

laurit:
> This helps reduce the potential opportunity for deadlocks (where another bit of code could also lock our instance). Instead of locking on `this` we just use a private/hidden lock object.

This is our internal class so the chance of someone synchronizing on it is pretty much non existent and this is purely a style issue. If you want to do an actual improvement then it would be nice if you could remember why `reportSlowRenders` needs to remove and add back all activities on each cycle and add a comment or remove it if it is not needed.

breedx-splk:
> This is our internal class so the chance of someone synchronizing on it is pretty much non existent and this is purely a style issue. 

Sure, this is an internal class, but software is long lived and it would be a simple mistake for someone to make down the road. This is intentioned to help reduce the chance of that and is not "purely a style issue". 

> If you want to do an actual improvement then it would be nice if you could remember why `reportSlowRenders` needs to remove and add back all activities on each cycle and add a comment or remove it if it is not needed.

Huh.

PR #315, opened 2022-07-13
Ignore background app starts

laurit:


splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
```diff
@@ -44,15 +57,76 @@ Span start(Tracer tracer) {
         return appStart;
     }
 
+    /** Called when Activity is created. */
+    void startUiInit() {
+        if (uiInitStarted || isStartedFromBackground) {
+            return;
+        }
+        uiInitStarted = true;
```
breedx-splk, line 65: This isn't threadsafe...do we care? Can two activities get created at the same time and step on each other?

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
```diff
@@ -44,15 +57,76 @@ Span start(Tracer tracer) {
...
+        }
+    }
+
+    void setCompletionCallback(Runnable completionCallback) {
+        this.completionCallback = completionCallback;
+    }
+
     void end() {
-        if (overallAppStartSpan != null) {
+        if (overallAppStartSpan != null && !uiInitTooLate && !isStartedFromBackground) {
```
breedx-splk, line 78: Are we guaranteed that `end()` is called from the same thread that might have assigned those fields? If not, should we make all those volatile as well?

breedx-splk (approved):
This looks good to me. Because this is fixing a lifecycle issue, I do want to make sure we're also extra mindful of any threading concerns.

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
laurit: This should be called from the UI thread, I wouldn't expect this to be called from multiple threads.

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
laurit: This should also be called from the UI thread. I make `isStartedFromBackground` volatile, could make others also if you feel it is better that way.

mateuszrzeszutek: `isStartedFromBackground` is only accessed from the UI thread, and does not need to be volatile, right?
I'd say that we either make all of these fields volatile, or make them all non-volatile and leave a comment saying sth like "these 3 booleans are only accessed from the UI thread"

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
laurit: Makes sense, reverted the previous change and added comment instead.

mateuszrzeszutek (approved):

PR #323, opened 2022-07-28
Crash reporting enhancements

mateuszrzeszutek:
This PR contains 2 changes:
* the attribute value length was increased, so that bigger exception stack traces would fit in
* only the first uncaught exception is treated as a crash, every next one will be marked as an ordinary error


breedx-splk (approved):
Sure, why not? What's the worst thing that could happen.....

laurit (approved):

PR #327, opened 2022-08-05
Remove the local IP address from span

mateuszrzeszutek:


laurit (approved):

splunk-otel-android/src/main/java/com/splunk/rum/CustomZipkinEncoder.java
```diff
@@ -49,6 +50,9 @@ public int sizeInBytes(Span span) {
 
     @Override
     public byte[] encode(Span span) {
+        Endpoint localEndpoint = Endpoint.newBuilder().serviceName(span.localServiceName()).build();
```
breedx-splk, line 53: Ah what a drag that this is now a rendering concern. It really feels like it belongs higher in the stack somewhere, but I looked high and low and couldn't find a better place to tap in right now. This works for the short term anyway.

mateuszrzeszutek: I added https://github.com/open-telemetry/opentelemetry-java/issues/4669 to the SDK repo, I'm hoping to fix it there before the release

PR #328, opened 2022-08-09
Add StandardAttributes.APP_BUILD_TYPE

agoallikmaa:


mateuszrzeszutek (approved):

laurit (approved):

breedx-splk (approved):
Users aren't expecting a default, are they? I assume not, we can add one later if it comes up again.

PR #329, opened 2022-08-10
Volley example and readme

breedx-splk:
Adds a README to the Volley instrumentation and gives a high-level description of how to use it. It also wires up the Volley tracing to the sample app, and it has a button on the first fragment to trigger it. 

As part of this effort, I also discovered that the `VolleyHttpClientAttributesGetter` was failing to find some headers due to casing mismatches. The headers matching is now case insensitive and there are tests to cover that.


aurbiztondo-splunk (approved):

splunk-otel-android-volley/src/main/java/com/splunk/rum/VolleyHttpClientAttributesGetter.java
```diff
@@ -68,18 +71,26 @@ public String method(RequestWrapper requestWrapper) {
 
     @Override
     public List<String> requestHeader(RequestWrapper requestWrapper, String name) {
-        String header;
+        Request<?> request = requestWrapper.getRequest();
         try {
-            header = requestWrapper.getRequest().getHeaders().get(name);
-            if (header == null) {
-                header = requestWrapper.getAdditionalHeaders().get(name);
+            String result = findCaseInsensitive(name, request.getHeaders());
```
laurit, line 76: I believe it would be more correct to return all headers that match the given name instead of returning only the first one.

sample-app/build.gradle.kts
```diff
@@ -61,6 +62,10 @@ dependencies {
     implementation("androidx.navigation:navigation-fragment:2.5.1")
     implementation("androidx.navigation:navigation-ui:2.5.1")
     implementation(project(":splunk-otel-android"))
+    implementation(project(":splunk-otel-android-volley"))
+    implementation("com.android.volley:volley:1.2.1")
+    implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:$otelVersion")
+    implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-semconv:$otelVersion")
```
mateuszrzeszutek, line 68: Shouldn't these be included transitively through our projects? Do the users need to add these dependencies too?

splunk-otel-android-volley/README.md
```diff
@@ -0,0 +1,59 @@
+
+# Splunk OpenTelemetry Android Volley Instrumentation
+
+Status: Experimental
```
mateuszrzeszutek, line 4: Nit: we could use the same markup as in the javaagent docs
```suggestion
> :construction: &nbsp;Status: Experimental
```

mateuszrzeszutek (approved):

splunk-otel-android-volley/src/main/java/com/splunk/rum/VolleyHttpClientAttributesGetter.java
mateuszrzeszutek: Looks like Volley uses `Map<String, String>` to store the request headers -- there's no way we'll ever have more than 1 value for a given name.

laurit: My understanding is that if you have {"foo":"bar", "FOO":"baz"} then asking for "Foo" should return ["bar", "baz"] because header names should be compared in case insensitive way.

breedx-splk: @laurit you're right, but as @mateuszrzeszutek pointed out this is a deficiency in the Volley client/request APIs. As far as I can tell, there isn't even any way of specifying multiple values for a single request header, without doing some case sensitive `Map` hackery (eg. `Cookie: abc`, `COokie: def`, ... `COOKIE: zzz`).

The response side does handle multiple values, but it might still suffer from case sensitivity. I'll revisit that and add missing test coverage.

sample-app/build.gradle.kts
breedx-splk: No, I don't think the main sdk includes these. Volley users will need to include them, yes.

splunk-otel-android-volley/src/main/java/com/splunk/rum/VolleyHttpClientAttributesGetter.java
breedx-splk: Oh, I think I see the point now tho -- if multiple were specified with different casing we would only return the first one. I'll fix it.

splunk-otel-android-volley/src/main/java/com/splunk/rum/VolleyHttpClientAttributesGetter.java
```diff
@@ -68,18 +70,27 @@ public String method(RequestWrapper requestWrapper) {
...
         try {
-            header = requestWrapper.getRequest().getHeaders().get(name);
-            if (header == null) {
-                header = requestWrapper.getAdditionalHeaders().get(name);
-            }
-
+            Map<String, String> headers = request.getHeaders();
+            Map<String, String> additionalHeaders = requestWrapper.getAdditionalHeaders();
+            List<String> result = new ArrayList<>(headers.size() + additionalHeaders.size());
+            result.addAll(findCaseInsensitive(name, headers));
```
mateuszrzeszutek, line 78: Nit: I think in the usual scenario there'll be much less header values than the total number of headers in the request:
```suggestion
            List<String> result = new ArrayList<>();
            result.addAll(findCaseInsensitive(name, headers));
```
Also, you could pass a reference to the `result` list to the `findCaseInsensitive` method to avoid allocating 2 more lists.

mateuszrzeszutek (approved):

splunk-otel-android-volley/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+First, you will create an instance of `VolleyTracing` by passing your `splunkRum` instance
+to the builder:
+
+```java
+VolleyTracing volleyTracing = VolleyTracing.builder(splunkRum).build();
+```
+You can also add extra request/response headers to capture to the builder.
+These headers will be attached to the RUM Volley HTTP client spans with
+a`http.response.header.` prefix.
```
laurit, line 37: Prefix is given only for response headers although previous sentence mentions both request and response headers. There should be a space between a`http.response.header.`

splunk-otel-android-volley/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Then, use the `hurlStack` to create your request queue,
+
+```java
+RequestQueue queue = new RequestQueue(new NoCache(), new BasicNetwork(hurlStack), 1,
+                        new ExecutorDelivery(Executors.newSingleThreadExecutor()));
+queue.start();
+```
+
+To make instrumented Volley HTTP requests, just use the `RequestQueue` like you normally would!
```
laurit, line 53: I suspect that you should not be using an exclamation mark here. I would have stopped at `hurlStack`, we can assume that the users know how to use this library and can figure out the rest on their own.

laurit (approved):

splunk-otel-android-volley/README.md
```diff
@@ -0,0 +1,59 @@
...
+    ...
+}
+```
+
+## How to use VolleyTracing
+
+See the class `VolleyExample.java` file in the sample application in this repository for an
+example of how to create an instrumented client HTTP call.
+
+First, you will create an instance of `VolleyTracing` by passing your `splunkRum` instance
```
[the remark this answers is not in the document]
breedx-splk: Because that's how the parameter name is spelled. If you think the type name is more important for docs like this, I can change it.

splunk-otel-android-volley/README.md
breedx-splk: Good catch. Thanks.

splunk-otel-android-volley/README.md
breedx-splk: Ok, I'll just drop that last bit and shorten it. Thanks.

splunk-otel-android-volley/src/main/java/com/splunk/rum/VolleyHttpClientAttributesGetter.java
mateuszrzeszutek: > I've always considered mutable parameters to be an antipattern/bad practice. I'd also call this premature optimization...

They're absolutely terrible in a public API, I 100% agree.
I sometimes use them in private, hidden away code, in some cases it's a useful trick.

breedx-splk: I went ahead and merged and will happily review a follow-up PR if you choose to submit one. 🙃

PR #330, opened 2022-08-10
Update badges

breedx-splk:
* Add GDI version
* Add OTEL version
* Add MAVEN-CENTRAL version
* Add RELEASE version

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ---
 
-<p style="text-align: center">
+<p align="center">
   <strong>
     <a href="#getting-started">Getting Started</a>
     &nbsp;&nbsp;&bull;&nbsp;&nbsp;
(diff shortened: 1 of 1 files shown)


aurbiztondo-splunk (approved):

laurit:
Looking at https://github.com/breedx-splk/splunk-otel-android/blob/update_badges/README.md it doesn't seem to render properly. https://github.com/github/markup indicates that inline styles are not supported.

mateuszrzeszutek:
Yeah, splunk-otel-java still uses `<p align="center">` even though it's a deprecated HTML attribute: https://github.com/signalfx/splunk-otel-java/blob/main/README.md?plain=1#L3

breedx-splk:
> Looking at https://github.com/breedx-splk/splunk-otel-android/blob/update_badges/README.md it doesn't seem to render properly. https://github.com/github/markup indicates that inline styles are not supported.

Ah ok, wild, it renders in IntelliJ just fine. I'll flip it back to the deprecated way.

mateuszrzeszutek (approved):

laurit (approved):

PR #341, opened 2022-08-29
Remove stateful mutable field from Config

mateuszrzeszutek:
The first (of many...) API cleanup PRs

### DIFF
--- a/splunk-otel-android/src/test/java/com/splunk/rum/SplunkRumTest.java
+++ b/splunk-otel-android/src/test/java/com/splunk/rum/SplunkRumTest.java
@@ -63,7 +63,7 @@ public class SplunkRumTest {
 
     private Tracer tracer;
     private final AtomicReference<Attributes> globalAttributes =
-            new AtomicReference<>(Attributes.empty());
+            new AtomicReference<>(Attributes.of(stringKey("key"), "value"));
 
     @Before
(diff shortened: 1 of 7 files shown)


splunk-otel-android/src/test/java/com/splunk/rum/SplunkRumTest.java
```diff
@@ -59,13 +60,10 @@
...
+
     private Tracer tracer;
-    private final Config config =
-            Config.builder()
-                    .realm("us0")
-                    .rumAccessToken("token")
-                    .applicationName("appName")
-                    .build();
+    private final AtomicReference<Attributes> globalAttributes =
+            new AtomicReference<>(Attributes.empty());
```
breedx-splk, line 66: I like to have a least one random attribute in test data like this to help ward off the case where the empty instance is being used on accident.

breedx-splk (approved):
Looks good to me. One small comment about test data.

splunk-otel-android/src/test/java/com/splunk/rum/SplunkRumTest.java
mateuszrzeszutek: Done. I had to create a separate attributes ref for the location tests though.

PR #347, opened 2022-09-06
Add the release tagging script

### LINKED ISSUE #345: [GDI Spec] GH Releases MUST use signed tags
Per https://github.com/signalfx/gdi-specification/blob/v1.4.0/specification/repository.md?plain=1#L130

> MUST use signed tags

Currently, no [GH Release](https://github.com/signalfx/splunk-otel-android/releases) has a signed tag.

Do you have some releasing docs or script to avoid such problems for future releases? Example from our Go distro: https://github.com/signalfx/splunk-otel-go/blob/main/RELEASING.md

mateuszrzeszutek:
Resolves #345

### DIFF
--- a/scripts/tag-release.sh
+++ b/scripts/tag-release.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOT_DIR="${SCRIPT_DIR}/../"
+cd ${ROOT_DIR}
+
+print_usage() {
+  cat <<EOF
+Usage: $(basename $0) release_version
+
+All versions MUST NOT begin with 'v'. Example: 1.2.3.
+EOF
+}
+
+if [[ $# < 1 ]]
+then
+  print_usage
+  exit 1
+fi
+
+release_version=$1
+release_tag="v$1"
+
+if [[ ! $release_version =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
+then
+  echo "Invalid release version: $release_version"
+  echo "Release version must follow the pattern major.minor.patch, e.g. 1.2.3"
+  exit 1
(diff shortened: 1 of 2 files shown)


scripts/tag-release.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [[ ! $release_version =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
+then
+  echo "Invalid release version: $release_version"
+  echo "Release version must follow the pattern major.minor.patch, e.g. 1.2.3"
+  exit 1
+fi
+
+git tag -m "Release $release_version" -s "$release_tag"
+git push origin "$release_tag"
```
pellared, line 34: Is `origin` for all of you pointing to https://github.com/signalfx/splunk-otel-android?
Personally, I use `origin` for fork and `upstream` for the "original" repo

scripts/tag-release.sh
```diff
@@ -0,0 +1,34 @@
...
+release_tag="v$1"
+
+if [[ ! $release_version =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
+then
+  echo "Invalid release version: $release_version"
+  echo "Release version must follow the pattern major.minor.patch, e.g. 1.2.3"
+  exit 1
+fi
+
+git tag -m "Release $release_version" -s "$release_tag"
```
pellared, line 33: Consider adding the "commit" (as the last argument). It is kind of a "double-check" that you are tagging exactly what you want.

pellared (commented):
In general LGTM.
I just left some remarks.

scripts/tag-release.sh
mateuszrzeszutek: I don't want to make it too complex and tedious; I agree that it will probably be safer, but am willing to keep the simpler version for now. We've never tagged a wrong commit by mistake until now.

aurbiztondo-splunk (approved):

PR #354, opened 2022-09-14
Add nullaway and errorprone checks

mateuszrzeszutek:
I thought it would be useful to add null-safety checking; especially so when we're publishing an API. Errorprone did pick up several minor bugs (like missing charsets, broken double checked locking, etc), but it also made using our internal builder classes kinda painful, as most of them have `@Nullable` fields for components that must be present -- I think we could refactor these in a subsequent PR to make them a bit nicer.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
@@ -34,8 +34,8 @@ class AppStartupTimer {
     final RumInitializer.AnchoredClock startupClock =
             RumInitializer.AnchoredClock.create(Clock.getDefault());
     private final long firstPossibleTimestamp = startupClock.now();
-    private volatile Span overallAppStartSpan = null;
-    private volatile Runnable completionCallback = null;
+    @Nullable private volatile Span overallAppStartSpan = null;
+    @Nullable private volatile Runnable completionCallback = null;
     // whether activity has been created
     // accessed only from UI thread
     private boolean uiInitStarted = false;
@@ -78,6 +78,7 @@ class AppStartupTimer {
     }
 
     void end() {
+        Span overallAppStartSpan = this.overallAppStartSpan;
         if (overallAppStartSpan != null && !uiInitTooLate && !isStartedFromBackground) {
             runCompletionCallback();
             overallAppStartSpan.end(startupClock.now(), TimeUnit.NANOSECONDS);
@@ -92,6 +93,7 @@ class AppStartupTimer {
 
     // visibleForTesting
     void runCompletionCallback() {
+        Runnable completionCallback = this.completionCallback;
         if (completionCallback != null) {
             completionCallback.run();
         }
--- a/splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
@@ -32,7 +32,7 @@ class BandwidthTracker {
     private final ArrayDeque<Long> sizes = new ArrayDeque<>();
 
     BandwidthTracker() {
-        this(Clock.systemDefaultZone());
+        this(Clock.systemUTC());
     }
 
     // Exists for testing
--- a/splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
@@ -41,11 +41,11 @@ class FileSender {
     private final BandwidthTracker bandwidthTracker;
     private final RetryTracker retryTracker;
 
-    private FileSender(Builder builder, RetryTracker retryTracker) {
+    private FileSender(Builder builder) {
         this.sender = requireNonNull(builder.sender);
         this.fileUtils = builder.fileUtils;
         this.bandwidthTracker = requireNonNull(builder.bandwidthTracker);
-        this.retryTracker = retryTracker;
+        this.retryTracker = builder.buildRetryTracker();
     }
 
     /**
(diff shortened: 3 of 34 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
```diff
@@ -78,6 +78,7 @@ void setCompletionCallback(Runnable completionCallback) {
     }
 
     void end() {
+        Span overallAppStartSpan = this.overallAppStartSpan;
```
breedx-splk, line 81: This is an interesting fix to an easily overlooked problem! Did one of the tools catch this?

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
```diff
@@ -32,7 +32,7 @@ class BandwidthTracker {
     private final ArrayDeque<Long> sizes = new ArrayDeque<>();
 
     BandwidthTracker() {
-        this(Clock.systemDefaultZone());
+        this(Clock.systemUTC());
```
breedx-splk, line 35: Why does this matter? Does it prevent problems around time changes or ?

splunk-otel-android/src/main/java/com/splunk/rum/FileSender.java
```diff
@@ -39,11 +41,11 @@ class FileSender {
     private final BandwidthTracker bandwidthTracker;
     private final RetryTracker retryTracker;
 
-    private FileSender(Builder builder) {
-        this.sender = builder.sender;
+    private FileSender(Builder builder, RetryTracker retryTracker) {
```
breedx-splk, line 44: This feels weird to me. Maybe rather than the builder passing it in, maybe the `FileSender` constructor should create it based on the fields on the builder? Maybe that violates DI...but the mixed-mode constructor here (builder param with bonus param) feels odd.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -174,7 +176,7 @@ SplunkRum initialize(ConnectionUtil.Factory connectionUtilFactory, Looper mainLo
     private SlowRenderingDetector buildSlowRenderingDetector(Tracer tracer) {
         if (!builder.slowRenderingDetectionEnabled) {
             Log.w(LOG_TAG, "Slow/frozen rendering detection has been disabled by user.");
-            return SlowRenderingDetector.NO_OP;
+            return NoOpSlowRenderingDetector.INSTANCE;
```
breedx-splk, line 179: Was there a problem detected with this or just aesthetic preference?

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -441,23 +442,24 @@ long now() {
...
-            synchronized (this) {
-                if (delegate == null) {
-                    delegate = s.get();
+            SpanExporter d = delegate;
+            if (d == null) {
+                synchronized (this) {
+                    d = delegate;
+                    if (d == null) {
+                        delegate = d = s.get();
+                    }
```
breedx-splk, line 459: This construct always gives me pause because it was broken for so long...and it's fairly complicated to reason about. Since the volatile is required for it to be "correct" I wonder if just using a lock or synchronizing the method isn't just simpler/better. I won't lose sleep over it tho.

breedx-splk (approved):
Had a few ideas/questions but no showstoppers. Thanks for getting us one step closer to a 1.0.0 release! 🎈

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
mateuszrzeszutek: This one's pretty interesting, there wasn't any problem that errorprone detected, but something I noticed when I was doing `instrumentation-api` cleanup: classes defined inside interfaces (like `SlowRenderingDetector`) are always public, even if the interface itself is package-private. We were accidentally making the noop implementation public 😅

splunk-otel-android/src/main/java/com/splunk/rum/BandwidthTracker.java
mateuszrzeszutek: It triggers this check: https://errorprone.info/bugpattern/JavaTimeDefaultTimeZone
Which probably does not matter at all for this project, but since the only thing we need the clock for is to get the current millis, it can be any clock, we don't care about the timezone.

splunk-otel-android/src/main/java/com/splunk/rum/AppStartupTimer.java
mateuszrzeszutek: Yep! NullAway actually -- because the second time you reference the field, it might as well have a different value.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
mateuszrzeszutek: It's worth noting that this whole thing will be replaced by just one word `lazy` when we finally migrate to Kotlin.

> I wonder if just using a lock or synchronizing the method isn't just simpler/better.

It probably is a lot simpler, but I don't know enough about the Android JVM/Dalvik/whatever to say how slower this will be; a volatile field access is supposed to be just one instruction, no idea about the `sychronized` monitor.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
breedx-splk: 1000 babies saved!

PR #358, opened 2022-09-20
Add carrier information to spans

breedx-splk:
This attempts to use the `TelephonyManager` to get the following carrier information and add it to spans as attributes:

* carrier name (semconv `net.host.carrier.name`)
* carrier country code (semconv `net.host.carrier.mcc`)
* carrier network code (semconv `net.host.carrier.mnc`)
* carrier network iso country code (2 char, semconv `net.host.carrier.icc`)

Even thought `TelephonyManager` was introduced in API 1, the methods used here were added in API 28. As a result, we guard against 28 (P) and only use the `CarrierFinder` from `PostApi29NetworkDetector`.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
@@ -22,7 +22,7 @@ import androidx.annotation.Nullable;
 import androidx.annotation.RequiresApi;
 import java.util.Objects;
 
-@RequiresApi(api = Build.VERSION_CODES.Q)
+@RequiresApi(api = Build.VERSION_CODES.P)
 class Carrier {
 
     private final int id;
--- a/splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
@@ -1,11 +1,25 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.splunk.rum;
 
 import android.os.Build;
 import android.telephony.TelephonyManager;
-
 import androidx.annotation.Nullable;
 import androidx.annotation.RequiresApi;
-
 import java.util.Objects;
 
 @RequiresApi(api = Build.VERSION_CODES.Q)
@@ -58,7 +72,11 @@ class Carrier {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         Carrier carrier = (Carrier) o;
-        return id == carrier.id && Objects.equals(name, carrier.name) && Objects.equals(mobileCountryCode, carrier.mobileCountryCode) && Objects.equals(mobileNetworkCode, carrier.mobileNetworkCode) && Objects.equals(isoCountryCode, carrier.isoCountryCode);
+        return id == carrier.id
+                && Objects.equals(name, carrier.name)
+                && Objects.equals(mobileCountryCode, carrier.mobileCountryCode)
+                && Objects.equals(mobileNetworkCode, carrier.mobileNetworkCode)
+                && Objects.equals(isoCountryCode, carrier.isoCountryCode);
     }
 
     @Override
@@ -73,7 +91,7 @@ class Carrier {
         private @Nullable String mobileNetworkCode = null;
         private @Nullable String isoCountryCode = null;
 
-        Carrier build(){
+        Carrier build() {
             return new Carrier(this);
         }
 
@@ -102,5 +120,4 @@ class Carrier {
             return this;
         }
     }
-
 }
--- a/splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
@@ -1,0 +1,106 @@
+package com.splunk.rum;
+
+import android.os.Build;
+import android.telephony.TelephonyManager;
+
+import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
+
+import java.util.Objects;
+
+@RequiresApi(api = Build.VERSION_CODES.Q)
+class Carrier {
+
+    private final int id;
+    private final @Nullable String name;
+    private final @Nullable String mobileCountryCode; // 3 digits
+    private final @Nullable String mobileNetworkCode; // 2 or 3 digits
+    private final @Nullable String isoCountryCode;
+
+    static Builder builder() {
+        return new Builder();
+    }
+
+    Carrier(Builder builder) {
+        this.id = builder.id;
+        this.name = builder.name;
+        this.mobileCountryCode = builder.mobileCountryCode;
+        this.mobileNetworkCode = builder.mobileNetworkCode;
+        this.isoCountryCode = builder.isoCountryCode;
+    }
+
+    int getId() {
+        return id;
+    }
+
+    @Nullable
+    String getName() {
+        return name;
+    }
+
+    @Nullable
+    String getMobileCountryCode() {
+        return mobileCountryCode;
+    }
+
+    @Nullable
+    String getMobileNetworkCode() {
+        return mobileNetworkCode;
+    }
+
+    @Nullable
+    String getIsoCountryCode() {
+        return isoCountryCode;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Carrier carrier = (Carrier) o;
+        return id == carrier.id && Objects.equals(name, carrier.name) && Objects.equals(mobileCountryCode, carrier.mobileCountryCode) && Objects.equals(mobileNetworkCode, carrier.mobileNetworkCode) && Objects.equals(isoCountryCode, carrier.isoCountryCode);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name, mobileCountryCode, mobileNetworkCode, isoCountryCode);
+    }
+
+    static class Builder {
+        private int id = TelephonyManager.UNKNOWN_CARRIER_ID;
+        private @Nullable String name = null;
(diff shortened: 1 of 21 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
```diff
@@ -16,15 +16,30 @@
...
+
+    CurrentNetwork(Carrier carrier, NetworkState state) {
+        this(carrier, state, null);
+    }
+
     CurrentNetwork(NetworkState state, @Nullable String subType) {
+        this(null, state, subType);
+    }
+
+    CurrentNetwork(@Nullable Carrier carrier, NetworkState state, @Nullable String subType) {
```
mateuszrzeszutek, line 41: Nit: do we need a builder for this class?

splunk-otel-android/src/main/java/com/splunk/rum/CarrierFinder.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        String isoCountryCode = telephonyManager.getSimCountryIso();
+        if (validString(isoCountryCode)) {
+            builder.isoCountryCode(isoCountryCode);
+        }
+        return builder.build();
+    }
+
+    private boolean validString(String str) {
+        return !(str == null || str.isEmpty());
```
mateuszrzeszutek, line 54: You could have `validString()` accept a `CharSequence` and avoid double null check in line 37:
```suggestion
    private boolean validString(CharSequence str) {
        return !(str == null || str.length() == 0);
```

splunk-otel-android/src/main/java/com/splunk/rum/PostApi29NetworkDetector.java
```diff
@@ -30,18 +30,21 @@
 import androidx.annotation.RequiresApi;
 import androidx.core.app.ActivityCompat;
 
-@RequiresApi(api = Build.VERSION_CODES.Q)
+@RequiresApi(api = Build.VERSION_CODES.P)
 class PostApi29NetworkDetector implements NetworkDetector {
```
mateuszrzeszutek, line 34: Should we rename the class to `PostApi28...`?

splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
```diff
@@ -84,9 +89,18 @@ public void onStart(Context parentContext, ReadWriteSpan span) {
         span.setAttribute(SplunkRum.SCREEN_NAME_KEY, currentScreen);
         CurrentNetwork currentNetwork = connectionUtil.getActiveNetwork();
         span.setAttribute(NET_HOST_CONNECTION_TYPE, currentNetwork.getState().getHumanName());
+        appendNetworkAttributes(span, currentNetwork);
+    }
+
+    @SuppressWarnings("NullAway")
+    static void appendNetworkAttributes(Span span, CurrentNetwork currentNetwork) {
```
mateuszrzeszutek, line 96: Should we include `NET_HOST_CONNECTION_TYPE` too? It's being set in both places

splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
```diff
@@ -48,18 +63,44 @@ public String toString() {
...
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(state, subType);
+        return Objects.hash(carrier, state, subType);
+    }
+
+    @SuppressWarnings("NullAway")
+    @Nullable
```
mateuszrzeszutek, line 80: `getSubType()` returns an `Optional` -- should we make it return a `@Nullable` value instead? So that the API in this class looks uniform

splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
```diff
@@ -48,18 +63,44 @@ public String toString() {
```
mateuszrzeszutek, line 63: Should we modify the `toString()` method of this class to reflect carrier?

splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
```diff
@@ -0,0 +1,123 @@
...
+    Carrier(Builder builder) {
+        this.id = builder.id;
+        this.name = builder.name;
+        this.mobileCountryCode = builder.mobileCountryCode;
+        this.mobileNetworkCode = builder.mobileNetworkCode;
+        this.isoCountryCode = builder.isoCountryCode;
+    }
+
+    int getId() {
+        return id;
```
mateuszrzeszutek, line 47: Is `id` used anywhere in the code? Or is it just for `equals()`/`hashCode()` comparisons?

splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
```diff
@@ -84,9 +89,18 @@ public void onStart(Context parentContext, ReadWriteSpan span) {
...
         currentNetwork
                 .getSubType()
                 .ifPresent(subtype -> span.setAttribute(NET_HOST_CONNECTION_SUBTYPE, subtype));
+        span.setAttribute(NET_HOST_CARRIER_NAME, currentNetwork.getCarrierName());
```
mateuszrzeszutek, line 100: NullAway complains about these, doesn't it?
I think we should null-check here and follow what the API says, even if we're pretty sure it can handle nulls.

splunk-otel-android/src/test/java/com/splunk/rum/PostApi29NetworkDetectorTest.java
```diff
@@ -35,121 +37,109 @@
 @Config(sdk = Build.VERSION_CODES.Q)
```
mateuszrzeszutek, line 37: Does it work on Android P?

mateuszrzeszutek (commented):
Left a couple of comments; overall it looks good! 👍

splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
breedx-splk: Yeah, might help actually with all the constructors. Lemme try it.

splunk-otel-android/src/main/java/com/splunk/rum/CarrierFinder.java
breedx-splk: Yeah good call thanks.

splunk-otel-android/src/main/java/com/splunk/rum/PostApi29NetworkDetector.java
breedx-splk: Absolutely. Good call.

splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
breedx-splk: Great catch! Will do.

splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
breedx-splk: Yeah, not sure why it was that way originally. Will make consistent and have a null party.

splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
breedx-splk: It's not. I thought initially we might ingest it as well, but since we have the other components we could drop it. How strongly do you feel about getting rid of it? Seems fairly harmless.

splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java
breedx-splk: Hard meh. Will do.

splunk-otel-android/src/test/java/com/splunk/rum/PostApi29NetworkDetectorTest.java
breedx-splk: Yeah, I had missed that one. I'll change it.

splunk-otel-android/src/main/java/com/splunk/rum/CarrierFinder.java
```diff
@@ -0,0 +1,56 @@
...
+    CarrierFinder(TelephonyManager telephonyManager) {
+        this.telephonyManager = telephonyManager;
+    }
+
+    Carrier get() {
+        Carrier.Builder builder = Carrier.builder();
+        int id = telephonyManager.getSimCarrierId();
+        builder.id(id);
+        CharSequence name = telephonyManager.getSimCarrierIdName();
+        if (validString(name.toString())) {
```
mateuszrzeszutek, line 37: Should probably be:
```suggestion
        if (validString(name)) {
```
I think you're getting an NPE in tests here

splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
```diff
@@ -16,19 +16,28 @@
...
 import java.util.Objects;
-import java.util.Optional;
 
 final class CurrentNetwork {
+    @Nullable private final Carrier carrier;
     private final NetworkState state;
     @Nullable private final String subType;
 
-    CurrentNetwork(NetworkState state, @Nullable String subType) {
+    private CurrentNetwork(
```
mateuszrzeszutek, line 28: Is this constructor used? It's private, doesn't seem to be referenced within this class

splunk-otel-android/src/main/java/com/splunk/rum/NetworkMonitor.java
```diff
@@ -78,19 +77,8 @@ public void onAvailable(boolean deviceIsOnline, CurrentNetwork activeNetwork) {
                 Span available =
                         tracer.spanBuilder("network.change")
                                 .setAttribute(NETWORK_STATUS_KEY, "available")
-                                .startSpan()
-                                // put these after span start to override what might be set in the
-                                // RumAttributeAppender.
```
mateuszrzeszutek, line 83: I'd leave the comment -- it does provide some additional explanation for why the `appendNetworkAttributes()` does not accept a `SpanBuilder` (or some kind of `setAttribute()` lambda interface, common for both span and builder)

splunk-otel-android/src/main/java/com/splunk/rum/Carrier.java
mateuszrzeszutek: I don't feel that strongly about it; I just thought "is this one even used?"

mateuszrzeszutek (approved):
👍

splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
breedx-splk: Oops, I think it was left over after the builder was completed. I'll remove.

PR #361, opened 2022-09-23
Refactor `SlowRenderingDetectorImpl` so that it's thread-safe

### LINKED ISSUE #359: Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=0; index=3
We're using version 0.16.0 and we're getting some crashes at SlowRenderingDetectorImpl:

```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=0; index=3
       at androidx.core.app.FrameMetricsAggregator$FrameMetricsApi24Impl$1.onFrameMetricsAvailable(FrameMetricsAggregator.java:368)
       at android.view.FrameMetricsObserver.onFrameMetricsAvailable(FrameMetricsObserver.java:59)
       at android.graphics.HardwareRendererObserver.lambda$notifyDataAvailable$0$HardwareRendererObserver(HardwareRendererObserver.java:93)
       at android.graphics.HardwareRendererObserver$$ExternalSyntheticLambda0.run(HardwareRendererObserver.java:2)
       at android.os.Handler.handleCallback(Handler.java:978)
       at android.os.Handler.dispatchMessage(Handler.java:104)
       at android.os.Looper.loopOnce(Looper.java:238)
       at android.os.Looper.loop(Looper.java:357)
       at android.os.HandlerThread.run(HandlerThread.java:67)
```

mateuszrzeszutek:
Fixes #359

Removed the `FrameMetricsAggregator`, because it kept internal state that was not subject to any sort of synchronization whatsoever, and was accessed from multiple threads in an unsafe way. I refactored (simplified perhaps?) `SlowRenderingDetectorImpl` so that it uses `FrameMetrics` directly -- which, what's interesting, will allow us to collect per-activity metrics in the future, if we wish to do so.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
@@ -178,19 +178,16 @@ class RumInitializer {
             Log.w(LOG_TAG, "Slow/frozen rendering detection has been disabled by user.");
             return NoOpSlowRenderingDetector.INSTANCE;
         }
-        try {
-            initializationEvents.add(
-                    new RumInitializer.InitializationEvent(
-                            "slowRenderingDetectorInitialized", timingClock.now()));
-            Class.forName("androidx.core.app.FrameMetricsAggregator");
-            return new SlowRenderingDetectorImpl(
-                    tracer, builder.slowRenderingDetectionPollInterval);
-        } catch (ClassNotFoundException e) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
             Log.w(
                     LOG_TAG,
-                    "FrameMetricsAggregator is not available on this platform - slow/frozen rendering detection is disabled.");
+                    "Slow/frozen rendering detection is not supported on platforms older than Android N (SDK version 24).");
             return NoOpSlowRenderingDetector.INSTANCE;
         }
+        initializationEvents.add(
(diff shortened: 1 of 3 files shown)


breedx-splk:
> Removed the `FrameMetricsAggregator`, because it kept internal state that was not subject to any sort of synchronization whatsoever, and was accessed from multiple threads in an unsafe way.

This is interesting because every usage of the `frameMetrics` instance within `SlowRenderingDetectorImpl` was definitely guarded by the `lock`, so I'm curious about where you observed unsafe accesses? Maybe you were thinking about the lack of coordination between the internal synchronization and our usage of it?

breedx-splk (approved):
It makes me a little nervous to kinda reimplement some of what already exists in the guts of the platform sdk class `FrameMetricsAggregator` and how this could be impacted in the future. However, I do appreciate the sensitivity to thread-safety and a desire to make that better.

Aside from the reimplementation, I don't see anything to stop this change (but I would still like to better understand where the original thread safety concern came from). In the interest in getting the RC build out, let's go with this.

mateuszrzeszutek:
> (but I would still like to better understand where the original thread safety concern came from).

`FrameMetricsAggregator` internally started a thread (a `HandlerThread` for running the frame metrics listener) that modified the `SparseIntArray[] metrics` field without any sort of synchronization. So, while we were synchronizing all the reads in our own, the writes happening in the core library were completely unguarded.

PR #371, opened 2022-10-04
Bump desugar_jdk_libs version to 1.2.2

mateuszrzeszutek:
We can update this one now, after we upgraded AGP to 7.3


breedx-splk:
What a wacky chain of events. I appreciate that you've stayed on top of it...this would be easy to miss.

breedx-splk (approved):

PR #373, opened 2022-10-07
Include activity name in slow and frozen render reports

breedx-splk:
Adds the `activity.name` field to slow and frozen render reports.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
@@ -92,7 +92,7 @@ class SlowRenderingDetectorImpl implements SlowRenderingDetector {
     @Override
     public void add(Activity activity) {
         PerActivityListener listener = new PerActivityListener(activity);
-        PerActivityListener existing = activities.put(activity, listener);
+        PerActivityListener existing = activities.putIfAbsent(activity, listener);
         if (existing == null) {
             activity.getWindow().addOnFrameMetricsAvailableListener(listener, frameMetricsHandler);
         }
--- a/splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
@@ -33,15 +33,14 @@ import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.Tracer;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
 @RequiresApi(api = Build.VERSION_CODES.N)
-class SlowRenderingDetectorImpl
-        implements SlowRenderingDetector, Window.OnFrameMetricsAvailableListener {
+class SlowRenderingDetectorImpl implements SlowRenderingDetector {
 
     static final int SLOW_THRESHOLD_MS = 16;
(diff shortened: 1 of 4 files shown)


splunk-otel-android/src/test/java/com/splunk/rum/SlowRenderingDetectorImplTest.java
```diff
@@ -79,8 +83,12 @@ public void add() {
 
         testInstance.add(activity);
 
+        ArgumentCaptor<SlowRenderingDetectorImpl.PerActivityListener> captor =
+                ArgumentCaptor.forClass(SlowRenderingDetectorImpl.PerActivityListener.class);
+
         verify(activity.getWindow())
-                .addOnFrameMetricsAvailableListener(testInstance, frameMetricsHandler);
+                .addOnFrameMetricsAvailableListener(captor.capture(), eq(frameMetricsHandler));
+        captor.getValue().getActivityName();
```
mateuszrzeszutek, line 91: Do you want to assert on this?

splunk-otel-android/src/test/java/com/splunk/rum/SlowRenderingDetectorImplTest.java
```diff
@@ -115,9 +126,14 @@ public void removeWithMetrics() {
 
         testInstance.add(activity);
 
+        ArgumentCaptor<SlowRenderingDetectorImpl.PerActivityListener> captor =
+                ArgumentCaptor.forClass(SlowRenderingDetectorImpl.PerActivityListener.class);
+
         for (long duration : makeSomeDurations()) {
             when(frameMetrics.getMetric(FrameMetrics.DRAW_DURATION)).thenReturn(duration);
-            testInstance.onFrameMetricsAvailable(null, frameMetrics, 0);
+            verify(activity.getWindow())
```
mateuszrzeszutek, line 134: This `verify` call could be made before the loop

splunk-otel-android/src/test/java/com/splunk/rum/SlowRenderingDetectorImplTest.java
```diff
@@ -145,9 +161,14 @@ public void start() {
...
+                    .addOnFrameMetricsAvailableListener(captor.capture(), any());
```
mateuszrzeszutek, line 170: Same here, it could be moved before the loop

splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
```diff
@@ -97,25 +91,20 @@ private static Looper startFrameMetricsLoop() {
...
     @Override
     public void add(Activity activity) {
-        boolean added;
-        synchronized (lock) {
-            added = activities.add(activity);
-        }
-        if (added) {
-            activity.getWindow().addOnFrameMetricsAvailableListener(this, frameMetricsHandler);
+        PerActivityListener listener = new PerActivityListener(activity);
+        PerActivityListener existing = activities.put(activity, listener);
```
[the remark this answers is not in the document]
breedx-splk: This is a good call. I actually think this is more important than we both had realized. In my initial `.put()` implementation, the map will be updated with a new listener, but the listener will _not_ be registered. Then later, when we `.remove()` the activity from the map and get the corresponding listener, I suspect the listener will not be removed correctly because of no equals/hashcode. 

In any case, your suggestion is more correct, and even tho it shouldn't really happen, we should do this. 🤘🏻

splunk-otel-android/src/test/java/com/splunk/rum/SlowRenderingDetectorImplTest.java
breedx-splk: Yeah, what? Heh, will fix.

splunk-otel-android/src/test/java/com/splunk/rum/SlowRenderingDetectorImplTest.java
breedx-splk: cool, probably reads a little cleaner that way.

mateuszrzeszutek (approved):

PR #377, opened 2022-10-11
Minor API fix

mateuszrzeszutek:
Renamed `disableNetworkMonitorEnabled` to `disableNetworkMonitor`; moved a static method higher, where other static methods are located


breedx-splk (approved):
🙈

PR #380, opened 2022-10-12
Decouple SlowRenderingDetector from ActivityCallbacks

mateuszrzeszutek:
Another minor change to make extracting out OTel things easier

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
@@ -19,7 +19,9 @@ package com.splunk.rum;
 import static com.splunk.rum.SplunkRum.LOG_TAG;
 
 import android.app.Activity;
+import android.app.Application;
 import android.os.Build;
+import android.os.Bundle;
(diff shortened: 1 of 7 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/SlowRenderingDetectorImpl.java
```diff
@@ -99,24 +124,22 @@ public void add(Activity activity) {
...
-        executorService.scheduleAtFixedRate(
-                this::reportSlowRenders,
-                pollInterval.toMillis(),
-                pollInterval.toMillis(),
-                TimeUnit.MILLISECONDS);
-    }
+    public void onActivityStopped(@NonNull Activity activity) {}
+
+    @Override
+    public void onActivitySaveInstanceState(@NonNull Activity activity, @NonNull Bundle outState) {}
```
breedx-splk, line 139: Really makes you wish that the interface had default nop implementations, huh? 😉

mateuszrzeszutek: Oh yes

PR #381, opened 2022-10-12
Experimental exporter to support react native.

mhennoch:
RN client side (the thread that runs JS) will send already completed spans to native side. There we need to recreate the span so it goes through android-sdk span processing pipeline in order to get all of the attributes set on that span. A very alpha version of how that span is constructed can be seen [here](https://github.com/mhennoch/react-native-basic-rn-module/blob/master/android/src/main/java/com/reactnativebasicrnmodule/BasicRnModuleModule.java#L167). As there is no way of changing the spanId and traceId this exporter is needed. Based on `SpanDataModifier`.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/reactnative/RNSpanData.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/reactnative/RNSpanData.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/reactnative/RNSpanData.java
```diff
@@ -0,0 +1,73 @@
...
+public final class RNSpanData extends DelegatingSpanData {
+
+    private final SpanContext modifiedContext;
+    private final Attributes modifiedAttributes;
+
+    static SpanData create(SpanData original) {
+        SpanContext originalSpanContext = original.getSpanContext();
+        Attributes attributes = original.getAttributes();
+        AttributeKey<String> traceIdKey = AttributeKey.stringKey("tmp.traceId");
+        AttributeKey<String> spanIdKey = AttributeKey.stringKey("tmp.spanId");
```
mateuszrzeszutek, line 35: Nit: WDYT about using a bit more descriptive names for the trace/span ID?
```suggestion
        AttributeKey<String> traceIdKey = AttributeKey.stringKey("_reactnative_traceId");
        AttributeKey<String> spanIdKey = AttributeKey.stringKey("_reactnative_spanId");
```
Also the underscore at the beginning makes it look a bit more "private"/"internal"

mateuszrzeszutek (approved):

mateuszrzeszutek:
LGTM, let's merge it as soon as possible -- I'll refactor the `SpanData` modifications after #376 is merged, and make sure that we don't expose to much public API stuff

PR #385, opened 2022-10-13
Refactor react native classes

mateuszrzeszutek:
Waiting for #376


splunk-otel-android/src/test/java/com/splunk/rum/SplunkSpanDataModifierTest.java
```diff
@@ -195,4 +195,95 @@ public void shouldCopySelectedResourceAttributes() {
...
+        Collection<SpanData> exportedSpans = exportedSpansCaptor.getValue();
+        assertThat(exportedSpans).hasSize(1);
+        assertThat(exportedSpans.iterator().next())
+                .hasName("SplunkRumSpan")
+                .hasTraceId(spanContext.getTraceId())
+                .hasSpanId(spanContext.getSpanId())
+                .hasAttributesSatisfyingExactly(
+                        equalTo(SPLUNK_OPERATION_KEY, "SplunkRumSpan"),
+                        equalTo(REACT_NATIVE_TRACE_ID_KEY, "99999999999999999999999999999999"),
+                        equalTo(REACT_NATIVE_SPAN_ID_KEY, "8888888888888888"));
```
breedx-splk, line 243: We still want to carry these forward even when reactive native support isn't active? I mean, I get that they're just attributes, but still....?

Not sure if this is a pro or con -- one the one hand, seeing these in the UI as attributes might at least help someone to understand where the span came from, but the downside is that it might just cause confusion. I dunno. Feels like a leaky implementation detail.

splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java
```diff
@@ -139,6 +138,14 @@ public SplunkRumBuilder enableDiskBuffering() {
         return this;
     }
 
+    /**
+     * Enables support for the React Native instrumentation.
+     *
+     * <p>This feature is disabled by default. You can enable it by calling this configuration
+     * method with a {@code true} value.
```
breedx-splk, line 145:
```suggestion
     * <p>This feature is disabled by default. You can enable it by calling this method.
```

splunk-otel-android/src/main/java/com/splunk/rum/SplunkSpanDataModifier.java
```diff
@@ -74,6 +81,15 @@ private SpanData modify(SpanData original) {
         List<EventData> modifiedEvents = new ArrayList<>(original.getEvents().size());
         AttributesBuilder modifiedAttributes = original.getAttributes().toBuilder();
 
+        SpanContext spanContext;
+        if (reactNativeEnabled) {
+            spanContext = extractReactNativeIdsIfPresent(original);
+            modifiedAttributes.remove(REACT_NATIVE_TRACE_ID_KEY);
+            modifiedAttributes.remove(REACT_NATIVE_SPAN_ID_KEY);
```
breedx-splk, line 88: We can change this later, but I think it's unfortunate that we have to deal with the react stuff here. I wish this was done somehow via composition -- with another exporter or "data modifier" that wrapped this (or that this wrapped) only got wired up when RN support was enabled.

breedx-splk (approved):
Seems ok. We should fix that comment tho.

splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java
mateuszrzeszutek: 🤦

splunk-otel-android/src/test/java/com/splunk/rum/SplunkSpanDataModifierTest.java
mateuszrzeszutek: Honestly, no idea. My idea was to simply treat them as "normal" attributes in case RN support is off; but always dropping them is also not that bad of a choice.

splunk-otel-android/src/main/java/com/splunk/rum/SplunkSpanDataModifier.java
mateuszrzeszutek: I've done it this way to avoid creating another `SpanData` copy (cause this is Android, and we probably should be careful and avoid too many allocations on the hot path).
But, now that I think of it, I guess we could split this class into several parts (implementing an internal interface) and pass a mutable `SplunkSpanBuilder` between them - I'll explore this in another PR.

breedx-splk: something something premature optimization.... 🙃

PR #391, opened 2022-10-18
Migrate tests to junit5 

breedx-splk:
We have to keep junit4 (vintage) around because we (still) rely on RoboElectric and there is a long history of continued incompatibility there. Maybe one day.... 🤞🏻

### DIFF
--- a/splunk-otel-android/build.gradle.kts
+++ b/splunk-otel-android/build.gradle.kts
@@ -40,10 +40,6 @@ android {
     }
 }
 
-tasks.withType<Test> {
-    useJUnitPlatform()
-}
-
 val otelVersion = "1.19.0"
 val otelAlphaVersion = "$otelVersion-alpha"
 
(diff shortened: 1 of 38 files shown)


splunk-otel-android/build.gradle.kts
```diff
@@ -71,5 +79,9 @@ dependencies {
     coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:1.1.8")
 }
 
+tasks.withType<Test> {
+    useJUnitPlatform()
+}
+
```
mateuszrzeszutek, line 85: Duplicate of lines 43-46

mateuszrzeszutek (approved):
Awesome! 🚀

splunk-otel-android/build.gradle.kts
breedx-splk: yeah whoops. fill fix.

PR #392, opened 2022-10-19
Update License header

pellared:
1. "license (verb)" - grant license to something
1. I think we want to license not only the releases but also other artifacts such as the codebase


mateuszrzeszutek:
cc @theletterf

theletterf (approved):

PR #399, opened 2022-11-03
Extract ANR detection instrumentation

mateuszrzeszutek:
Another part of the upstreaming saga.

### DIFF
--- a/splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/anr/AnrDetectorToggle.java
+++ b/splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/anr/AnrDetectorToggle.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.rum.internal.instrumentation.anr;
+
+import androidx.annotation.Nullable;
+import io.opentelemetry.rum.internal.instrumentation.ApplicationStateListener;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+final class AnrDetectorToggle implements ApplicationStateListener {
+
+    private final Runnable anrWatcher;
+    private final ScheduledExecutorService anrScheduler;
+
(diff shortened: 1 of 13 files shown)


splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/anr/AnrDetectorToggle.java
```diff
@@ -0,0 +1,51 @@
...
+
+package io.opentelemetry.rum.internal.instrumentation.anr;
+
+import androidx.annotation.Nullable;
+import io.opentelemetry.rum.internal.instrumentation.ApplicationStateListener;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+final class AnrDetectorToggle implements ApplicationStateListener {
```
breedx-splk, line 25: Consider renaming to `AnrWatcherToggler`? I think it's a little more accurate. Also "toggle" as a noun doesn't really fit here, so "toggler" or "initiator" or something.

splunk-otel-android/src/test/java/io/opentelemetry/rum/internal/instrumentation/anr/AnrDetectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class AnrDetectorTest {
+
+    @Mock Looper mainLooper;
+    @Mock ScheduledExecutorService scheduler;
+    @Mock InstrumentedApplication instrumentedApplication;
+
+    @Captor ArgumentCaptor<AnrWatcher> anrWatcherCaptor;
```
breedx-splk, line 45: `@Captor` is a new annotation for me. Awesome!

splunk-otel-android/src/test/java/io/opentelemetry/rum/internal/instrumentation/anr/AnrDetectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+                        .setMainLooper(mainLooper)
+                        .setScheduler(scheduler)
+                        .addAttributesExtractor(constant(stringKey("test.key"), "abc"))
+                        .build();
+        anrDetector.installOn(instrumentedApplication);
+
+        // verify that the ANR scheduler was started
+        verify(scheduler)
+                .scheduleAtFixedRate(
+                        anrWatcherCaptor.capture(), eq(1L), eq(1L), eq(TimeUnit.SECONDS));
```
breedx-splk, line 63: If we don't care about the captured value, isn't it better just to use `isA(AnrWatcher.class)`?

breedx-splk (changes requested):
I love how the separation of concerns is really shaping up here! Nice decomposition for sure.

A couple of small suggestions, but looking good overall.

splunk-otel-android/src/test/java/io/opentelemetry/rum/internal/instrumentation/anr/AnrDetectorTest.java
mateuszrzeszutek: Oops, that was a leftover of my experiments with these tests. Removed!

breedx-splk (approved):
Thanks thanks! 🥳

PR #400, opened 2022-11-03
Don't include first frame in draw duration histogram for slow renders.

breedx-splk:
The javadoc in `FrameMetrics` says:
```
     * First draw frames are expected to be slow and should usually be exempt
     * from display jank calculations as they do not cause skips in animations
     * and are usually hidden by window animations or other tricks.
```


mateuszrzeszutek (approved):
👍

PR #402, opened 2022-11-04
Extract crash reporting instrumentation

mateuszrzeszutek:


splunk-otel-android/src/main/java/com/splunk/rum/RuntimeDetailsExtractor.java
```diff
@@ -22,44 +22,69 @@
 import android.content.IntentFilter;
 import android.os.BatteryManager;
 import androidx.annotation.Nullable;
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
 import java.io.File;
 
 /** Represents details about the runtime environment at a time */
-final class RuntimeDetails extends BroadcastReceiver {
+final class RuntimeDetailsExtractor<RQ, RS> extends BroadcastReceiver
```
mateuszrzeszutek, line 30: I left this entire class in the Splunk part of the codebase -- OTel will probably want to export this stuff as metrics, not span attributes.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/crash/CrashReportingExceptionHandler.java
```diff
@@ -0,0 +1,60 @@
...
+        this.existingHandler = existingHandler;
+    }
+
+    @Override
+    public void uncaughtException(@NonNull Thread t, @NonNull Throwable e) {
+        reportCrash(t, e);
+
+        // do our best to make sure the crash makes it out of the VM
+        CompletableResultCode flushResult = sdkTracerProvider.forceFlush();
+        flushResult.join(10, TimeUnit.SECONDS);
```
breedx-splk, line 47: 👍🏻

breedx-splk (approved):
Looks good. One small idea about the constructor/builder but otherwise nicely done!

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/crash/CrashReporterBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package io.opentelemetry.rum.internal.instrumentation.crash;
+
+import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
+import java.util.ArrayList;
+import java.util.List;
+
+/** A builder of {@link CrashReporter}. */
+public final class CrashReporterBuilder {
```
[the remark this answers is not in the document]
mateuszrzeszutek: We could, but I think I prefer following the general library instrumentation conventions here -- use builder for all optional arguments. I'd like all these instrumentations to look very similar to each other.

PR #407, opened 2022-11-14
Guard against empty files.

breedx-splk:
This allows empty (zero-byte) span files to be deleted when encountered, but also prevents creating them in the first place when the span batch is empty.


sample-app/src/main/java/com/splunk/android/sample/SampleApplication.java
```diff
@@ -55,6 +55,7 @@ public void onCreate() {
                                 spanFilter
                                         .removeSpanAttribute(stringKey("http.user_agent"))
                                         .rejectSpansByName(spanName -> spanName.contains("ignored"))
+                                        .rejectSpansByName(name -> !name.equals("nope"))
```
mateuszrzeszutek, line 58: Should we keep this change?

mateuszrzeszutek (approved):

PR #410, opened 2022-11-15
Refactor CurrentNetwork: add a method that returns all network Attrib…

mateuszrzeszutek:
…utes

I've started working on extracting the network monitoring instrumentation and it's super tangled; I'll maybe try to clean up the spaghetti in several smaller PRs.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
@@ -16,18 +16,8 @@
 
 package com.splunk.rum;
 
-import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_HOST_CARRIER_ICC;
-import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_HOST_CARRIER_MCC;
-import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_HOST_CARRIER_MNC;
-import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_HOST_CARRIER_NAME;
-import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_HOST_CONNECTION_SUBTYPE;
-import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_HOST_CONNECTION_TYPE;
-
 import android.os.Build;
 import androidx.annotation.Nullable;
-import io.opentelemetry.api.common.AttributeKey;
-import io.opentelemetry.api.common.Attributes;
-import io.opentelemetry.api.common.AttributesBuilder;
 import java.util.Objects;
 
 final class CurrentNetwork {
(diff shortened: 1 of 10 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
```diff
@@ -39,22 +49,30 @@ NetworkState getState() {
...
+
+        setIfNotNull(builder, NET_HOST_CONNECTION_SUBTYPE, subType);
+        if (haveCarrier()) {
+            setIfNotNull(builder, NET_HOST_CARRIER_NAME, carrier.getName());
+            setIfNotNull(builder, NET_HOST_CARRIER_MCC, carrier.getMobileCountryCode());
+            setIfNotNull(builder, NET_HOST_CARRIER_MNC, carrier.getMobileNetworkCode());
+            setIfNotNull(builder, NET_HOST_CARRIER_ICC, carrier.getIsoCountryCode());
+        }
+
+        return builder.build();
```
breedx-splk, line 64: I'm not excited about having this here. Previously, the `CurrentNetwork` class was a very simple data model object that simply pulled together the concepts of network carrier, network state, and network subtype. It is constructed simply and then exposes getters for the various specific details. It had no ties to instrumentation or to opentelemetry semantic conventions.  

I understand the need, but I think the mapping of domain-specific attributes into a telemetry system (otel or otherwise) is a separate responsibility. This responsibility should be outside of the data model class. If the `NetworkMonitor` doesn't feel like a good place, then maybe a new class that just does the mapping.

splunk-otel-android/src/test/java/com/splunk/rum/PostApi28NetworkDetectorTest.java
```diff
@@ -145,6 +146,7 @@ public void carrierIsSet() {
                 new PostApi28NetworkDetector(
                         connectivityManager, telephonyManager, carrierFinder, context);
         CurrentNetwork currentNetwork = networkDetector.detectCurrentNetwork();
-        assertThat(currentNetwork.getCarrierName()).isEqualTo("flib");
+        assertThat(currentNetwork.getNetworkAttributes())
+                .containsEntry(SemanticAttributes.NET_HOST_CARRIER_NAME, "flib");
```
breedx-splk, line 150: Yeah, so I think this fundamentally shows the problem around changing a data model object to have semantics in a new domain. Meh.

splunk-otel-android/src/main/java/com/splunk/rum/NetworkMonitor.java
```diff
@@ -81,7 +81,7 @@ public void onAvailable(boolean deviceIsOnline, CurrentNetwork activeNetwork) {
                                 .startSpan();
                 // put these after span start to override what might be set in the
                 // RumAttributeAppender.
-                RumAttributeAppender.appendNetworkAttributes(available, activeNetwork);
+                available.setAllAttributes(activeNetwork.getNetworkAttributes());
```
breedx-splk, line 84: Are the `NetworkMonitor` unit tests lacking? A little surprised that this didn't impact a unit test.

breedx-splk (commented):
I like separating the network-specific attributes from the giant singular attribute appender, but I think we should take a slightly different approach.

splunk-otel-android/src/main/java/com/splunk/rum/NetworkMonitor.java
mateuszrzeszutek: Yeah, we weren't setting the carrier there at all.

splunk-otel-android/src/main/java/com/splunk/rum/CurrentNetwork.java
mateuszrzeszutek: > maybe a new class that just does the mapping.

Hm, yeah, I think I'll just do that.

breedx-splk (approved):
Thanks!

PR #464, opened 2023-02-01
Extract the slow rendering detection instrumentation

mateuszrzeszutek:


laurit (approved):

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/slowrendering/SlowRenderListener.java
```diff
@@ -194,7 +189,7 @@ private void reportSlowRenders() {
         try {
             activities.forEach((activity, listener) -> reportSlow(listener));
         } catch (Exception e) {
-            Log.w(LOG_TAG, "Exception while processing frame metrics", e);
+            Log.w("OpenTelemetryRum", "Exception while processing frame metrics", e);
```
breedx-splk, line 192: we should still make a constant field for that `OpenTelemetryRum` even if it's not `LOG_TAG`. What happened to that anyway?

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -191,15 +190,19 @@ SplunkRum initialize(
...
+                        SlowRenderingDetector.builder()
+                                .setSlowRenderingDetectionPollInterval(
+                                        builder.slowRenderingDetectionPollInterval)
+                                .build()
+                                .installOn(instrumentedApplication);
+                        initializationEvents.add(
+                                new RumInitializer.InitializationEvent(
+                                        "slowRenderingDetectorInitialized", timingClock.now()));
+                    });
+        }
```
breedx-splk, line 205: If this was factored out to a named method like `initializeSlowRenderingDetection(otelRumBuilder)` or something, you could return early from the if and save some indent depth. It's getting a little burly in there.

breedx-splk (approved):
Looks good. 🤘🏻

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/slowrendering/SlowRenderListener.java
mateuszrzeszutek: `LOG_TAG` is placed in `SplunkRum`, can't use it here.

> we should still make a constant field for that `OpenTelemetryRum`

👍

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
mateuszrzeszutek: Done. I also extracted similar private methods for other instrumentations.

PR #476, opened 2023-02-15
Initial refactoring of Activity callbacks

breedx-splk:
Part of refactoring/cleanup effort before donation. It doesn't yet move `ActivityCallbacks` (and friends) into the otel package yet, because it's already getting quite large. It's expected that there will be at least one or two more PRs around this "Activity" code as it's moved into the otel package.

Here's a summary of what has changed:

* `ActivityCallbacks` was a rather complicated central branching out point for too many things lifecycle related. This aims to decompose it into several other several `ActivityLifecycleCallbacks` implementations, each with their own responsibility. 
* As a result, the number of helpers needed by `ActivityCallbacks` is greatly reduced. Its only real job now is to create events when Activity lifecycle stuff happens.
* All of these `ActivityLifecycleCallbacks` classes are registered into the app by the `RumInitializer`. There's more future related cleanup to do there.
* There was a similar activity caching logic in both `ActivityCallbacks` and `Pre29ActivityCallbacks`, so that was factored out to its own class: `ActivityTracerCache`.
* The `DefaultingActivityLifecycleCallbacks` interface was introduced to remove some of the empty implementation boilerplate caused by missing `default` methods. Is that API compatible? I hope so!
* Added some test coverage for new classes. Is it complete?

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ActivityTracerCache.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ActivityTracerCache.java
@@ -28,7 +28,7 @@ import java.util.function.Function;
  * Encapsulates the fact that we have an ActivityTracer instance per Activity class, and provides
  * convenience methods for adding events and starting spans.
  */
-public class ActivityTracerCache {
+class ActivityTracerCache {
 
     private final Map<String, ActivityTracer> tracersByActivityClassName = new HashMap<>();
 
@@ -62,20 +62,20 @@ public class ActivityTracerCache {
         this.tracerFactory = tracerFactory;
     }
 
-    public ActivityTracer addEvent(Activity activity, String eventName) {
+    ActivityTracer addEvent(Activity activity, String eventName) {
         return getTracer(activity).addEvent(eventName);
     }
 
-    public ActivityTracer startSpanIfNoneInProgress(Activity activity, String spanName) {
+    ActivityTracer startSpanIfNoneInProgress(Activity activity, String spanName) {
         return getTracer(activity).startSpanIfNoneInProgress(spanName);
     }
 
-    public ActivityTracer initiateRestartSpanIfNecessary(Activity activity) {
+    ActivityTracer initiateRestartSpanIfNecessary(Activity activity) {
         boolean isMultiActivityApp = tracersByActivityClassName.size() > 1;
         return getTracer(activity).initiateRestartSpanIfNecessary(isMultiActivityApp);
     }
 
-    public ActivityTracer startActivityCreation(Activity activity) {
+    ActivityTracer startActivityCreation(Activity activity) {
         return getTracer(activity).startActivityCreation();
     }
 
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ActivityTracerCache.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ActivityTracerCache.java
@@ -2,24 +2,37 @@ package com.splunk.rum;
 
 import android.app.Activity;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
+
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 
 import io.opentelemetry.api.trace.Tracer;
 
+/**
+ * Encapsulates the fact that we have an ActivityTracer instance per Activity class,
+ * and provides convenience methods for adding events and starting spans.
+ */
 public class ActivityTracerCache {
 
     private final Map<String, ActivityTracer> tracersByActivityClassName = new HashMap<>();
-    private final AtomicReference<String> initialAppActivity = new AtomicReference<>();
-    private final Tracer tracer;
-    private final VisibleScreenTracker visibleScreenTracker;
-    private final AppStartupTimer startupTimer;
+
+    private final Function<Activity,ActivityTracer> tracerFactory;
 
     public ActivityTracerCache(Tracer tracer, VisibleScreenTracker visibleScreenTracker, AppStartupTimer startupTimer) {
-        this.tracer = tracer;
-        this.visibleScreenTracker = visibleScreenTracker;
-        this.startupTimer = startupTimer;
+        this(tracer, visibleScreenTracker, new AtomicReference<>(), startupTimer);
+    }
+    @VisibleForTesting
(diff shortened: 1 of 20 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/ActivityTracerCache.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+
+/**
+ * Encapsulates the fact that we have an ActivityTracer instance per Activity class, and provides
+ * convenience methods for adding events and starting spans.
+ */
+public class ActivityTracerCache {
```
mateuszrzeszutek, line 31: Nit: probably shouldn't be public
```suggestion
class ActivityTracerCache {
```

splunk-otel-android/src/main/java/com/splunk/rum/VisibleScreenTracker.java
```diff
@@ -35,7 +35,7 @@
  * <p>We have to treat DialogFragments slightly differently since they don't replace the launching
  * screen, and the launching screen never leaves visibility.
  */
-class VisibleScreenTracker {
+public class VisibleScreenTracker {
```
mateuszrzeszutek, line 38:
```suggestion
class VisibleScreenTracker {
```

splunk-otel-android/src/test/java/io/opentelemetry/rum/internal/instrumentation/activity/RumFragmentActivityRegistererTest.java
```diff
@@ -0,0 +1,89 @@
...
+                RumFragmentActivityRegisterer.create(fragmentCallbacks);
+
+        underTest.onActivityPreCreated(activity, null);
+        verify(manager).registerFragmentLifecycleCallbacks(fragmentCallbacks, true);
+    }
+
+    @Test
+    void callbackIgnoresNonFragmentActivity() {
+        Activity activity = mock(Activity.class);
+        FragmentManager manager = mock(FragmentManager.class);
```
mateuszrzeszutek, line 55: The `manager` is not actually used anywhere in this test; you could just remove it and `verifyNoInteractions(manager)` would still be (implicitly) true

splunk-otel-android/src/test/java/com/splunk/rum/Pre29VisibleScreenLifecycleBindingTest.java
```diff
@@ -0,0 +1,49 @@
...
+        verifyNoMoreInteractions(tracker);
+    }
+
+    @Test
+    void prePaused() {
+        Pre29VisibleScreenLifecycleBinding underTest =
+                new Pre29VisibleScreenLifecycleBinding(tracker);
+        underTest.onActivityPaused(activity);
+        verify(tracker).activityPaused(activity);
+    }
```
mateuszrzeszutek, line 48: Nit: WDYT about adding `verifyNoMoreInteractions` to this method too?

splunk-otel-android/src/test/java/com/splunk/rum/ActivityTracerCacheTest.java
```diff
@@ -0,0 +1,138 @@
...
+        when(tracerCreator.apply(activity)).thenReturn(activityTracer);
+        when(activityTracer.addEvent(anyString())).thenReturn(activityTracer);
+
+        ActivityTracerCache underTest = new ActivityTracerCache(tracerCreator);
+        ActivityTracer result1 = underTest.addEvent(activity, "beep1");
+        ActivityTracer result2 = underTest.addEvent(activity, "beep2");
+        ActivityTracer result3 = underTest.addEvent(activity, "beep3");
+        assertSame(activityTracer, result1);
+        assertSame(activityTracer, result2);
+        assertSame(activityTracer, result3);
```
mateuszrzeszutek, line 72: This test will always pass even if you remove the cache, because of how the `tracerCreator` mock is set up. Consider doing 
```java
when(tracerCreator.apply(activity)).thenReturn(activityTracer, someOtherTracer);
```
and then verify that the other tracer was never returned

splunk-otel-android/src/test/java/com/splunk/rum/ActivityTracerCacheTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+        when(tracerCreator.apply(activity)).thenReturn(activityTracer);
+        when(tracerCreator.apply(activity2)).thenReturn(activityTracer2);
+        when(activityTracer.addEvent(anyString())).thenReturn(activityTracer);
+        when(activityTracer.initiateRestartSpanIfNecessary(true)).thenReturn(activityTracer);
+
+        ActivityTracerCache underTest = new ActivityTracerCache(tracerCreator);
+
+        underTest.addEvent(activity, "foo");
+        underTest.addEvent(activity2, "bar");
```
mateuszrzeszutek, line 121: Any verification for this?

mateuszrzeszutek (commented):
Nice 👍

splunk-otel-android/src/test/java/com/splunk/rum/ActivityTracerCacheTest.java
breedx-splk: I confirmed that it does still pass. :( I think the suggestion falls a little short -- the real intent is to confirm that multiple tracers weren't created by the factory, but that approach just makes sure that another one wasn't created. 

I think adding
`verify(tracerCreator).apply(activity);` at the end (with the implied `times(1)`) does what we want.

splunk-otel-android/src/test/java/com/splunk/rum/ActivityTracerCacheTest.java
breedx-splk: No, because the event stuff is tested in other methods. This is purely state/setup code required to get the cache set up with more than 1 activity. The test readily fails without either of these calls.

splunk-otel-android/src/main/java/com/splunk/rum/ActivityTracerCache.java
breedx-splk: not accepting suggestion because I also got some `public` methods. 👍🏻

mateuszrzeszutek (approved):

PR #478, opened 2023-02-17
Refactor ActivityCallbacks

breedx-splk:
This mostly moves around existing functionality to create 2 new packages in the otel internal instrumentation: `activity` and `fragment`. It also decouples the `AnchoredClock` from the `SplunkRumInitializer` and also brings over some constants into `RumConstants`.


splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
```diff
@@ -115,38 +119,96 @@ private Span createSpanWithParent(String spanName, @Nullable Span parentSpan) {
...
+        public Builder activeSpan(ActiveSpan activeSpan) {
+            this.activeSpan = activeSpan;
+            return this;
+        }
+
+        public String getActivityName() {
+            return activity.getClass().getSimpleName();
+        }
+
+        public RumScreenName getRumScreenName() {
```
mateuszrzeszutek, line 206:
```suggestion

        String getActivityName() {
            return activity.getClass().getSimpleName();
        }

        RumScreenName getRumScreenName() {
```

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
```diff
@@ -115,38 +119,96 @@ private Span createSpanWithParent(String spanName, @Nullable Span parentSpan) {
...
+        private AtomicReference<String> initialAppActivity = new AtomicReference<>();
+        private Tracer tracer;
+        private AppStartupTimer appStartupTimer;
+        private ActiveSpan activeSpan;
+
+        public Builder(Activity activity) {
+            this.activity = activity;
+        }
+
+        public Builder visibleScreenTracker(VisibleScreenTracker visibleScreenTracker) {
```
mateuszrzeszutek, line 172: We usually use the `set...` naming scheme in builders

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/RumConstants.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static io.opentelemetry.api.common.AttributeKey.stringKey;
+
+import io.opentelemetry.api.common.AttributeKey;
+
+public class RumConstants {
+
+    public static final String OTEL_RUM_LOG_TAG = "OpenTelemetryRum";
+
+    public static final AttributeKey<String> COMPONENT_KEY = AttributeKey.stringKey("component");
```
mateuszrzeszutek, line 27: Would it be possible to leave the `component` attribute in the "Splunk space"? Kinda like I did for crash reporting: `.addAttributesExtractor(constant(COMPONENT_KEY, COMPONENT_ERROR))`

I'm wondering if the OTel spec will include this attribute.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/Pre29ActivityCallbacks.java
```diff
@@ -14,18 +14,18 @@
...
+package io.opentelemetry.rum.internal.instrumentation.activity;
 
 import android.app.Activity;
 import android.os.Bundle;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import io.opentelemetry.rum.internal.DefaultingActivityLifecycleCallbacks;
 
-class Pre29ActivityCallbacks implements DefaultingActivityLifecycleCallbacks {
+public class Pre29ActivityCallbacks implements DefaultingActivityLifecycleCallbacks {
```
mateuszrzeszutek, line 25: WDYT about moving the SDK-version check logic in to the OTel package; hiding it behind a factory method, or an `installOn` method (kinda like in other instrumentations)
Perhaps we could keep some of these classes package-private then

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
breedx-splk: Eww. Ok, I'll use setters.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/RumConstants.java
breedx-splk: Oh yeah great point. I missed that API usage earlier....so this is a good idea.

As far as spec goes, that's hard to say. I feel like right now it might be kinda superfluous and can often be extrapolated from other things (span name and existence of other attributes)....but I also don't fully grok how our backend/ui might be using it.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/Pre29ActivityCallbacks.java
breedx-splk: Probably fine, can we do in a following PR?

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/RumConstants.java
breedx-splk: Ok, I managed to pull the `COMPONENT_KEY` back into the Splunk namespace. One key difference is that the activity/startup/fragment stuff uses the SDK more directly and does not go through the instrumentation APIs like the other code....

So rather than leveraging the `AttributesExtractor` like before, I wrapped the injected tracer. Please have a look at tell me what you think. If the only use case was to append a vendor-specific attribute (component), then passing an additional attribute or attributes seemed like overkill.

This exercise has me thinking more about the future in otel when many of these spans almost certainly turn into events. When that happens, I suspect that the existing instrumentation API won't serve us as well because it's pretty span-centric.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/Pre29ActivityCallbacks.java
breedx-splk: I have another branch that factors out the lifecycle instrumentation into the otel package in a way that's more consistent with the other `.installOn(app)` stuff. I think this is addressed there.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/RumConstants.java
mateuszrzeszutek: > This exercise has me thinking more about the future in otel when many of these spans almost certainly turn into events. When that happens, I suspect that the existing instrumentation API won't serve us as well because it's pretty span-centric.

Hmm, yeah, the existing instrumentation api assumes that things have a start and an end, and that they take time. I think we'll have to ensure that whatever OTel SIG comes up with is extensible enough to pass some extra attributes in.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
```diff
@@ -14,21 +14,27 @@
...
+package io.opentelemetry.rum.internal.instrumentation.activity;
+
+import static io.opentelemetry.rum.internal.RumConstants.APP_START_SPAN_NAME;
+import static io.opentelemetry.rum.internal.RumConstants.SCREEN_NAME_KEY;
+import static io.opentelemetry.rum.internal.RumConstants.START_TYPE_KEY;
 
 import android.app.Activity;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import com.splunk.rum.RumScreenName;
```
mateuszrzeszutek, line 26: Oh, I totally forgot about this annotation. I think we'll have to introduce some `(Fragment|Activity)NameExtractor` in a future PR/story that'll help us decouple this.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/ActiveSpan.java
```diff
@@ -14,30 +14,32 @@
...
+import static io.opentelemetry.rum.internal.RumConstants.LAST_SCREEN_NAME_KEY;
 
 import androidx.annotation.Nullable;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Scope;
 import java.util.function.Supplier;
 
-class ActiveSpan {
-    private final VisibleScreenTracker visibleScreenTracker;
+public class ActiveSpan {
```
mateuszrzeszutek, line 26: WDYT about moving this to `...util`?

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
breedx-splk: Ah sheesh, I forgot about it too. We'll have to do another sweep of the remaining `splunk` stuff in there anyway (there were others last I checked).

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/ActiveSpan.java
breedx-splk: oh sure. The name is more data class than utility, but it does wrap and assist so that seems fine.

PR #490, opened 2023-03-03
Move SpanFilterBuilder and collaborators to otel package

breedx-splk:
No real changes, just a lift-and-shift.

### DIFF
--- a/sample-app/build.gradle.kts
+++ b/sample-app/build.gradle.kts
@@ -51,7 +51,7 @@ android {
     }
 }
 
-val otelVersion = "1.23.1"
+val otelVersion = "1.23.1e"
 val otelAlphaVersion = "$otelVersion-alpha"
 val otelInstrumentationVersion = "1.23.0"
 val otelInstrumentationAlphaVersion = "$otelInstrumentationVersion-alpha"
--- a/sample-app/build.gradle.kts
+++ b/sample-app/build.gradle.kts
@@ -51,7 +51,7 @@ android {
     }
 }
 
-val otelVersion = "1.23.1e"
+val otelVersion = "1.23.1"
 val otelAlphaVersion = "$otelVersion-alpha"
 val otelInstrumentationVersion = "1.23.0"
 val otelInstrumentationAlphaVersion = "$otelInstrumentationVersion-alpha"
--- a/splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java
@@ -240,27 +240,13 @@ public final class SplunkRumBuilder {
     }
 
     /**
-     * Configures span data filtering. This method is deprecated in favor of {@link #configureSpanFilter(Consumer)}.
+     * Configures span data filtering.
      *
      * @param configurer A function that will configure the passed {@link SpanFilterBuilder}.
      * @return {@code this}
      */
-    @Deprecated
     public SplunkRumBuilder filterSpans(Consumer<SpanFilterBuilder> configurer) {
         configurer.accept(spanFilterBuilder);
-        return this;
-    }
-
-    /**
-     * Configures span data filtering.
-     *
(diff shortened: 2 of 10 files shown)


splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/SpanFilterBuilder.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.splunk.rum;
+package io.opentelemetry.rum.internal;
```
mateuszrzeszutek, line 17: The `SpanFilterBuilder` is part of the public API; we can't just move it to another package without introducing a breaking change.
WDYT about copying the existing `SpanFilterBuilder` over to the otel package and leaving a shim in the splunk package, that'd simply call the otel filter builder methods?

breedx-splk: Yeah good catch. I missed the fact that it is "api surface".

breedx-splk: Ok, I copied it back and turned it into a wrapper. Added method `configureSpanFilter` on `SplunkRumBuilder` and deprecated the other method that takes the splunk class. LMK what you think.

sample-app/build.gradle.kts
```diff
@@ -51,7 +51,7 @@ android {
     }
 }
 
-val otelVersion = "1.23.1"
+val otelVersion = "1.23.1e"
```
mateuszrzeszutek, line 54: e?
```suggestion
val otelVersion = "1.23.1"
```

splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java
```diff
@@ -240,16 +240,30 @@ public SplunkRumBuilder setDeploymentEnvironment(String environment) {
     }
 
     /**
-     * Configures span data filtering.
+     * Configures span data filtering. This method is deprecated in favor of configureSpanFilter().
```
mateuszrzeszutek, line 243: Nit
```suggestion
     * Configures span data filtering. This method is deprecated in favor of {@link #configureSpanFilter(Consumer)}.
```

splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java
```diff
@@ -240,16 +240,30 @@ public SplunkRumBuilder setDeploymentEnvironment(String environment) {
...
 
+    /**
+     * Configures span data filtering.
+     *
+     * @param configurer A function that will configure the passed {@link
+     *     io.opentelemetry.rum.internal.SpanFilterBuilder}.
+     * @return {@code this}
+     */
+    public SplunkRumBuilder configureSpanFilter(
+            Consumer<io.opentelemetry.rum.internal.SpanFilterBuilder> configurer) {
```
mateuszrzeszutek, line 262: In general I think deprecating the old method and adding a new is a great idea, but I think we should probably wait a bit before we actually do that - at least until we upstream the OTel parts. If we add this as it is right now, we'll have our public API depend on the `io.opentelemetry.rum.internal` package.

sample-app/build.gradle.kts
breedx-splk: Sorry, clearly a screw up.

splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java
breedx-splk: Yeah ok, fair enough...we don't want to have references to the internal package in our public API. I get it now. We're going to have to deal with the fact that EVERYTHING in otel is in internal at some point tho. 🤷🏻 

I removed the deprecation and removed the new configure method.

mateuszrzeszutek (approved):

PR #491, opened 2023-03-08
More RumInitializer refactoring

breedx-splk:
The main points:

* Split and flatten large anonymous `TracerProviderCustomizer` lambda in the `RumInitializer`. Instead, add as a series of sequential steps.
* Encapsulate the init events into a new `InitializationEvents` class.
* Factor out block of boolean config items from builder to own `ConfigFlags` class.
* Change `SessionIdRatioBasedSampler` to take a `Supplier<String>` for the session id (instead of `Supplier<SplunkRum>`

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ConfigFlags.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ConfigFlags.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.rum;
+
+public class ConfigFlags {
+    private boolean debugEnabled = false;
+    private boolean diskBufferingEnabled = false;
+    private boolean reactNativeSupportEnabled = false;
+    private boolean crashReportingEnabled = true;
+    private boolean networkMonitorEnabled = true;
+    private boolean anrDetectionEnabled = true;
+    private boolean slowRenderingDetectionEnabled = true;
+
+    public void enableDebug() {
+        debugEnabled = true;
+    }
+
+    public void enableDiskBuffering() {
+        diskBufferingEnabled = true;
+    }
+
+    public void enableReactNativeSupport() {
+        reactNativeSupportEnabled = true;
+    }
+
+    public void disableCrashReporting() {
+        crashReportingEnabled = false;
+    }
+
+    public void disableNetworkMonitor() {
+        networkMonitorEnabled = false;
+    }
+
+    public void disableAnrDetection() {
+        anrDetectionEnabled = false;
+    }
+
+    public void disableSlowRenderingDetection() {
+        slowRenderingDetectionEnabled = false;
+    }
+
+    public boolean isDebugEnabled() {
+        return debugEnabled;
+    }
+
+    public boolean isAnrDetectionEnabled() {
+        return anrDetectionEnabled;
+    }
+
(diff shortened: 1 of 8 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/InitializationEvents.java
```diff
@@ -0,0 +1,98 @@
...
+
+    InitializationEvents(AppStartupTimer startupTimer) {
+        this.startupTimer = startupTimer;
+    }
+
+    void emit(String eventName) {
+        events.add(new Event(eventName, startupTimer.clockNow()));
+    }
+
+    void recordInitializationSpans(long startTimeNanos, ConfigFlags flags, Tracer delegateTracer) {
```
mateuszrzeszutek, line 43: Can the `startTimeNanos` be encapsulated in this class? You could set it in the constructor

splunk-otel-android/src/main/java/com/splunk/rum/InitializationEvents.java
```diff
@@ -0,0 +1,98 @@
...
+                        delegateTracer
+                                .spanBuilder(spanName)
+                                .setAttribute(COMPONENT_KEY, COMPONENT_APPSTART);
+
+        Span overallAppStart = startupTimer.start(tracer);
+        Span span =
+                tracer.spanBuilder("SplunkRum.initialize")
+                        .setParent(Context.current().with(overallAppStart))
+                        .setStartTimestamp(startTimeNanos, TimeUnit.NANOSECONDS)
+                        .setAttribute(COMPONENT_KEY, COMPONENT_APPSTART)
```
mateuszrzeszutek, line 55: The `tracer` used here already does that
```suggestion
```

splunk-otel-android/src/main/java/com/splunk/rum/InitializationEvents.java
```diff
@@ -0,0 +1,98 @@
...
+            span.addEvent(initializationEvent.name, initializationEvent.time, TimeUnit.NANOSECONDS);
+        }
+        long spanEndTime = startupTimer.clockNow();
+        // we only want to create SplunkRum.initialize span when there is a AppStart span so we
+        // register a callback that is called right before AppStart span is ended
+        startupTimer.setCompletionCallback(() -> span.end(spanEndTime, TimeUnit.NANOSECONDS));
+    }
+
+    @NonNull
+    private String buildConfigSettings(ConfigFlags flags) {
```
mateuszrzeszutek, line 71: Nit: WDYT about putting all this into `ConfigFlags#toString()`?

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -105,81 +101,102 @@ SplunkRum initialize(
...
+        initializationEvents.emit("resourceInitialized");
 
         CurrentNetworkProvider currentNetworkProvider =
                 currentNetworkProviderFactory.apply(application);
-        initializationEvents.add(
-                new InitializationEvent("connectionUtilInitialized", startupTimer.clockNow()));
+        initializationEvents.emit("connectionUtilInitialized");
+
+        // TODO: How truly important is the order of these span processors? The location of event
+        // generation should probably not be altered...
```
mateuszrzeszutek, line 111: Hah, good question. Ideally we should have some sort of a contract that dictates what we can do with these (or, no contract meaning we can do whatever)

mateuszrzeszutek (commented):
Really nice!

splunk-otel-android/src/main/java/com/splunk/rum/ConfigFlags.java
```diff
@@ -0,0 +1,83 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.rum;
+
+public class ConfigFlags {
```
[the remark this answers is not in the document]
breedx-splk: Looks like it doesn't need to be, I'll make package.

splunk-otel-android/src/main/java/com/splunk/rum/InitializationEvents.java
breedx-splk: I like it!

mateuszrzeszutek (approved):

PR #501, opened 2023-03-22
Migrate sampler to otel package

breedx-splk:
So this really does two things.

1. It attempts to address the TODO in the `RumInitializer` about the lazy init of a static/mutable variable. It does this by leveraging the fact that the session ID is also in the attributes passed to the `shouldSample()` method. One downside is that we currently rely on "splunk" in that name and so there's still splunk leakage in the otel package (this just moves things around however). We will need to figure this out -- especially because the name the sampler uses will need to match the name the `SessionIdSpanAppender` uses.
2. It moves the sampler out of the splunk package and into the otel package.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
@@ -46,6 +46,7 @@ import io.opentelemetry.exporter.zipkin.ZipkinSpanExporter;
 import io.opentelemetry.rum.internal.GlobalAttributesSpanAppender;
 import io.opentelemetry.rum.internal.OpenTelemetryRum;
 import io.opentelemetry.rum.internal.OpenTelemetryRumBuilder;
+import io.opentelemetry.rum.internal.SessionIdRatioBasedSampler;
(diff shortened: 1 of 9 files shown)


mateuszrzeszutek:
> 1. It does this by leveraging the fact that the session ID is also in the attributes passed to the `shouldSample()` method.

Is it, really? Session id is appended in a `SpanProcessor`, which runs after the `Sampler`.

> 1. One downside is that we currently rely on "splunk" in that name and so there's still splunk leakage in the otel package (this just moves things around however). We will need to figure this out -- especially because the name the sampler uses will need to match the name the `SessionIdSpanAppender` uses.

I suppose we could use the `SplunkSpanDataModifier` to rename the attribute just before the span is exported.

breedx-splk:
> > 1. It does this by leveraging the fact that the session ID is also in the attributes passed to the `shouldSample()` method.
> 
> Is it, really? Session id is appended in a `SpanProcessor`, which runs after the `Sampler`.

Ugh, you're right. My testing was invalid because I hadn't actually set the config setting that enables the sampler (it's off by default). 

Time to regroup.

PR #503, opened 2023-03-23
Migrate sampler to otel package (take 2)

breedx-splk:
The `SessionId` is now created much earlier in the startup/init lifecycle. This allows us to pass the instance to the sampler and not rely on lazy `Supplier` action. The sampler is now moved into the otel package.

### DIFF
--- a/splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRumBuilder.java
+++ b/splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRumBuilder.java
@@ -41,6 +41,7 @@ import java.util.function.Consumer;
  */
 public final class OpenTelemetryRumBuilder {
 
+    private final SessionId sessionId;
     private Resource resource = Resource.getDefault();
     private final List<BiFunction<SdkTracerProviderBuilder, Application, SdkTracerProviderBuilder>>
             tracerProviderCustomizers = new ArrayList<>();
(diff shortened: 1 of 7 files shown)


splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRumBuilder.java
```diff
@@ -51,7 +52,9 @@ public final class OpenTelemetryRumBuilder {
     private final List<Consumer<InstrumentedApplication>> instrumentationInstallers =
             new ArrayList<>();
 
-    OpenTelemetryRumBuilder() {}
+    OpenTelemetryRumBuilder(SessionId sessionId) {
```
mateuszrzeszutek, line 55: Nit: why not just create `SessionId` and `SessionIdTimeoutHandler` here, instead of the `builder()` method?

splunk-otel-android/src/test/java/io/opentelemetry/rum/internal/SessionIdRatioBasedSamplerTest.java
```diff
@@ -29,17 +30,16 @@
...
 import java.util.List;
-import java.util.function.Supplier;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith(MockitoExtension.class)
 class SessionIdRatioBasedSamplerTest {
+    @Mock SessionId sessionId;
```
mateuszrzeszutek, line 40: Nit: consider moving it lower, to where instance fields are located

mateuszrzeszutek (approved):

PR #504, opened 2023-03-23
Move @RumScreenName to otel package.

breedx-splk:
The `ActivityTracer` had already been migrated, but had a dependency back to the splunk package. By moving the annotation into the otel package, we eliminate this cyclic/reverse dependency.


mateuszrzeszutek:
The annotation is part of our public API though -- if we simply move it to the otel package we'll introduce a breaking change for whoever's currently using it.

Instead, WDYT about adding an interface (not necessarily in this shape, that's just an idea)
```java
interface ScreenNameExtractor {

  @Nullable
  String extract(Activity a);

  @Nullable
  String extract(Fragment f);
}
```
that we could implement and pass the implementation to `AndroidLifecycleInstrumentation`?

breedx-splk:
> The annotation is part of our public API though -- if we simply move it to the otel package we'll introduce a breaking change for whoever's currently using it.
<snip>

Yeah ok, good point. I hate that we can't change things easily, but so it goes.....

I ended up restoring the one in our package and also made a new one. Made an interface as suggested and two impls -- one for vanilla otel that looks at the otel annotation and one for our distro that uses our annotation first and falls back to otel.

Lemme know what you think.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
```diff
@@ -196,12 +195,17 @@ private String getActivityName() {
             return activity.getClass().getSimpleName();
         }
 
-        private RumScreenName getRumScreenName() {
-            return activity.getClass().getAnnotation(RumScreenName.class);
-        }
+        //        private RumScreenName getRumScreenName() {
+        //            return activity.getClass().getAnnotation(RumScreenName.class);
+        //        }
```
mateuszrzeszutek, line 201: To be removed?
```suggestion
```

mateuszrzeszutek (approved):

PR #506, opened 2023-03-23
Rename splunk.rum.previous_session_id

breedx-splk:
This builds on #505 and will need a rebase. It's similar, but renames `splunk.rum.previous_session_id` to `rum.previous_session_id` and updates the `SplunkSpanDataModifier` to patch back in the old/existing name while leaving the new one in place.


splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/RumConstants.java
```diff
@@ -24,12 +24,17 @@ public class RumConstants {
 
     public static final String OTEL_RUM_LOG_TAG = "OpenTelemetryRum";
 
+    public static final AttributeKey<String> SESSION_ID_KEY = stringKey("rum.session.id");
```
mateuszrzeszutek, line 27: Should we rename it to `rum.session_id`? Or perhaps rename the other attribute to `rum.session.previous_id`?

breedx-splk: Oh, I like `rum.session.previous_id` a lot. Thanks!

mateuszrzeszutek (approved):

PR #515, opened 2023-04-04
Translate OTel logs/events to spans

mateuszrzeszutek:
Implements a fairly simple translation layer for OTel logs & events.

I thought of 2 ways to do this:
1) Implement a `LogRecordProcessor` that uses the Tracer API to start & end spans -- this option has the advantage of reusing all our `SpanProcessor` decoration logic
2) Implement a `LogRecordProcessor` that converts the logs directly to `SpanData` and places them on BSP -- this option will most likely have a much better performance.

I implemented option 1 in this PR. We might wanna switch to 2 at some point, perhaps after the upstream instrumentations switch to events (instead of spans).

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/LogToSpanBridge.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/LogToSpanBridge.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/LogToSpanBridge.java
```diff
@@ -0,0 +1,121 @@
...
+            builder.setSchemaUrl(schemaUrl);
+        }
+        return builder.build();
+    }
+
+    private static String getSpanName(LogRecordData log) {
+        String operationName = log.getAttributes().get(OPERATION_NAME);
+        if (operationName != null) {
+            return operationName;
+        }
```
mateuszrzeszutek, line 90: Spans have to have a name; log records don't have one. In future instrumentations that use logs/events we could use that attribute to inject the span name; e.g. perhaps by using an `AttributesExtractor`-like thing for customization.

breedx-splk (approved):
I think this is a good first step. Aside from the change in `SecondFragement`, nothing is yet generating logs. Is the idea to (somewhat soon?) migrate the existing instrumentation away from spans and to logs/events? Assuming yes, do you have a sense of how we might verify the bridge satisfies zipkin/ingest requirements? Thanks!

PR #524, opened 2023-04-12
Migrate the bulk of resource building to otel package.

breedx-splk:
A few things of note:

* This also renames `splunk.rum.version` to `rum.sdk.version`
* Add new class `AndroidResource` that will create the default resource for this platform
* New method `OpenTelemetryRumBuilder.mergeResource()` to allow user to augment instead of replace entirely
* The `RumInitializer` still contains splunk-specific (distro) resource requirements.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
@@ -204,7 +204,7 @@ class RumInitializer {
         // Lifecycle events instrumentation are always installed.
         installLifecycleInstrumentations(otelRumBuilder, visibleScreenTracker);
 
-        OpenTelemetryRum openTelemetryRum = otelRumBuilder.build(application);
+        OpenTelemetryRum openTelemetryRum = otelRumBuilder.build();
 
         initializationEvents.recordInitializationSpans(
                 builder.getConfigFlags(),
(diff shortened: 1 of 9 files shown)


breedx-splk:
Need to make some fixes.

splunk-otel-android/src/main/java/io/opentelemetry/rum/internal/AndroidResource.java
```diff
@@ -0,0 +1,84 @@
...
+                .put(OS_VERSION, Build.VERSION.RELEASE)
+                .build();
+    }
+
+    private static String readAppName(Application application) {
+        return trapTo(
+                () -> {
+                    int stringId =
+                            application.getApplicationContext().getApplicationInfo().labelRes;
+                    return application.getApplicationContext().getString(stringId);
```
mateuszrzeszutek, line 59: Nice

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -248,30 +241,20 @@ private void installLifecycleInstrumentations(
...
 
-    private Resource createResource() {
+    private Resource createSplunkResource() {
         // applicationName can't be null at this stage
         String applicationName = requireNonNull(builder.applicationName);
         ResourceBuilder resourceBuilder =
-                Resource.getDefault().toBuilder()
-                        .put(APP_NAME_KEY, applicationName)
-                        .put(SERVICE_NAME, applicationName);
+                Resource.getDefault().toBuilder().put(APP_NAME_KEY, applicationName);
```
mateuszrzeszutek, line 248: Should we override the default `SERVICE_NAME` with `applicationName`? To keep them consistent

As a separate story/improvement, I guess we could make it optional and allow not specifying it in the `SplunkRumBuilder`

mateuszrzeszutek (approved):

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
breedx-splk: Yeah, but I didn't want to keep that in the splunk side of things -- the setting of `SERVICE_NAME` with the app name is happening over in the `AndroidResource`. I thought it was important enough to elevate.

PR #526, opened 2023-04-12
Split splunk/otel into separate modules

breedx-splk:
This split should help to force the separation and make the seam a little more visible. Should also make peeling part off simpler.


opentelemetry-android-instrumentation/build.gradle.kts
```diff
@@ -0,0 +1,86 @@
+plugins {
+    id("com.android.library")
+    id("splunk.android-library-conventions")
```
mateuszrzeszutek, line 3: This is going to be published as `com.splunk:opentelemetry-android-instrumentation` version 1.x -- while the `internal` package is a really good idea, should we perhaps do anything more to avoid user confusion? E.g. adding `-internal` to the project name, or using `-alpha` version

mateuszrzeszutek (approved):

opentelemetry-android-instrumentation/build.gradle.kts
breedx-splk: Can we just prevent it from publishing at all? We don't need it to, right?

mateuszrzeszutek: We have to, unfortunately -- since the main `splunk-otel-android` depends on that. (Unless we shade this module into `splunk-otel-android`)

breedx-splk: Oh yeah, I had expected to shade and probably didn't include that huh. I should work on that.

mateuszrzeszutek: > Oh yeah, I had expected to shade and probably didn't include that huh. I should work on that.

👍 
Do you wanna do that in a separate PR perhaps? I'm guessing it'll involve some gradle magic, and this PR is already large enough.

mateuszrzeszutek (approved):

breedx-splk:
Yeah, separate would be good. As long as we don't release that should be fine.

PR #587, opened 2023-06-30
Allow building RUM with preconfigured SDK.

breedx-splk:
As part of [#1400](https://github.com/open-telemetry/community/issues/1400) there was [this conversation](https://github.com/open-telemetry/community/issues/1400#issuecomment-1494825874) about "bring your own" sdk. While this has not been historically important with one vendor (Splunk), this will need to be relaxed when going into upstream. 

So this provides a peer of `OpenTelemetryRumBuilder` called `SdkPreconfiguredRumBuilder`. This class does not provide a means to customize any of the various `*ProviderCustomizer` and assumes the user has already done all the hard work. For a single-vendor solution this is a bit of a foot gun, but we expect that opinionated distros can help fill in the important parts for their users.

### DIFF
--- a/opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRum.java
+++ b/opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRum.java
@@ -19,6 +19,9 @@ package io.opentelemetry.rum.internal;
 import android.app.Application;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.logs.SdkLoggerProvider;
+import io.opentelemetry.sdk.metrics.SdkMeterProvider;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
 
 /**
  * Entrypoint for the OpenTelemetry Real User Monitoring library for Android.
--- a/opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRum.java
+++ b/opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRum.java
@@ -29,10 +29,10 @@ import io.opentelemetry.sdk.OpenTelemetrySdk;
 public interface OpenTelemetryRum {
 
     /**
-     * Returns a new {@link OpenTelemetryRumBuilder} for {@link OpenTelemetryRum}.
-     * Use this version if you would like to configure individual aspects of the OpenTelemetry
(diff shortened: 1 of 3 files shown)


mateuszrzeszutek:
Looks like you need to run spotless

opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/SdkPreconfiguredRumBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.rum.internal.instrumentation.InstrumentedApplication;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+
+public class SdkPreconfiguredRumBuilder {
```
mateuszrzeszutek, line 13:
```suggestion
public final class SdkPreconfiguredRumBuilder {
```

opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRum.java
```diff
@@ -29,13 +30,33 @@ public interface OpenTelemetryRum {
...
      */
     static OpenTelemetryRumBuilder builder(Application application) {
         return new OpenTelemetryRumBuilder(application);
     }
 
+    /**
+     * Returns a new {@link SdkPreconfiguredRumBuilder} for {@link OpenTelemetryRum}. This version
+     * requires the user to preconfigure and create their own OpenTelemetrySdk instance. If you
+     * prefer to use the builder to configure individual aspects of the OpenTelemetry SDK and to
+     * create and manage it for you, call the two-argument version.
```
mateuszrzeszutek, line 47:
```suggestion
     * create and manage it for you, call the one-argument version.
```

opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/OpenTelemetryRum.java
```diff
@@ -29,13 +30,33 @@ public interface OpenTelemetryRum {
...
+     *
+     * Specific consideration should be given to the creation of your provided SDK to ensure
+     * that the SdkTracerProvider, SdkMeterProvider, and SdkLoggerProvider are configured
```
mateuszrzeszutek, line 50:
```suggestion
     * that the {@link SdkTracerProvider}, {@link SdkMeterProvider}, and {@link SdkLoggerProvider} are configured
```

mateuszrzeszutek (approved):
LGTM 👍

PR #588, opened 2023-07-03
Remove Guava dep

mateuszrzeszutek:
Supersedes #586

Turns out it was only used in one place, and just one trivial method. Let's just remove it instead of solving hellish dependency resolution issues.


laurit (approved):

breedx-splk:
Hah, nice, and it was a test no less! Thanks!

PR #591, opened 2023-07-11
MOB-3245: Deprecating RumScreenName

dvernon-splunk:
Adds a deprecation tag to RumScreenName.java located in the com.splunk.rum package, and directs users to use RumScreenName.java in the opentelemetry package.

Updates usages of RumScreenName in tests and the sample app to use this new filepath.


splunk-otel-android/src/main/java/com/splunk/rum/RumScreenName.java
```diff
@@ -23,7 +23,9 @@
 
 /**
  * This annotation can be used to customize the {@code screen.name} attribute for an instrumented
- * Fragment or Activity. TODO: Deprecate this so the otel one can be used instead
+ * Fragment or Activity.
+ *
+ * @deprecated RumScreenName moved to io.opentelemetry.rum.internal.instrumentation package
```
breedx-splk, line 28: nit: Startcap to `@Deprecated`

splunk-otel-android/src/main/java/com/splunk/rum/SplunkScreenNameExtractor.java
```diff
@@ -19,6 +19,7 @@
 import android.app.Activity;
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
+import io.opentelemetry.rum.internal.instrumentation.RumScreenName;
```
breedx-splk, line 22: Well, it's a bummer that a test didn't catch this -- but I believe that this breaks existing functionality for users using the splunk annotation. I think we need to take this back out in the short term, so that the fallback mechanism can still operate.

splunk-otel-android/src/main/java/com/splunk/rum/RumScreenName.java
dvernon-splunk: Changing to uppercase D breaks the tag on my local environment, even though the official android developer reference shows it uppercase. Do you know why that might be?

laurit: @dvernon-splunk  `@Deprecated` is an annotation, what you are using is a javadoc tag which is supposed to be lowercase

dvernon-splunk: Thank you @laurit!

breedx-splk (approved):

PR #599, opened 2023-07-25
Fixed Rum version being appended twice

surbhimaheshwari:
The rum version was added from upstream as well as from SplunkResource. Removed one of them to get rid of duplicate attribute being appended.


breedx-splk (approved):
Thanks @surbhimaheshwari. We will need to keep an eye on changes in upstream to prevent this changing out from under us in the short term. If/when that changes, we will likely need to add something like this back in.

PR #605, opened 2023-08-10
Dependency and gradle updates

breedx-splk:
This updates to android build tools to 33.0.1 and update gradle wrapper to 8.2.1. 
Also updates the volley getter to use the new apis and removes the now unused network attr getter.


splunk-otel-android-volley/src/main/java/com/splunk/rum/VolleyHttpClientAttributesGetter.java
```diff
@@ -116,4 +127,9 @@ static List<String> headersToList(List<Header> headers, String name) {
         }
         return headersList;
     }
+
+    @Override
+    public String getTransport(RequestWrapper requestWrapper, HttpResponse httpResponse) {
+        return SemanticAttributes.NetTransportValues.IP_TCP;
+    }
```
mateuszrzeszutek, line 134: Can you also add a `getNetworkTransport()` method? The `getTransport()` one will be removed in 2.0
```suggestion
    }
    
    @Override
    public String getNetworkTransport(RequestWrapper requestWrapper, HttpResponse httpResponse) {
        return "tcp";
    }
```

mateuszrzeszutek (approved):

splunk-otel-android-volley/src/main/java/com/splunk/rum/VolleyHttpClientAttributesGetter.java
breedx-splk: Cool sorry I missed that one! Thanks!

PR #607, opened 2023-08-10
Improve how the SpanFilterBuilder delegates to the SpanDataModifier.

### LINKED ISSUE #585: SpanFilterBuilder - Explore delegating instead of storing
As part of #576, this comment came up:

https://github.com/signalfx/splunk-otel-android/pull/576#discussion_r1228213062

I think the idea is to create the delegated builder early and then delegate to the builder, instead of storing each filter item individually.

breedx-splk:
Resolves #585. 

Thanks for the idea @mateuszrzeszutek it cleaned up quite well. 🥳


breedx-splk:
Build breakage is fixed in #605. Will rebase after that's merged.

mateuszrzeszutek (approved):
Nice 👍

Issue #606, opened 2023-08-10
Prepare for session id name change


breedx-splk:
It's likely that https://github.com/open-telemetry/semantic-conventions/pull/215 will merge before too long, and we should prepare to handle data with the newer convention. 

Recommend doing the following:
* Rename `rum.session.id` to `session.id`. This allows us to use the new name internally everywhere
* Leverage the `SessionIdSpanAppender` to copy the `session.id` back into the legacy `rum.session.id` name. We should then send both attributes. This allows us to remain backwards compatible while also allowing the backend to eventually begin using the new name.

breedx-splk:
We are using `session.id` from upstream RumConstants and turning it back into `splunk.rumSessionId` in the `SplunkSpanDataModifier`.

PR #610, opened 2023-08-14
Added session.id as key along with rum.session.id for span

surbhimaheshwari:
As per semantic convention added the replaced the `rum.session.id` to `session.id`. Adding both of them to the attribute in order to make this backward compatible.

Resolves https://github.com/signalfx/splunk-otel-android/issues/606


breedx-splk (commented):
Oof, I had forgotten that the `SessionIdSpanAppender` is in the upstream package and we will eventually be sourcing it from there. This means that we really shouldn't be making changes in the `io.opentelemetry.*` packages in this project (they will go away as soon as we can source from an upstream artifact). Sorry for the misleading guidance in the issue.

I think we will need another component that copies from the upstream field name (likely `session.id` into our rum name). This will need to also be compatible with the `LogToSpanBridge` sometime soon.

surbhimaheshwari:
Closing it for now until the project is changed to use upstream project `opentelemetry-android`

PR #614, opened 2023-08-21
Disable Background Task Instrumentation

surbhimaheshwari:
Provides user with an option to disable background process instrumentation using `disableBackgroundTaskReporting`. If the app is launched using background process, we initialize SplunkRum with No-op instance and thus stop the user from reporting any instrumetnation

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/BackgroundProcessDetector.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/BackgroundProcessDetector.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.splunk.rum;
 
 import android.annotation.SuppressLint;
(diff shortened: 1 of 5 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/BackgroundProcessDetector.java
```diff
@@ -0,0 +1,42 @@
...
+
+import android.annotation.SuppressLint;
+import android.app.Application;
+import android.os.Build;
+import java.lang.reflect.Method;
+
+public class BackgroundProcessDetector {
+    public static Boolean isBackgroundProcess(String applicationId) {
+        String applicationProcessName = "";
+        if (Build.VERSION.SDK_INT >= 28) applicationProcessName = Application.getProcessName();
```
breedx-splk, line 27: I'm surprised that this passed spotless checks. We want all `if` statements to be multi-line and have curly brace blocks around them.

splunk-otel-android/src/main/java/com/splunk/rum/BackgroundProcessDetector.java
```diff
@@ -0,0 +1,42 @@
...
+                Class<?> activityThread = Class.forName("android.app.ActivityThread");
+                String methodName = "currentProcessName";
+                @SuppressLint("PrivateApi")
+                Method getProcessName = activityThread.getDeclaredMethod(methodName);
+                applicationProcessName = (String) getProcessName.invoke(null);
+            } catch (Exception e) {
+            }
+        }
+
+        return applicationProcessName != null && applicationProcessName.equals(applicationId);
```
breedx-splk, line 40: This will read a little cleaner if you factor out the above to something like `String applicationProcessName = getApplicationProcessName()` and then in that factored out method you can return early and simplify the if/else.

breedx-splk (changes requested):
Looks good, really just a couple of small suggestions. Thanks!

splunk-otel-android/src/main/java/com/splunk/rum/BackgroundProcessDetector.java
```diff
@@ -23,6 +23,9 @@
 import java.util.Objects;
 
 public final class BackgroundProcessDetector {
+
+    BackgroundProcessDetector() {}
```
breedx-splk, line 27:
```suggestion
    private BackgroundProcessDetector() {}
```

breedx-splk (approved):
Thanks!

PR #623, opened 2023-08-25
Fix up errorprone and warnings that fail the check build

breedx-splk:
This is related to upstream https://github.com/open-telemetry/opentelemetry-android/pull/46 and https://github.com/open-telemetry/opentelemetry-android/pull/48.

We had a number of issues that were not caught when errorprone started not working on android builds. This should help with those.

### DIFF
--- a/opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
+++ b/opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
@@ -150,12 +150,15 @@ public class ActivityTracer {
     }
 
     static class Builder {
+        private static final ActiveSpan INVALID_ACTIVE_SPAN = new ActiveSpan(() -> null);
+        private static final Tracer INVALID_TRACER = spanName -> null;
+        private static final AppStartupTimer INVALID_TIMER = new AppStartupTimer();
         private final Activity activity;
-        public String screenName;
+        public String screenName = "unknown_screen";
         private AtomicReference<String> initialAppActivity = new AtomicReference<>();
-        private Tracer tracer;
-        private AppStartupTimer appStartupTimer;
-        private ActiveSpan activeSpan;
+        private Tracer tracer = INVALID_TRACER;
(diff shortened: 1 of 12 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/LogToSpanBridge.java
```diff
@@ -34,6 +34,7 @@
 import io.opentelemetry.sdk.logs.data.LogRecordData;
 import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
 import java.util.concurrent.TimeUnit;
+import org.jetbrains.annotations.Nullable;
```
laurit, line 37: VolleyHttpClientAttributesGetter uses `androidx.annotation.Nullable`, elsewhere we use `javax.annotation.Nullable`

opentelemetry-android-instrumentation/src/main/java/io/opentelemetry/rum/internal/instrumentation/activity/ActivityTracer.java
```diff
@@ -195,13 +198,22 @@ private String getActivityName() {
...
-
         public Builder setScreenName(String screenName) {
             this.screenName = screenName;
             return this;
         }
+
+        public ActivityTracer build() {
+            if (activeSpan == INVALID_ACTIVE_SPAN) {
+                throw new IllegalStateException("activeSpan must be configured.");
+            }
```
mateuszrzeszutek, line 209: Why not `if (activeSpan == null)`?
I'm guessing that'd require adding some `@Nullable` and `requireNonNull()`, but isn't that less bad than creating fake invalid instances/

breedx-splk: Let's see what it looks like....

mateuszrzeszutek: Bleh... welp, too bad, at least we tried.

mateuszrzeszutek (approved):

PR #624, opened 2023-08-29
Fix the background process detector api

surbhimaheshwari:


splunk-otel-android/src/main/java/com/splunk/rum/BackgroundProcessDetector.java
```diff
@@ -28,7 +28,7 @@ private BackgroundProcessDetector() {}
 
     public static Boolean isBackgroundProcess(String applicationId) {
         String applicationProcessName = getApplicationProcessName();
-        return Objects.equals(applicationProcessName, applicationId);
+        return !Objects.equals(applicationProcessName, applicationId);
```
laurit, line 31: it would be nice to have a comment describing why this works for detecting background processes.

PR #629, opened 2023-08-29
Remove module and depend on upstream

breedx-splk:
Draft for now because upstream might do some repackaging or coordinates tweaking...but it's progress!


breedx-splk:
Closing in favor of #640.

PR #632, opened 2023-08-30
Fix build and prepare for 1.1.1 branch.

breedx-splk:
This simplifies the build a bit and actually produces the artifacts we need.

The `skipSigning` parameter is used when running locally with `publishToMavenLocal` when the signing secrets are not available.

### DIFF
--- a/buildSrc/src/main/kotlin/splunk.android-library-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/splunk.android-library-conventions.gradle.kts
@@ -50,6 +50,17 @@ if (project.findProperty("release") == "true" && project.findProperty("skipSigni
 }
 
 project.afterEvaluate {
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/splunk.android-library-conventions.gradle.kts
```diff
@@ -28,8 +36,34 @@ publishing {
             }
         }
     }
-    publications {
-        register<MavenPublication>("maven") {
+}
+
+
+if (project.findProperty("release") == "true" && project.findProperty("skipSigning") != "true") {
```
mateuszrzeszutek, line 42: Could be
```suggestion
if (isARelease && project.findProperty("skipSigning") != "true") {
```
?

mateuszrzeszutek (approved):

PR #640, opened 2023-09-13
Remove module and depend on upstream.

breedx-splk:
Supersedes #629.

This is the first change that will remove the otel classes which now exist in upstream. We source it as a dependency.


sample-app/build.gradle.kts
```diff
@@ -57,7 +57,7 @@ val otelAlphaVersion = otelVersion.replaceFirst("(-SNAPSHOT)?$".toRegex(), "-alp
 dependencies {
     api(platform("io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:$otelAlphaVersion"))
 
-    implementation(project(":opentelemetry-android-instrumentation"))
+    implementation("io.opentelemetry.android:instrumentation:0.1.0-alpha-SNAPSHOT")
```
mateuszrzeszutek, line 60:
```suggestion
    implementation("io.opentelemetry.android:instrumentation:0.1.0-alpha")
```

splunk-otel-android/build.gradle.kts
```diff
@@ -44,7 +44,8 @@ val otelAlphaVersion = otelVersion.replaceFirst("(-SNAPSHOT)?$".toRegex(), "-alp
 
 dependencies {
     api(platform("io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:$otelAlphaVersion"))
-    implementation(project(":opentelemetry-android-instrumentation"))
+    implementation("io.opentelemetry.android:instrumentation:0.1.0-alpha")
```
mateuszrzeszutek, line 47: I wonder, should we perhaps expose it as an `api` dep?

mateuszrzeszutek (approved):

PR #648, opened 2023-09-26
Feature/background aware span export

rezastallone:
## Description
Enable configuration to opt out spans created when app in background to be sent to Splunk server, however if the app is brought to foreground in the same session, the span created previously in the background will be sent to Splunk server. If the process in the background gets killed before brought to foreground, any span recorded will never be sent and will be deleted in house keeping process by DeviceSpanStorageLimiter.

## Changes

- MemoryBufferingExporter doesn't send background span if  disableBackgroundTaskReporting param is true
- ZipkinToDiskSender doesn't send background span if  disableBackgroundTaskReporting param is true
- Store background span to /spans/background/<uuid>/ will stay there forever if session never go foreground
- Store in-memory backlog to background backlog, will not be sent if session never go foreground

## How Has This Been Tested?

Create background worker that run periodically every 15 minute, you can fast-forward 15 minute through phone setting to reproduce easily.

**[Case 1] App start in background, stays in background until process killed**
Session ID created but no span should be sent
Result :

[backgroundnosendsplunksample.txt](https://github.com/signalfx/splunk-otel-android/files/12732155/backgroundnosendsplunksample.txt)

As you can see no span uploaded successfully since process never go foreground, if i look up session ce98849ce929eb473b1f4b9c8855dd0f, b843461af379816ef295a73d21d7e610, b4def7e60cb3e1e9ffc134a09d5585a3 it is not found.
Screenshot 2023-09-27 at 04 45 00

Screenshot 2023-09-27 at 04 44 36

Screenshot 2023-09-27 at 04 45 28

**[Case 2] App start in background, process not killed yet but brought to foreground**

Session ID created, span stored in spans/background/<uuid> moved to spans/ for eventual send

[eventualsendfrombackgroundsplunksample.txt](https://github.com/signalfx/splunk-otel-android/files/12732169/eventualsendfrombackgroundsplunksample.txt)

Session ID 3a2acb8c338202ea689b5821b54c9650 timeline : 

Screenshot 2023-09-27 at 04 48 14

[CLA.md](https://github.com/signalfx/splunk-otel-android/files/12920051/CLA.md)


jek-bao-choo:
Thank you @rezastallone 

+ @breedx-splk @mateuszrzeszutek

breedx-splk (commented):
Hey thanks @rezastallone. I've looked at most of this, seems reasonable so far. I had some comments so let's work through those and also get the build passing. Thanks!

rezastallone:
Thanks for the review @breedx-splk, i'll work on it on 29th September Jakarta time as we're currently on National Holiday.
Again thank you for the feedback.

splunk-otel-android/src/main/java/com/splunk/rum/StartTypeAwareBacklogProvider.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public List<SpanData> fillFromBacklog() {
+        List<SpanData> retries = new ArrayList<>(backlog);
+        backlog.clear();
+
+        if (visibleScreenTracker.getPreviouslyVisibleScreen() != null){
+            retries.addAll(backgroundSpanBacklog);
+            backgroundSpanBacklog.clear();
```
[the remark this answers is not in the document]
rezastallone: Yes correct because app started in foreground can reuse the process started in the background if it's not killed yet. this way the foreground app wont missing any information regarding the Application class onCreate (or other process creation information) started by the background process beforehand.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -217,6 +217,23 @@ SplunkRum initialize(
         return new SplunkRum(openTelemetryRum, globalAttributesSpanAppender);
     }
 
+    @NonNull
+    private BacklogProvider constructBacklogProvider(VisibleScreenTracker visibleScreenTracker) {
+        if (builder.isBackgroundTaskReportingDisabled()){
```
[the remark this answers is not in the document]
rezastallone: I agree, i'll work on it

rezastallone:
@breedx-splk seems unable to resolve 0.2.0-alpha-SNAPSHOTS so i downgrade to 0.1.0 & applied spotless. tested on local build ran succesfully

breedx-splk:
> @breedx-splk seems unable to resolve 0.2.0-alpha-SNAPSHOTS so i downgrade to 0.1.0 & applied spotless. tested on local build ran succesfully

Sorry about that, there was a problem in upstream that caused the 0.2.0 snapshots to be broken yesterday. I believe it is fixed now, so please let's try again with 0.2.0. 🙏🏻  Thank you.

rezastallone:
> > @breedx-splk seems unable to resolve 0.2.0-alpha-SNAPSHOTS so i downgrade to 0.1.0 & applied spotless. tested on local build ran succesfully
> 
> Sorry about that, there was a problem in upstream that caused the 0.2.0 snapshots to be broken yesterday. I believe it is fixed now, so please let's try again with 0.2.0. 🙏🏻 Thank you.

Okay, I just reverted it

ardiprakasa:
Hi @breedx-splk, thanks for the review of the PR. I am Hardian from DANA as well. Do you think we can safely use this branch in our app now? Or should we wait for further assessment from the Splunk team and wait until there is a new SDK release that includes this PR?

Thank you.

breedx-splk:
Hey @rezastallone and @ardiprakasa -- thanks again for the PR. We're still working on getting it fully reviewed and the functionality verified. We will likely have a few more suggestions or questions. In the meantime, any additional testing/verification you can do with this branch would be helpful. 

After the PR is approved but before we can merge, we will also need to confirm that you have signed the [CLA](https://www.splunk.com/en_us/form/contributions.html) as per our [CONTRIBUTING.md](https://github.com/signalfx/splunk-otel-android/blob/main/CONTRIBUTING.md#contributor-license-agreement).

breedx-splk:
So I've managed to run the sample app with this branch and I'm surprised/concerned by a couple of behaviors. 

First, I changed the periodic worker to run every minute. The work request is created in the MainActivity `onCreate` and I would expect to see it logging every minute. It doesn't, it only logs once. Any ideas why?

Next, when the app is launched there is a file created in `spans/background/{id}/{other-id}.spans`, even though the application is running and foregrounded. The file and directory seem to stay remain indefinitely, at least while the app remains foregrounded. When the app is backgrounded, the file gets harvested (moved to parent directory and then processed), but the directory it was in. 

Over time, the background directory looks to be filling up with subdirectories.

> [Case 1] App start in background, stays in background until process killed

Can you describe to me how you are launching the sample app in the background?

rezastallone:
> So I've managed to run the sample app with this branch and I'm surprised/concerned by a couple of behaviors.
> 
> First, I changed the periodic worker to run every minute. The work request is created in the MainActivity `onCreate` and I would expect to see it logging every minute. It doesn't, it only logs once. Any ideas why?
> 
> Next, when the app is launched there is a file created in `spans/background/{id}/{other-id}.spans`, even though the application is running and foregrounded. The file and directory seem to stay remain indefinitely, at least while the app remains foregrounded. When the app is backgrounded, the file gets harvested (moved to parent directory and then processed), but the directory it was in.
> 
> Over time, the background directory looks to be filling up with subdirectories.
> 
> > [Case 1] App start in background, stays in background until process killed
> 
> Can you describe to me how you are launching the sample app in the background?

The minimum period for periodic work allowed is 15 minute anything lower than that wont trigger the job https://developer.android.com/guide/background/persistent/getting-started/define-work , if you want to trigger the job without have to wait 15 minute you can advance time with device setting.

May i know what device are you running on ?
Did the directories in spans/background/ created at the same time you launched the app in foreground ? or is it from previous launches ? If the background directories is from previous launches it will not be moved to the parent directory & will not be sent for eventual sending because it has different id with the current id stored in StartTypeAwareSpanStorage & it will remain in background indefinitely until it gets clean up by DeviceSpanStorageLimiter.
Every foreground may initially create spans in background directory because the information of visible activity may be delayed or not available yet especially when the app has not opened the Launcher Activity.

To run the app in background, first make sure the minimum period is at least 15 minute no smaller than that, after that I open the app and make sure the code for scheduling the period work at least executed once, then I kill the app finally I advance device clock by 15 minute or more through date & time setting, after that the background work should be executed & log start showing up.

rezastallone:
@breedx-splk Let me know if you have further questions

breedx-splk:
> The minimum period for periodic work allowed is 15 minute anything lower than that wont trigger the job

Ah ok, I hadn't realized that limitation of the platform. Thank you.

> May i know what device are you running on ?

Emulating a Pixel 7 api 34.

> Did the directories in spans/background/ created at the same time you launched the app in foreground ?

It happens immediately when the app is launched into the foreground. It happens both when opening through the desktop app icon on device and also when launching through Android Studio. This still seems like a problem to be resolved -- the startup should not be considered background.

> it will remain in background indefinitely until it gets clean up by DeviceSpanStorageLimiter.

This seems a little wasteful and problematic. I wonder if we can clean those up in a more consistent way before triggering the limiter.

Thanks for the clarifications about reproducing. I will try again and either wait 15 minutes or advance the clock. While looking at this again, one interesting aspect that occurs to me is that the session id is designed to time out after 15 minutes of background activity.

breedx-splk:
Ok, I was able to wait and see the background worker doing its thing and creating telemetry in the `background/{id}/{id2}.spans` files. I launched the app into the foreground by hand and the telemetry remained in the background directory until I used the app enough to create additional, non-backgrounded telemetry.

This is not what I expected from the original description. I expected that the background telemetry buffered in device storage would be sent to ingest shortly after the app was relaunched.

rezastallone:
Let me check with api 34, i have checked on that api.

splunk-otel-android/src/test/java/com/splunk/rum/StartTypeAwareSpanStorageTest.java
```diff
@@ -51,14 +51,15 @@ void setup() {
...
         List<File> spans = fileProvider.getPendingFiles().collect(Collectors.toList());
         assertEquals(0, spans.size());
     }
 
     @Test
-    void
-            getPendingFiles_givenPrevouslyInBackground_shouldMoveBackgroundSpanToForegroundSpanForSending() {
+    void getPendingFiles_givenPrevouslyInBackground_shouldMoveBackgroundSpanToForegroundSpanForSending() {
         when(visibleScreenTracker.getPreviouslyVisibleScreen()).thenReturn("MainActivity");
+        when(visibleScreenTracker.getCurrentlyVisibleScreen()).thenReturn("MainActivity");
```
breedx-splk: Might be a little weird that both the previously visible screen AND the currently visible screen have the same name. Can we make the previous one something different?

splunk-otel-android/src/test/java/com/splunk/rum/StartTypeAwareSpanStorageTest.java
```diff
@@ -95,6 +97,7 @@ void getSpanPath_givenInBackground_shouldReturnBackgroundSpanPath() {
     @Test
     void getSpanPath_givenInForeground_shouldReturnForegroundSpanPath() {
         when(visibleScreenTracker.getPreviouslyVisibleScreen()).thenReturn("MainActivity");
+        when(visibleScreenTracker.getCurrentlyVisibleScreen()).thenReturn("MainActivity");
```
breedx-splk: Same comment as above for the previous/current being the same name.

splunk-otel-android/src/main/java/com/splunk/rum/StartTypeAwareSpanStorage.java
```diff
@@ -0,0 +1,136 @@
...
+                        file -> {
+                            String path = file.getPath();
+                            String pathId = path.substring(path.lastIndexOf("/") + 1);
+                            return !pathId.equals(uniqueId);
+                        })
+                .forEach(
+                        file -> {
+                            Log.d(SplunkRum.LOG_TAG, "Cleaning up " + file.getPath());
+                            fileUtils.listFilesRecursively(file).forEach(fileUtils::safeDelete);
+                            fileUtils.safeDelete(file);
```
breedx-splk, line 133:
```suggestion
                        dir -> {
                            Log.d(SplunkRum.LOG_TAG, "Cleaning up " + dir.getPath());
                            fileUtils.listFilesRecursively(dir).forEach(fileUtils::safeDelete);
                            fileUtils.safeDelete(dir);
```

splunk-otel-android/src/main/java/com/splunk/rum/StartTypeAwareSpanStorage.java
```diff
@@ -0,0 +1,136 @@
...
+        if (pathToReturn.exists() || pathToReturn.mkdirs()) {
+            return pathToReturn;
+        }
+        Log.e(
+                SplunkRum.LOG_TAG,
+                "Error creating path " + pathToReturn + " for span buffer, defaulting to parent");
+        return rootDir;
+    }
+
+    private void cleanUpUnsentBackgroundSpan() {
```
breedx-splk, line 120: Let's pluralize this to `cleanUpUnsentBackgroundSpans()`

splunk-otel-android/src/main/java/com/splunk/rum/StartTypeAwareSpanStorage.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public StartTypeAwareSpanStorage(
+            VisibleScreenTracker visibleScreenTracker, FileUtils fileUtils, File rootDir) {
+        this.visibleScreenTracker = visibleScreenTracker;
+        this.fileUtils = fileUtils;
+        this.rootDir = rootDir;
+        this.spanDir = fileUtils.getSpansDirectory(rootDir);
+        this.uniqueId = UUID.randomUUID().toString();
+        // if new id then background span directory with old ids can be deleted
+        cleanUpUnsentBackgroundSpan();
```
breedx-splk, line 48: I think it's usually considered bad practice to have a side effect like this just from instance creation. I can think of two options -- 

1. Make the method accessible and call it from the `RumInitializer` right after creation.
2. Make a static `create()` factory method that does the two things -- create the instance to return and also, before returning, calls the cleanup.

I think 2 is slightly cleaner.

splunk-otel-android/src/main/java/com/splunk/rum/StartTypeAwareSpanStorage.java
```diff
@@ -0,0 +1,136 @@
...
+                            File destinationFile = new File(spanDir, file.getName());
+                            boolean isMoved = file.renameTo(destinationFile);
+                            Log.d(
+                                    LOG_TAG,
+                                    "Moved background span "
+                                            + file.getPath()
+                                            + " success ? "
+                                            + isMoved
+                                            + " for eventual send");
+                        });
```
breedx-splk, line 89: I think we should remove the subdirectory here as well.
```suggestion
        File subdir = getCurrentSessionBackgroundFile();
        fileUtils.listSpanFiles(subdir)
                .forEach(
                        file -> {
                            File destinationFile = new File(spanDir, file.getName());
                            boolean isMoved = file.renameTo(destinationFile);
                            Log.d(
                                    LOG_TAG,
                                    "Moved background span "
                                            + file.getPath()
                                            + " success ? "
                                            + isMoved
                                            + " for eventual send");
                        });
        fileUtils.safeDelete(subdir);
```

splunk-otel-android/src/main/java/com/splunk/rum/ConfigFlags.java
```diff
@@ -26,7 +26,8 @@ class ConfigFlags {
     private boolean networkMonitorEnabled = true;
     private boolean anrDetectionEnabled = true;
     private boolean slowRenderingDetectionEnabled = true;
-    private boolean backgroundTaskInstrumentationEnabled = true;
+    private boolean isSubprocessInstrumentationEnabled = true;
+    private boolean isBackgroundInstrumentationDeferredUntilForeground = false;
```
breedx-splk, line 30: So the javabean style getters for these should have the `is` prefix, but the variables themselves should not.
```suggestion
    private boolean subprocessInstrumentationEnabled = true;
    private boolean backgroundInstrumentationDeferredUntilForeground = false;
```

splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java
```diff
@@ -368,7 +381,11 @@ boolean isReactNativeSupportEnabled() {
         return configFlags.isReactNativeSupportEnabled();
     }
 
-    boolean isBackgroundTaskReportingDisabled() {
-        return !configFlags.isBackgroundTaskInstrumentationEnabled();
+    boolean isSubprocessInstrumentationEnabled() {
+        return !configFlags.isSubprocessInstrumentationEnabled();
+    }
```
breedx-splk, line 386: I think this method is named wrong now. Because it's inverted, it should be disabled.
```suggestion
    boolean isSubprocessInstrumentationDisabled() {
        return !configFlags.isSubprocessInstrumentationEnabled();
    }
```

breedx-splk:
Thanks for sticking with this, I think we're getting close. I've added a few more comments above, please have a look.

I confirmed that the new behavior does more of what I had expected, thanks for the fix.

Right now I'm thinking about two hurdles:

1. The`SplunkRumBuilder` is kinda mixing the terms "background task" and "subprocess". I am hoping we can standardize this new feature on one of those.
2. Our intention in upstream `opentelemetry-android` is to eventually include the [through-storage buffering](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/disk-buffering) from contrib as a default. I am a little concerned that (ab)using the storage mechanism as a means of filtering background telemetry will further complicate this already complicated goal. I bet we can make something work though.

splunk-otel-android/src/main/java/com/splunk/rum/StartTypeAwareSpanStorage.java
rezastallone: I agree, 2nd solution is better

rezastallone:
Thanks again for the review!

> The SplunkRumBuilder is kinda mixing the terms "background task" and "subprocess". I am hoping we can standardize this new feature on one of those.

I am not sure what you mean by that. My current understanding is that the 'subprocess' configuration prevents runtime initialization, while the 'background task' configuration filters background spans through directory manipulation. Are you suggesting that you want to standardize both configurations in the hope that they apply the same solution?

> Our intention in upstream opentelemetry-android is to eventually include the [through-storage buffering](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/disk-buffering) from contrib as a default. I am a little concerned that (ab)using the storage mechanism as a means of filtering background telemetry will further complicate this already complicated goal. I bet we can make something work though.

Right now, I cannot think of a solution that would be less invasive or have a lesser impact on the overall behavior of the Android RUM SDK. I am open to suggestions. However, if you believe that this solution is both safe and suitable for our current situation, and if you still have concerns about merging it into this repository outside bug or behavior correctness , we would like to proceed with cherry-picking this feature and applying it to our fork. This way, the impact will be confined to our use case and will not affect the broader community of Splunk RUM users. Thank you.

rezastallone:
> The SplunkRumBuilder is kinda mixing the terms "background task" and "subprocess". I am hoping we can standardize this new feature on one of those.

With regard to this point, I would recommend removing the 'subprocess' configuration entirely and instead rely on the 'background task' configuration because spans originating from subprocesses could be valuable, especially if they are initiated by genuine user interactions.

breedx-splk:
>> The SplunkRumBuilder is kinda mixing the terms "background task" and "subprocess". I am hoping we can standardize this new feature on one of those.
>
> I am not sure what you mean by that.

Just literally that method `disableBackgroundTaskReporting()` is changing the config flag with `disableSubprocessInstrumentation()`. The terms "task" and "subprocess" are being intermixed there. Maybe that's the only place. The field on ConfigFlags used to be called `backgroundTaskInstrumentationEnabled` and got renamed to `isSubprocessInstrumentationEnabled` in this PR. If you feel that subprocess is a more accurate descriptor, let's make it consistent. 

If you're worried about changing an existing builder method name, you could add an override and mark the task one deprecated if you'd like.

> Right now, I cannot think of a solution that would be less invasive or have a lesser impact on the overall behavior of the Android RUM SDK.

That's completely pragmatic, and we won't make that a showstopper for merging. It complicates things, but we will work through that when the time comes. I think we can probably merge soon, just a couple other talking points up there...so I don't think you _need_ to cherry pick.

Thanks again for sticking with it.

rezastallone:
> The terms "task" and "subprocess" are being intermixed there. 

Understood. I have rephrased the term 'backgroundTaskReporting' to 'subprocessInstrumentation' in several instances because I believe it is a more accurate description.

> It complicates things, but we will work through that when the time comes. I think we can probably merge soon, just a couple other talking points up there...so I don't think you need to cherry pick.

Thank you once again for your review; we genuinely appreciate it. We hope that this pull request will be beneficial not only for our specific use case but also for other members of the Splunk RUM community who may encounter a similar situation in the future.

rezastallone:
There seems to be a problem with the pipeline again because local build success @breedx-splk

breedx-splk:
Yeah sorry, #658 had to get merged first to fix the main HEAD build. You can cheery pick those commits and things should fix right up.

akubik-splunk:
Hello @rezastallone 
Could you sign the CLA which is no attached to the PR - we have changed our tool for tracking the signatures to be integrated with the repository 
Thank You

rezastallone:
[CLA.md](https://github.com/signalfx/splunk-otel-android/files/12920047/CLA.md)

rezastallone:

breedx-splk (approved):

breedx-splk:
Oh @rezastallone there is one more thing in the way of merging -- as a security measure, we require all commits to be signed. You can learn more about this feature here: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

I think you should be able to amend the commits to provide signing on the existing commits. 🙏🏻 Thanks.

PR #698, opened 2023-11-28
Change session-based probabilistic sampling computation

breedx-splk:
We stray from the path here a little bit by introducing a bespoke sampler that differs from the upstream implementation. This is a stop-gap temporary measure to better align with web and iOS. Sessions sampling decisions can now be more consistent between the platforms.

See https://github.com/signalfx/splunk-otel-ios/pull/185 for a similar change in iOS that this borrowed from.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/internal/UInt32QuadXorTraceIdRatioSampler.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/internal/UInt32QuadXorTraceIdRatioSampler.java
@@ -63,6 +63,7 @@ public class UInt32QuadXorTraceIdRatioSampler implements Sampler {
         } else if (ratio == 1.0) {
             idUpperBound = Long.MAX_VALUE;
         } else {
+            // ratio * UInt32 max value
             idUpperBound = (long) (ratio * 0xFFFFFFFFL);
         }
         String description =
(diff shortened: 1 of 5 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/internal/UInt32QuadXorTraceIdRatioSampler.java
```diff
@@ -0,0 +1,112 @@
...
+            idUpperBound = Long.MIN_VALUE;
+        } else if (ratio == 1.0) {
+            idUpperBound = Long.MAX_VALUE;
+        } else {
+            // ratio * UInt32 max value
+            idUpperBound = (long) (ratio * 0xFFFFFFFFL);
+        }
+        String description =
+                String.format(
+                        Locale.getDefault(), "UInt32QuadXorTraceIdRatioSampler{radio:%f}", ratio);
```
johnbley, line 71: s/radio/ratio/

johnbley (changes requested):

splunk-otel-android/src/main/java/com/splunk/rum/internal/UInt32QuadXorTraceIdRatioSampler.java
breedx-splk: 📻 oh now I get it.

johnbley (approved):

PR #739, opened 2024-01-25
Added possibility to provide custom headers supplier and disabling requests gzip compression

magdamagda:
Description
Added possibility to provide custom headers to requests. It is needed for example to provide auth headers.
Also added possibility to disable gzip compression as it is not always supported by splunk server. 

Changes

- Added two additional configurations in SplunkRumBuilder: setHeadersSupplier() and disableGzipCompression().
- Added request interceptor: CustomHeadersRequestInterceptor, that adds headers from headers supplier to the request.
- Configuring OkHttpSender in RumInitializer based on values in SplunkRumBuilder. If header supplier is provided, then CustomHeadersRequestInterceptor is added. Compression is enabled or disabled based on value of SplunkRumBuilder::gzipCompressionEnabled

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/CustomHeadersRequestInterceptor.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/CustomHeadersRequestInterceptor.java
@@ -1,0 +1,33 @@
+package com.splunk.rum;
+
+import androidx.annotation.NonNull;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class CustomHeadersRequestInterceptor implements Interceptor {
+
+    @NonNull
+    private final Supplier<Map<String, String>> headersSupplier;
+
+
+    public CustomHeadersRequestInterceptor(@NonNull Supplier<Map<String, String>> headersSupplier) {
+        this.headersSupplier = headersSupplier;
+    }
+
+    @NonNull
+    @Override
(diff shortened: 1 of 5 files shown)


breedx-splk:
Hi @magdamagda thanks for the submission. Before we can accept a 3rd party contribution, you must first sign our CLA (see above).

Also, I'm very curious about your use case. The http sender that you have modified is the one that is used to send zipkin spans to Splunk RUM ingest (beacon endpoint). That ingest endpoint doesn't require additional headers, and we prefer to use gzip on the wire because it saves on bandwidth from user devices.

Can you help us understand the motivation behind the changes here? Thanks.

magdamagda:
> Hi @magdamagda thanks for the submission. Before we can accept a 3rd party contribution, you must first sign our CLA (see above).
> 
> Also, I'm very curious about your use case. The http sender that you have modified is the one that is used to send zipkin spans to Splunk RUM ingest (beacon endpoint). That ingest endpoint doesn't require additional headers, and we prefer to use gzip on the wire because it saves on bandwidth from user devices.
> 
> Can you help us understand the motivation behind the changes here? Thanks.

Hi, thank you for quick response.
We use splunk, that is behind api gateway and it requires auth header.
I am still validating if gzip compressions can be enabled in our case.

breedx-splk:
@magdamagda Ok, that helps me understand, thanks for clarifying.

Because of CLA and because I'm not completely confident in how we may want to adapt this in the future with OTLP export coming soon, I'm going to propose another idea in a PR and I'll tag to you to see if it will work for you.

breedx-splk:
@magdamagda Can you please have a look at #742 and leave a comment there if you think that approach will work for you?

PR #788, opened 2024-03-28
Initial experimental support for OTLP export.

breedx-splk:
Adds initial support for OTLP export from Splunk Android RUM instrumentation. This is not enabled by default, and is currently incompatible with the disk buffering feature. As such, if disk buffering is abled, OTLP exporting will not work and a warning will be logged.

There are not yet any beacon endpoints that will correctly handle this data anyway, but a collector can be used to inspect the payloads.

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
@@ -46,6 +46,7 @@ import io.opentelemetry.android.instrumentation.slowrendering.SlowRenderingDetec
 import io.opentelemetry.android.instrumentation.startup.AppStartupTimer;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.exporter.logging.LoggingSpanExporter;
+import io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporter;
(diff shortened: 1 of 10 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -392,18 +404,37 @@ SpanExporter getToDiskExporter(SpanStorage spanStorage) {
...
-                            .setLocalIpAddressSupplier(() -> null)
-                            .setSender(buildCustomizedZipkinSender())
-                            .build();
-                });
+        return new LazyInitSpanExporter(exporterSupplier);
+    }
+
+    @NonNull
+    private Supplier<SpanExporter> supplyOtlpExporter() {
+        // TODO: Do we want/need to append the auth query parameter like zipkin?
```
johnbley, line 418: Presumably this will get nailed down as we collaborate with the backend team?

johnbley (approved):

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
breedx-splk: Actually I got confirmation today that it works the same way, so I should update this. How about I do it in a follow-up PR?

PR #802, opened 2024-04-10
Placeholder for 2.x work

breedx-splk:
Integrating with upstream instrumentation 2.x and latest sdk and semconv.


breedx-splk:
💥

PR #806, opened 2024-04-11
OTLP fixes for Resource and SpanEvents.

breedx-splk:
This prevents the SpanDataModifier from flattening Resource and SpanEvents into (zipkin) attributes when experimental OTLP support is enabled.


johnbley (approved):
LGTM.  The repeated checks on `otlpEnabled` suggest an interface with two implementations, but I won't belabor that since there will only be one implementation soon enough.

PR #829, opened 2024-05-03
Make the build look more like upstream.

breedx-splk:
This also simplifies the release process. There was some historical reason (I can't remember exactly) why we were manually logging into sonatype and pushing the close/release cycle....which is tedious toilsome manual work. Since the upstream repos (including android!) have been auto-closing/releasing, it's time to try that here now.

If this goes well, the next step could be to stage the release page with the gh cli (also like upstream)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -42,7 +42,9 @@ build:
   extends: .prepare-android-environment
   script:
     - touch local.properties
-    - ./gradlew build publish -PmavenCentralUsername=$SONATYPE_USERNAME -PmavenCentralPassword=$SONATYPE_PASSWORD
+    - export SONATYPE_USER="$SONATYPE_USERNAME"
+    - export SONATYPE_KEY="$SONATYPE_PASSWORD"
+    - ./gradlew build publish
 
 sast-scan:
   stage: verify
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -42,7 +42,9 @@ build:
   extends: .prepare-android-environment
   script:
     - touch local.properties
-    - ./gradlew build publish -PmavenCentralUsername=$SONATYPE_USERNAME -PmavenCentralPassword=$SONATYPE_PASSWORD
+    - export SONATYPE_USER="$SONATYPE_USERNAME"
```
johnbley, line 45: Why not just align the gitlab variable names with what the tooling expects?

johnbley (approved):

RELEASING.md
```diff
@@ -14,19 +14,14 @@ This is the process to use to do a release:
    assumes that the remote is named `origin`, if you named yours differently you might have to push
    the tag manually.
 
-3) Wait for gitlab to run the release job.
+3) Wait for gitlab to run the release job. If all goes well, it will automatically close
+   and release the "staging" repository...which means the build has been published to sonatype
+   and will appear in maven with in a day or two at most (typically a few hours).
 
-4) Log in to `oss.sonatype.org` with a profile that has permissions to see and release the `com.splunk`
```
laurit, line 19: happy to see the manual steps gone

laurit (approved):

PR #851, opened 2024-06-06
Jetpack compose support sketch

jtmalinowski:


jtmalinowski:
Currently builds and runs, but crashes on navigation to the JetpackComposeActivity, presumably because of some missing runtime dependency.

jtmalinowski:
@breedx-splk @laurit hey!

jtmalinowski:

splunk-otel-android/src/main/java/com/splunk/rum/ExplicitVisibleScreenNameTracker.java
```diff
@@ -0,0 +1,39 @@
+package com.splunk.rum;
+
+import androidx.annotation.Nullable;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.opentelemetry.android.instrumentation.activity.VisibleScreenTracker;
+
+public class ExplicitVisibleScreenNameTracker extends VisibleScreenTracker {
```
breedx-splk, line 9: I think that this type of polymorphic inheritance can become problematic. Would much prefer to use delegation, and I think that long term perhaps `VisibleScreenTracker` needs to migrate to an interface (or two). I also wouldn't want to count on `VisibleScreenTracker` remaining an external API class for too long, things are in flux over there.

sample-app/src/main/java/com/splunk/android/sample/FirstFragment.java
```diff
@@ -184,12 +188,22 @@ public LiveData<String> getSessionId() {
     public void onDestroyView() {
         super.onDestroyView();
         binding = null;
+
+        SplunkRum.getInstance().setScreenName(null);
```
breedx-splk, line 192: I think we want to keep the existing Fragments unchanged and if anything make a new one to demonstrate the behavior.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -86,7 +88,7 @@ class RumInitializer {
     }
 
     SplunkRum initialize(Looper mainLooper) {
-        VisibleScreenTracker visibleScreenTracker = new VisibleScreenTracker();
+        ExplicitVisibleScreenNameTracker visibleScreenTracker = new ExplicitVisibleScreenNameTracker();
```
breedx-splk, line 91: This is also something that will need to change, based on divergence from the upstream base code. The `VisibleScreenTracker` is also created by the `OpenTelemetryRumBuilder`, so we'd need to figure out how to NOT have two instances in play.

breedx-splk (commented):
I think we should hold on this for now. Once the upstream repo has finished the instrumentation refactor/reorg, we should circle back to see how we do this custom screen name setting up there. 

If we absolutely _need_ to proceed with this for some immediate thing, we must be clear that it's experimental and will break/change shortly.

jtmalinowski:
> we must be clear that it's experimental and will break/change shortly.

In browser RUM we had `experimental` prefix for cases like this, so the signature would become:

```
public void experimentalSetScreenName(String screenName) {
```

I think it's fair to assume that this functionality might be unavailable in some versions as OTel is being restructured, I'll rebuild it when we know the new API. Perhaps even OTel itself will have something for Jetpack Compose in the near future.

So overall I think it's a good compromise, where customers are able to instrument Jetpack Compose via a workaround and under the new API we'll look at it again.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
jtmalinowski: Definitely! For now the only usage, ie. https://github.com/open-telemetry/opentelemetry-android/blob/main/android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java#L484 is addressed, because we replace their `ScreenAttributesSpanProcessor` with our own.

splunk-otel-android/src/main/java/com/splunk/rum/ExplicitVisibleScreenNameTracker.java
jtmalinowski: Hm, how would you implement delegation here? For example when we construct `StartTypeAwareMemorySpanBuffer` it wants an instance of `VisibleScreenTracker`, same with `ScreenAttributesSpanProcessor`, right? So with the existing API I need something that `extends VisibleScreenTracker` anyway? I'm open to ideas.

jtmalinowski:
I managed to configure the kotlin compiler after all, so the example is an actual Jetpack Compose Activity.

sample-app/src/main/java/com/splunk/android/sample/FirstFragment.java
jtmalinowski: I just added a link to my new activity, where I added Jetpack Compose Navigation, so I think this is now addressed.

splunk-otel-android/src/main/java/com/splunk/rum/ExplicitVisibleScreenNameTracker.java
jtmalinowski: Okay! Moved the logic to a span processor, so we don't have the inheritance anymore.

splunk-otel-android/src/main/java/com/splunk/rum/NoOpSplunkRum.java
```diff
@@ -79,4 +79,9 @@ public void integrateWithBrowserRum(WebView webView) {
     void flushSpans() {
         // no-op
     }
+
+    @Override
+    public void experimentalSetScreenName(String screenName) {
+        // no-op
+    }
```
breedx-splk, line 86: There is an overload for this that also needs no-op coverage.

```suggestion
    @Override
    public void experimentalSetScreenName(String screenName) {
        // no-op
    }
    
    @Override
    public void experimentalSetScreenName(String screenName, String spanType) {
      // no-op
    }
```

splunk-otel-android/src/main/java/com/splunk/rum/SplunkRum.java
```diff
@@ -113,6 +117,23 @@ static SplunkRum initialize(SplunkRumBuilder builder, Application application) {
         return INSTANCE;
     }
 
+    public void experimentalSetScreenName(String screenName, String spanType) {
```
breedx-splk, line 120: These two new public API methods needs javadoc so that users can get a better idea about what they do and when they should/shouldn't use these.

breedx-splk (approved):
Ok let's experiment!

PR #857, opened 2024-06-13
Add new experimental api to customize the currently visible screen name.

breedx-splk:
This adds a new api to the `SplunkRumBuilder` called `setExperimentalCurrentScreenCustomizer()`. A user can use this to pass a customizer which allows them to take full control over determining the name of the current screen. A new interface called `CurrentlyVisibleScreen` is introduced, to prevent additional use of or coupling to the existing upstream `VisibleScreenTracker`. Additionally, a new `SpanProcessor` called `CurrentlyVisibleScreenAttributeSpanProcessor` is added (to replace `ScreenAttributesAppender`), which does not couple to the `VisibleScreenTracker` but instead uses an instance of `CurrentlyVisibleScreen`.

Specifying the "current screen name" can be done thought delegation to the existing/default `CurrentlyVisibleScreen` instance and/or through screen tracking mechanism a user might wish to build.

This is an alternate approach to #851 that does not subclass `VisibleScreenTracker`. Instead, it uses the more established OpenTelemetry java pattern of allowing the user to pass a customizer. This customizer is given control of determining the name of the "currently visible screen".

This approach is NOT opinionated about specific frameworks (such as Jetpack compose), and there is no built-in tracking of lifecycle for unknown components. The user of this new experimental API is responsible for tracking their own navigation state and determining when a "screen" has changed, and to generate the appropriate events (create, pause, destroy, whatever their framework offers).

### DIFF
--- a/sample-app/src/main/java/com/splunk/android/sample/SampleApplication.java
+++ b/sample-app/src/main/java/com/splunk/android/sample/SampleApplication.java
@@ -80,6 +80,8 @@ public class SampleApplication extends Application {
                                                 return chain.proceed(requestBuilder.build());
                                             });
                         })
+                .setExperimentalCurrentScreenCustomizer(vs ->
+                        () -> "customized: " + vs.get())
                 .build(this);
     }
 }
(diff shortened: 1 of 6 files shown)


breedx-splk:
So no feedback on this, let's just go ahead and experiment with #851 then.

PR #878, opened 2024-07-12
Remove several configurations from fossa scanning.

breedx-splk:
Reduce false positives.


laurit:
Imo you are listing way too many configurations. I'd add only the configuration with runtime dependencies.

breedx-splk:
> Imo you are listing way too many configurations. I'd add only the configuration with runtime dependencies.

Do you have a recipe for determining which configurations result in runtime dependencies? I don't know which ones...

laurit:
> Do you have a recipe for determining which configurations result in runtime dependencies? I don't know which ones...

You could eliminate the ones that you definitely don't need. Like anything for tests, for compiling and for tooling like annotation processors. You could list what dependencies the configurations have and compare them with what https://repo1.maven.org/maven2/com/splunk/splunk-otel-android/1.6.0/splunk-otel-android-1.6.0.pom has. Probably the only one that really matters is `runtimeClasspath`

breedx-splk:
Ok, thanks for the advice. It might still be too big for your liking, but I shrunk it down considerably.

PR #963, opened 2024-09-18
Fix maven central link

breedx-splk:
And use reference forms, which matches the core repo. 🤷🏻


README.md
```diff
@@ -10,24 +10,12 @@
   </strong>
 </p>
 
-<p align="center">
```
laurit, line 13: I think the original code looked nicer because it is center aligned

laurit (approved):

README.md
breedx-splk: I agree, and I kinda hate this reference syntax anyway...

PR #993, opened 2024-10-01
Added versionCode, applicationID, and customUUID from AndroidManifest.XML (if it exists) to crash and ANR spans

tonzhan2:
Implemented an extractor class, which will fetch the app's applicationID and versionCode, and a custom uuid (if it exists in the app manifest as metadata under key "SPLUNK_OLLY_CUSTOM_UUID").

These will be added as attributes to the ANR and Crash reporters in order to help correlate and identify stacktraces with their correct mapping files. 

Added unit tests and also manually tested with sample app and verified

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.splunk.rum;
 
 import android.app.Application;
@@ -7,30 +23,28 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.os.Bundle;
 import android.util.Log;
-
 import androidx.annotation.Nullable;
 
 public class ErrorIdentifierExtractor {
 
     private static final String SPLUNK_UUID_MANIFEST_KEY = "SPLUNK_OLLY_CUSTOM_UUID";
-    @Nullable
-    private static ErrorIdentifierExtractor instance = null;
+    @Nullable private static ErrorIdentifierExtractor instance = null;
 
     private final Context context;
     private final PackageManager packageManager;
     private final ApplicationInfo applicationInfo;
 
-    @Nullable
-    private final String applicationId;
-    @Nullable
-    private final String versionCode;
-    @Nullable
-    private final String customUUID;
+    @Nullable private final String applicationId;
+    @Nullable private final String versionCode;
+    @Nullable private final String customUUID;
 
-    private ErrorIdentifierExtractor(Application application) throws PackageManager.NameNotFoundException {
+    private ErrorIdentifierExtractor(Application application)
+            throws PackageManager.NameNotFoundException {
         this.context = application.getApplicationContext();
         this.packageManager = context.getPackageManager();
-        this.applicationInfo = packageManager.getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
+        this.applicationInfo =
+                packageManager.getApplicationInfo(
+                        context.getPackageName(), PackageManager.GET_META_DATA);
 
         // Initialize values
         this.applicationId = applicationInfo.packageName;
@@ -44,7 +58,9 @@ public class ErrorIdentifierExtractor {
             try {
                 instance = new ErrorIdentifierExtractor(application);
             } catch (PackageManager.NameNotFoundException e) {
-                Log.e(SplunkRum.LOG_TAG, "Failed to initialize ErrorIdentifierExtractor: " + e.getMessage());
+                Log.e(
+                        SplunkRum.LOG_TAG,
+                        "Failed to initialize ErrorIdentifierExtractor: " + e.getMessage());
             }
         }
         return instance;
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
@@ -1,0 +1,90 @@
+package com.splunk.rum;
+
+import android.app.Application;
+import android.content.Context;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.os.Bundle;
+import android.util.Log;
+
+import androidx.annotation.Nullable;
+
+public class ErrorIdentifierExtractor {
+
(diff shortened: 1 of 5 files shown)


tonzhan2:

splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
```diff
@@ -0,0 +1,105 @@
...
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.os.Bundle;
+import android.util.Log;
+import androidx.annotation.Nullable;
+
+public class ErrorIdentifierExtractor {
+
+    private static final String SPLUNK_UUID_MANIFEST_KEY = "SPLUNK_OLLY_CUSTOM_UUID";
```
breedx-splk, line 30: Ones not L's! :)

```suggestion
    private static final String SPLUNK_UUID_MANIFEST_KEY = "SPLUNK_O11Y_CUSTOM_UUID";
```

splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
```diff
@@ -0,0 +1,105 @@
...
+                packageManager.getApplicationInfo(
+                        context.getPackageName(), PackageManager.GET_META_DATA);
+
+        this.applicationId = applicationInfo.packageName;
+        this.versionCode = retrieveVersionCode();
+        this.customUUID = retrieveCustomUUID();
+    }
+
+    @Nullable
+    public static synchronized ErrorIdentifierExtractor getInstance(Application application) {
```
breedx-splk, line 55: Prefer never to have `synchronized` on public methods. If it's really necessary to make this method thread safe, then let's use a private lock or concurrency classes.

I'd also question if this class really needs to be a singleton. It's probably only ever used at startup/creation time, and is the operation really that expensive?

I also think it's a bummer that this class has more than one responsibility -- it's both the data class that contains the 3 fields that matter (appId, code, uuid), but is also the thing that knows how to extract those from the application context. Might be nice to split these responsibilities apart.

splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
```diff
@@ -0,0 +1,105 @@
...
+import androidx.annotation.Nullable;
+
+public class ErrorIdentifierExtractor {
+
+    private static final String SPLUNK_UUID_MANIFEST_KEY = "SPLUNK_OLLY_CUSTOM_UUID";
+    @Nullable private static ErrorIdentifierExtractor instance = null;
+
+    private final Context context;
+    private final PackageManager packageManager;
+    private final ApplicationInfo applicationInfo;
```
breedx-splk, line 35: I think these 3 fields don't really need to be on this class...and it's a symptom of the class being more than one thing.

splunk-otel-android/src/test/java/com/splunk/rum/ErrorIdentifierExtractorTest.java
```diff
@@ -0,0 +1,85 @@
...
+        mockApplicationInfo.metaData = mockMetadata;
+
+        when(mockPackageManager.getApplicationInfo("com.splunk.test", PackageManager.GET_META_DATA))
+                .thenReturn(mockApplicationInfo);
+        when(mockMetadata.getString(SPLUNK_UUID_MANIFEST_KEY)).thenReturn("test-uuid");
+
+        mockPackageInfo.versionCode = 123;
+        when(mockPackageManager.getPackageInfo("com.splunk.test", 0)).thenReturn(mockPackageInfo);
+
+        extractor = ErrorIdentifierExtractor.getInstance(mockApplication);
```
breedx-splk, line 65: To minimize test interactions, please make sure that you create a new instance of the test object explicitly within each test method.

splunk-otel-android/src/test/java/com/splunk/rum/ErrorIdentifierExtractorTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        mockApplicationInfo.packageName = "com.splunk.test";
+        mockApplicationInfo.metaData = mockMetadata;
+
+        when(mockPackageManager.getApplicationInfo("com.splunk.test", PackageManager.GET_META_DATA))
+                .thenReturn(mockApplicationInfo);
+        when(mockMetadata.getString(SPLUNK_UUID_MANIFEST_KEY)).thenReturn("test-uuid");
+
+        mockPackageInfo.versionCode = 123;
+        when(mockPackageManager.getPackageInfo("com.splunk.test", 0)).thenReturn(mockPackageInfo);
```
breedx-splk, line 63: It would be cool if the package name string wasn't repeated here. Maybe create a constant?

splunk-otel-android/src/test/java/com/splunk/rum/ErrorIdentifierExtractorTest.java
```diff
@@ -0,0 +1,85 @@
...
+        assertEquals("123", extractor.getVersionCode());
+    }
+
+    @Test
+    public void testGetCustomUUID() {
+        extractor = ErrorIdentifierExtractor.getInstance(mockApplication);
+
+        assert extractor != null;
+        assertEquals("test-uuid", extractor.getCustomUUID());
+    }
```
breedx-splk, line 84: I think we're missing coverage for the @Nullable return cases for these 3 things...

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
```diff
@@ -286,45 +291,88 @@ private Resource createSplunkResource() {
...
+
+                    if (applicationId != null && !applicationId.isEmpty()) {
+                        builder.addAttributesExtractor(constant(APPLICATION_ID_KEY, applicationId));
+                    }
+                    if (versionCode != null && !versionCode.isEmpty()) {
+                        builder.addAttributesExtractor(constant(APP_VERSION_CODE_KEY, versionCode));
+                    }
+                    if (uuid != null && !uuid.isEmpty()) {
+                        builder.addAttributesExtractor(constant(SPLUNK_OLLY_UUID_KEY, uuid));
+                    }
```
breedx-splk, line 356: So if the `ErrorIdentifierExtractor` were to implement `AttributesExtractor` AND not be nullable, then this could just read something like
```java
.addAttributesExtractor(new CrashComponentExtractor())
.addAttributesExtractor(new ErrorIdentifierExtractor(application))
```
...which I think is considerably cleaner and follows the existing/established patterns better.

breedx-splk (changes requested):
Thanks @tonzhan2.

splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
tonzhan2: I would have liked to do that, would be much cleaner. However we can’t add these as attribute extractors. 

CrashReporterBuilder’s addAttributesExtractor takes in an AttributeExtractor that takes an objectCrashDetails object.
ANRDetectorBuilder’s addAttributeExtractor takes in an AttributeExtractor that takes a StackTraceElements object

To implement this as an attributesExtractor we’d have to pass in the Application object. And then add an additional addAttributesExtractor method to both builder classes. This stuff is in the otel android repo

Down the line when this is pushed upstream in otel android, we can add those changes to the otel android repo and then streamline this part. But by then we really can just add all this straight to otel android.

The nullable annotations are a bit verbose. But just trying to catch all edge cases in case there are any issues fetching any of the classes that we use to look up these fields.

breedx-splk (approved):
Thanks, this looks like a pretty good start now, appreciate it.

PR #1077, opened 2024-11-20
fix: multiple server-tracing headers aren't handled

jtmalinowski:
When multiple `server-tracing` headers are present, handle all of them. Previously only the first instance would be handled.

### DIFF
--- a/splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
+++ b/splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
@@ -38,28 +38,62 @@ class RumResponseAttributesExtractorTest {
     void spanDecoration() {
         Request fakeRequest = mock(Request.class);
         Response response =
-                new Response.Builder()
-                        .request(fakeRequest)
-                        .protocol(Protocol.HTTP_1_1)
-                        .message("hello")
-                        .code(200)
-                        .addHeader("Server-Timing", "othervalue 1")
-                        .addHeader("Server-Timing", "traceparent;desc=\"00-9499195c502eb217c448a68bfe0f967c-fe16eca542cd5d86-01\"")
-                        .addHeader("Server-Timing", "othervalue 2")
+                getBaseRuestBuilder(fakeRequest)
+                        .addHeader(
+                                "Server-Timing",
+                                "traceparent;desc=\"00-00000000000000000000000000000001-0000000000000001-01\"")
                         .build();
-
-        RumResponseAttributesExtractor attributesExtractor =
-                new RumResponseAttributesExtractor(new ServerTimingHeaderParser());
-        AttributesBuilder attributesBuilder = Attributes.builder();
-        attributesExtractor.onStart(attributesBuilder, Context.root(), fakeRequest);
-        attributesExtractor.onEnd(attributesBuilder, Context.root(), fakeRequest, response, null);
-        Attributes attributes = attributesBuilder.build();
+        Attributes attributes = performAttributesExtraction(fakeRequest, response);
 
         assertThat(attributes)
                 .containsOnly(
                         entry(COMPONENT_KEY, "http"),
-                        entry(LINK_TRACE_ID_KEY, "9499195c502eb217c448a68bfe0f967c"),
-                        entry(LINK_SPAN_ID_KEY, "fe16eca542cd5d86"));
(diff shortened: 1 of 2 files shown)


splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
```diff
@@ -36,22 +36,20 @@ class RumResponseAttributesExtractorTest {
...
         Response response =
                 new Response.Builder()
                         .request(fakeRequest)
                         .protocol(Protocol.HTTP_1_1)
                         .message("hello")
                         .code(200)
-                        .addHeader("Server-Timing", "headerValue")
+                        .addHeader("Server-Timing", "othervalue 1")
+                        .addHeader("Server-Timing", "traceparent;desc=\"00-9499195c502eb217c448a68bfe0f967c-fe16eca542cd5d86-01\"")
+                        .addHeader("Server-Timing", "othervalue 2")
```
breedx-splk, line 48: Do we know the circumstances that would cause this to happen? Seems a little strange, but doesn't violate the http spec, so I suppose the server side can do all kinds of odd things. 

I think the test should should be enhanced to cover the case were there are 2+ `Server-Timing` headers, both of which are parseable but with different values. The logic currently reads like the last one "wins" .... is this intentional? If so, let's assert that intention with test coverage. 👍🏻  Thanks!

jek-bao-choo:
Thank you @jtmalinowski!

splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
jek-bao-choo: Thank you @breedx-splk for your help.

surbhiia (approved):
Approving this as the solution looks good to me, other than the open question above - if there are multiple valid headers, how should that be handled. Thanks!

splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
jtmalinowski: CDN in the customer's account is adding its own `Server-Timing`. They have their own prefix instead of `traceparent`, so we can distinguish them.

johnbley: In most header-emission apis, it is much easier to simply add a single header line than to find an existing one with the same key and then modify or upsert it.  So, yes, this is probably going to happen.  Likewise, header receipt apis differ in the guarantees they make in terms of how they present this situation (as a unified comma-separated list, as an array of values from a key, etc.). So, I think the client side of this should 
- search exhaustively for a `Server-Timing` metric named `traceparent` (across comma-separate values, across multiple header lines), but
- make no guarantees about which one will be used if multiple (validly formed) `traceparent`s are present (just do whatever is easier given the api surface we have at hand).  

An alternative would be to add *all* found valid values as links, but that would require using actual otel span links (which do not require unique names and can have identical attribute keys/values) instead of our current `link.*` attribute pair.  I'd consider that a future goal.

breedx-splk (approved):
Alright, looking good. Thanks thanks!

breedx-splk:
@jtmalinowski Please do a similar PR against the 1.7.x branch as well.

splunk-otel-android/src/test/java/com/splunk/rum/RumResponseAttributesExtractorTest.java
```diff
@@ -70,21 +102,27 @@ void spanDecoration_noLinkingHeader() {
...
         AttributesBuilder attributesBuilder = Attributes.builder();
-        attributesExtractor.onEnd(attributesBuilder, Context.root(), fakeRequest, response, null);
         attributesExtractor.onStart(attributesBuilder, Context.root(), fakeRequest);
+        attributesExtractor.onEnd(attributesBuilder, Context.root(), fakeRequest, response, null);
         Attributes attributes = attributesBuilder.build();
+        return attributes;
+    }
 
-        assertThat(attributes).containsOnly(entry(COMPONENT_KEY, "http"));
+    private Response.Builder getBaseRuestBuilder(Request fakeRequest) {
```
surbhiia, line 121: nit: missed 'q' in request -> `getBaseRuestBuilder`

PR #1103, opened 2024-12-06
Downgrade androidx to allow lower compileSDK

surbhiia:
- Avoid forcing customers to upgrade to compileSDK = 35 and as a result to force them to upgrade AGP / Gradle. 
- Added renovate rule to ignore upgradation of known culprit dependencies. 

**Note** - We do not need to downgrade compileSDK version but I've downgraded it so next when renovate bumps the version of any of our androidx dependencies, build fails and we can add that dependency to renovate rule for ignoring. This is needed as Google has not mentioned exactly what androidx dependencies are requiring compileSDK = 35. The [documentation](https://developer.android.com/jetpack/androidx/versions) vaguely mentions "Various AndroidX libraries will start requiring compilesdk 35 after August 7th release. If you see AGP warnings when you upgrade compilesdk, you could suppress them."

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -57,6 +57,15 @@
     {
       "matchPackagePrefixes": ["io.opentelemetry.instrumentation"],
       "groupName": "upstream otel instrumentation"
+    },
+    {
+      // Various androidx libraries released after Aug 7th, 2024 requires android compileSDK 35,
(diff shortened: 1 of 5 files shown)


surbhiia:

SenNeonoveNoci (approved):
LGTM

aditi-s3 (approved):
LGTM

PR #1168, opened 2025-02-07
Add custom tracking module

surbhiia:
This PR covers the following:
- Add custom tracking module with two public APIs
- Add tests for same to sample app
- Add RumConstants in :common:otel to hold certain re-usable constant
- Update rum-common-otel-otel published library name to rum-common-otel

### DIFF
--- a/common/otel/build.gradle.kts
+++ b/common/otel/build.gradle.kts
@@ -15,7 +15,7 @@ apply<ConfigAndroidLibrary>()
 apply<ConfigPublish>()
 
 ext {
-    set(artifactIdProperty, "$artifactPrefix${commonPrefix}otel-${project.name}")
+    set(artifactIdProperty, "$artifactPrefix${commonPrefix}${project.name}")
     set(versionProperty, Configurations.sdkVersionName)
 }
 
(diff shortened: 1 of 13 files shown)


SenNeonoveNoci (approved):
The only thing I was considering is whether we should have the integration/instrumentation pair, but for this specific functionality, I think it's okay to have the instrumentation directly.

surbhiia:
> The only thing I was considering is whether we should have the integration/instrumentation pair, but for this specific functionality, I think it's okay to have the instrumentation directly.

Yeah, I considered this as well. As there wasn't any configuration to be supplied to this module at this point, so thought we'll wait on that for now until it's clearly known that we need it.

common/otel/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ apply<ConfigAndroidLibrary>()
 apply<ConfigPublish>()
 
 ext {
-    set(artifactIdProperty, "$artifactPrefix${commonPrefix}otel-${project.name}")
```
tonzhan2, line 18: Maybe I missed the context here but how come we're removing the 'otel' part of this name?

tonzhan2 (approved):

TomasChladekSL (approved):

common/otel/build.gradle.kts
surbhiia: As mentioned in the what PR covers bullet points - The published library name was earlier - rum-common-otel-otel which was changed to rum-common-otel

PR #1172, opened 2025-02-12
Add Track Exception APIs, and Error Identifier Attributes Span Processor

tonzhan2:
- Added trackException and trackException APIs to custom tracking module
- Added buttons to sample app to test the behavior
- Add common:utils module which contains the Error Identifier Extractor logic from the splunk-otel-agent
- Add a Error Identifier Attributes Span Processor to :integration:agent:api module that adds the error identifier attributes to any span that has component = "error" or "crash" (so crashes, ANRs, and custom tracked exceptions)
- Add some rum constants
- Add the new span processor when the agent is initialized.

**Testing notes**
Logged out spans to logcat in AndroidSpanExporter. After clicking test buttons, can see the exception spans generated, with error identifying attributes service.application_id and service.version_code, as well as any additional attributes from calling the trackException API that allows attributes

logs attached: 
[trackExceptionSpans.txt](https://github.com/user-attachments/files/18774873/trackExceptionSpans.txt)


common/otel/src/main/java/com/splunk/sdk/common/otel/internal/RumConstants.kt
```diff
@@ -20,5 +20,12 @@ import io.opentelemetry.api.common.AttributeKey
 
 object RumConstants {
     const val RUM_TRACER_NAME: String = "SplunkRum"
+    const val COMPONENT_ERROR: String = "error"
+    const val COMPONENT_CRASH: String = "crash"
     val WORKFLOW_NAME_KEY: AttributeKey<String> = AttributeKey.stringKey("workflow.name")
+    val COMPONENT_KEY: AttributeKey<String> = AttributeKey.stringKey("component")
```
TomasChladekSL, line 26: I don't think it's a good idea to have attributes as global properties. It goes against the principle of an independent modular architecture.

common/utils/proguard-rules.pro
```diff
@@ -0,0 +1 @@
+-repackageclasses 'com.smartlook.sdk.common.utils'
```
TomasChladekSL, line 1: smartlook?

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierExtractor.kt
```diff
@@ -0,0 +1,78 @@
...
+import android.content.pm.PackageManager
+import android.os.Build
+import android.util.Log
+
+class ErrorIdentifierExtractor(private val application: Application) {
+    private val packageManager: PackageManager = application.packageManager
+    private val applicationInfo: ApplicationInfo?
+
+    init {
+        var appInfo: ApplicationInfo?
```
TomasChladekSL, line 14:
```
applicationInfo = try {
                packageManager.getApplicationInfo(application.packageName, PackageManager.GET_META_DATA)
        } catch (e: Exception) {
            Logger.e(TAG, "Failed to initialize ErrorIdentifierExtractor: ${e.message}")
            null
        }
```

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierExtractor.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    fun extractInfo(): ErrorIdentifierInfo {
+        var applicationId: String? = null
+        val versionCode = retrieveVersionCode()
+        val customUUID = retrieveCustomUUID()
+
+        if (applicationInfo != null) {
+            applicationId = applicationInfo.packageName
+        } else {
+            Log.e(TAG, "ApplicationInfo is null, cannot extract applicationId")
```
TomasChladekSL, line 38: You should use `Logger` class

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierExtractor.kt
```diff
@@ -0,0 +1,78 @@
...
+            }
+        } catch (e: Exception) {
+            Log.e(TAG, "Failed to get application version code", e)
+            return null
+        }
+    }
+
+    private fun retrieveCustomUUID(): String? {
+        if (applicationInfo == null) {
+            Log.e(TAG, "ApplicationInfo is null; cannot retrieve Custom UUID.")
```
TomasChladekSL, line 61: You should use `Logger` class

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierExtractor.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    init {
+        var appInfo: ApplicationInfo?
+        try {
+            appInfo =
+                packageManager.getApplicationInfo(
+                    application.packageName, PackageManager.GET_META_DATA
+                )
+        } catch (e: Exception) {
+            Log.e(
```
TomasChladekSL, line 21: You should use `Logger` class

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierExtractor.kt
```diff
@@ -0,0 +1,78 @@
...
+        val bundle = applicationInfo.metaData
+        if (bundle != null) {
+            return bundle.getString(SPLUNK_UUID_MANIFEST_KEY)
+        } else {
+            Log.e(TAG, "Application MetaData bundle is null")
+            return null
+        }
+    }
+
+    companion object {
```
TomasChladekSL, line 73: private companion object

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierInfo.kt
```diff
@@ -0,0 +1,7 @@
+package com.splunk.sdk.utils
+
+class ErrorIdentifierInfo(
```
TomasChladekSL, line 3: Why it is in `utils` module? Maybe `api`?

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesSpanProcessor.kt
```diff
@@ -0,0 +1,57 @@
...
+import android.util.Log
+import com.splunk.sdk.common.otel.internal.RumConstants
+import com.splunk.sdk.utils.ErrorIdentifierExtractor
+import com.splunk.sdk.utils.ErrorIdentifierInfo
+import io.opentelemetry.context.Context
+import io.opentelemetry.sdk.trace.ReadWriteSpan
+import io.opentelemetry.sdk.trace.ReadableSpan
+import io.opentelemetry.sdk.trace.SpanProcessor
+
+internal class ErrorIdentifierAttributesSpanProcessor(private val application: Application) : SpanProcessor {
```
TomasChladekSL, line 13: The `application` argument is used only one time and only when the instance is created. You do not need to store it as a global property.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesSpanProcessor.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    private var applicationId: String? = null
+    private var versionCode: String? = null
+    private var customUUID: String? = null
+    private var isInitialized = false
+
+    private fun initializeAttributes() {
+        if (!isInitialized) {
```
TomasChladekSL, line 21: The function is always called when the class instance is created.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesSpanProcessor.kt
```diff
@@ -0,0 +1,57 @@
...
+            isInitialized = true
+        }
+    }
+
+    init {
+        initializeAttributes()
+    }
+
+    override fun onStart(parentContext: Context, span: ReadWriteSpan) {
+        if (span.getAttribute(RumConstants.COMPONENT_KEY).equals(RumConstants.COMPONENT_ERROR) ||
```
TomasChladekSL, line 38: You should use `==` operator instead of `equals` function.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesSpanProcessor.kt
TomasChladekSL, line 13: Class named `ErrorIdentifierAttributesSpanProcessor` adds `applicationId`, `versionCode`, `customUUID`. There is no any error. Please consider of using `GlobalAttributeSpanProcessor`.

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierExtractor.kt
```diff
@@ -0,0 +1,78 @@
+package com.splunk.sdk.utils
+
+import android.app.Application
+import android.content.pm.ApplicationInfo
+import android.content.pm.PackageManager
+import android.os.Build
+import android.util.Log
+
+class ErrorIdentifierExtractor(private val application: Application) {
```
TomasChladekSL, line 9: Class named `ErrorIdentifierExtractor` provides `applicationId`, `versionCode`, `customUUID`. There is no any error.

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierInfo.kt
TomasChladekSL, line 3: Class named `ErrorIdentifierInfo` has properties `applicationId`, `versionCode`, `customUUID`. There is no any error.

instrumentation/runtime/customtracking/src/main/java/com/splunk/rum/customtracking/CustomTracking.kt
```diff
@@ -53,6 +52,42 @@ class CustomTracking internal constructor() {
...
+     * handling in your application.
+     *
+     *
+     * This event will be turned into a Span and sent to the RUM ingest along with other,
+     * auto-generated spans.
+     *
+     * @param throwable A [Throwable] associated with this event.
+     */
+    fun trackException(throwable: Throwable) {
+        trackException(throwable, Attributes.empty())
```
TomasChladekSL, line 66: Wouldn't it be better to pass `null`?

TomasChladekSL (changes requested):

instrumentation/runtime/customtracking/src/main/java/com/splunk/rum/customtracking/CustomTracking.kt
tonzhan2: It doesn't make a difference

common/utils/proguard-rules.pro
tonzhan2: renamed to splunk. This was copied over from the proguard-rules.pro files of the other modules in common (etc common:storage). Those will probably have to have the smartlook name renamed at some point

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierInfo.kt
tonzhan2: I removed it entirely, making the three methods in the extractor class directly accessible. Should be fine since each of them is only referenced once. So no need to have this extra API

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesSpanProcessor.kt
tonzhan2: The application ID, version code, and uuid (if the customer adds it) only need to be added to crash and error spans that have stacktraces. This is because these attributes are used to correlate an android stacktrace with the correct mapping file for deobfuscation in the backend.

They do not need to be added to any other spans, as they are only relevant to identifying errors. That is why this class is called Error Identifier Attributes Span Processor, its the span processor that adds the attributes that help identify error spans for the backend and is only added to error and crash spans

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierInfo.kt
tonzhan2: Removed this class

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierExtractor.kt
tonzhan2: Please see above comment:

These fields: 

-application ID
-version code
- uuid (if the customer adds it) 

only need to be added to crash and error spans that have stacktraces. This is because these attributes are used to correlate an android stacktrace with the correct mapping file for deobfuscation in the backend.

They do not need to be added to any other spans, as they are only relevant to identifying errors. That is why this class is called the error identifier extractor. Because it extracts the attributes used to identify an error stacktrace

instrumentation/runtime/customtracking/src/main/java/com/splunk/rum/customtracking/CustomTracking.kt
TomasChladekSL: According to the [source code](https://github.com/open-telemetry/opentelemetry-java/blob/3e1d9536f80603ce00a421b8e1cbc4cd82126738/api/all/src/main/java/io/opentelemetry/api/trace/Span.java#L182), using `Attributes.empty()` requires two additional comparisons to achieve the same result.

TomasChladekSL (approved):

common/utils/src/main/java/com/splunk/sdk/utils/ErrorIdentifierExtractor.kt
```diff
@@ -0,0 +1,65 @@
+package com.splunk.sdk.utils
```
surbhiia, line 1: nit: Should we add "common" word in the package name like the other common packages - com.splunk.sdk.common.utils ?

common/otel/src/main/java/com/splunk/sdk/common/otel/internal/RumConstants.kt
```diff
@@ -20,5 +20,12 @@ import io.opentelemetry.api.common.AttributeKey
...
+
+    val APPLICATION_ID_KEY: AttributeKey<String> = AttributeKey.stringKey("service.application_id")
```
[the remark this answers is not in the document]
surbhiia: Some of these are going to be used by multiple modules, so that's why it's helpful to keep them in a common module. I also came across another class we already have where we can park commonly used attributes (I Missed it in my PR as well where I added workflow.name) - [AttributeConstants](https://github.com/signalfx/splunk-otel-android/blob/feature/next-gen/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/AttributeConstants.kt)

surbhiia (approved):

PR #1176, opened 2025-02-21
Upgrade kotlin, AGP, gradle versions and proguard-rules

surbhiia:
Upgraded kotlin, AGP, gradle versions
- Kotlin and KGP version = 1.8.0
- AGP version = 8.6.0
- Gradle version = 8.7
- minSdk = 24
- targetSdk = 34
-----------------------------------------
- Removed @RequiresApi(Build.VERSION_CODES.LOLLIPOP) i.e. min API level 21 required annotation -  which is obsolete as anyway minSdk = 24
- Adding keep proguard-rules to prevent R8 "Missing class" type compile-time errors. 
- - Based on the decisions made on our R8 support discussion item, we can migrate these to consumer-rules.pro in a later PR.
- -  Added repackageClasses rule only where there were R8 issues on "duplicate classes found" between multiple modules. We can perhaps decide to add it in all modules, can update it as per our decision in upcoming PRs. 
- For `"-dontwarn com.google.auto.value.AutoValue"`, it should ideally be added to the opentelemetry libraries that use it so we and our customers don't need to add it. I will add an issue in relevant repo's to get this corrected, until then keeping the rule it in our app proguard-rules.


common/storage/proguard-rules.pro
```diff
@@ -1 +1,4 @@
--repackageclasses 'com.smartlook.sdk.common.storage'
\ No newline at end of file
+-keep interface com.splunk.sdk.common.storage.IAgentStorage
```
TomasChladekSL, line 1: Please set `isMinifyEnabled = false` in `ConfigAndroidLibrary` and remove all `keep`.

TomasChladekSL (changes requested):

common/storage/proguard-rules.pro
surbhiia: Let's review this item in our discussion document and collaboratively reach a consensus before making a decision and requesting any changes.

TomasChladekSL (approved):

PR #1181, opened 2025-02-26
POC Global Attributes implementation 

tonzhan2:
POC Implementation of global attributes feature.

- added global attributes class and exposed API calls
- implemented only basic set and remove for now
- added global attributes to agent config
- copied a global attribute span processor into the integration:agent:api module and modified slightly so it can process spans from the global attributes class located in integration:agent:api
- tested both adding global attributes from agent config as well as adding and removing via sample app button during runtime

### DIFF
--- a/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributeSpanProcessor2.kt
+++ b/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributeSpanProcessor2.kt
@@ -1,0 +1,34 @@
+package com.splunk.rum.integration.agent.api.attributes
+
+import com.cisco.android.common.utils.extensions.forEachFast
+import com.splunk.rum.integration.agent.api.attributes.GlobalAttributes
+import io.opentelemetry.context.Context
+import io.opentelemetry.sdk.trace.ReadWriteSpan
+import io.opentelemetry.sdk.trace.ReadableSpan
+import io.opentelemetry.sdk.trace.SpanProcessor
+
+class GlobalAttributeSpanProcessor2 : SpanProcessor {
+
+    override fun onStart(parentContext: Context, span: ReadWriteSpan) {
+        // Fetch the current attributes from GlobalAttributes and apply them to the span
+        GlobalAttributes.instance.attributes.forEach { attribute ->
+            when (attribute) {
+                is GlobalAttributes.Attribute.Boolean -> span.setAttribute(attribute.name, attribute.value)
+                is GlobalAttributes.Attribute.Double -> span.setAttribute(attribute.name, attribute.value)
+                is GlobalAttributes.Attribute.Long -> span.setAttribute(attribute.name, attribute.value)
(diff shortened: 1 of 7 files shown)


integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
```diff
@@ -0,0 +1,60 @@
...
+                is Double -> Attribute.Double(key.key, value)
+                else -> throw IllegalArgumentException("Unsupported attribute type")
+            }
+        } else {
+            throw IllegalArgumentException("Key must be either String or AttributeKey")
+        }
+    }
+
+    fun remove(key: String) {
+        _attributes.removeAll { it.name == key }
```
TomasChladekSL, line 48: There will be always at most one element with the `key`.

```
for (i in _attributes.indices)
   if (_attributes[i].name == key) {
      _attributes.removeAt(i)
     break
}
```

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributeSpanProcessor2.kt
```diff
@@ -0,0 +1,34 @@
...
+import io.opentelemetry.context.Context
+import io.opentelemetry.sdk.trace.ReadWriteSpan
+import io.opentelemetry.sdk.trace.ReadableSpan
+import io.opentelemetry.sdk.trace.SpanProcessor
+
+class GlobalAttributeSpanProcessor2 : SpanProcessor {
+
+    override fun onStart(parentContext: Context, span: ReadWriteSpan) {
+        // Fetch the current attributes from GlobalAttributes and apply them to the span
+        GlobalAttributes.instance.attributes.forEach { attribute ->
```
TomasChladekSL, line 14: Please use `forEachFast`

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributeSpanProcessor2.kt
```diff
@@ -0,0 +1,34 @@
+package com.splunk.rum.integration.agent.api.attributes
+
+import com.cisco.android.common.utils.extensions.forEachFast
+import com.splunk.rum.integration.agent.api.attributes.GlobalAttributes
+import io.opentelemetry.context.Context
+import io.opentelemetry.sdk.trace.ReadWriteSpan
+import io.opentelemetry.sdk.trace.ReadableSpan
+import io.opentelemetry.sdk.trace.SpanProcessor
+
+class GlobalAttributeSpanProcessor2 : SpanProcessor {
```
TomasChladekSL, line 10: Please remove `GlobalAttributeSpanProcessor` if possible

TomasChladekSL (commented):
Please take a look

tonzhan2:
Thanks for the comments @TomasChladekSL , This is just a rough proof of concept to ensure everybody is on board with the approach, and not the final implementation. But I'll be sure to incorporate those things in the actual PR

PR #1191, opened 2025-03-17
Global Attributes Implementation

tonzhan2:
Implemented Global Attributes feature according to specs

- implemented global attributes class and attribute storage using AttributeBuilder API
- implemented prefix key management and type checking
- implemented all API calls
- exposed global attributes as an extension to splunkRum instance
- added global attributes to agent config
- created a global attributes span processor
- refactored original global attributes span processors to "SplunkInternalGlobalAttributesSpanProcessor"
- added use of these APIs to buttons in sample app for manual testing

APIs introduced:
```
get(key: AttributeKey<T>)
get(key: String)
set(key: String, value: String)
set(key: String, value: Long)
set(key: String, value: Double)
set(key: String, value: Boolean)
set(key: AttributeKey<T>, value: T & Any)
remove(key: String)
remove(key: AttributeKey<T>) 
removeAll()
setAll(attributes: Attributes)
getAll(): Attributes
```


SenNeonoveNoci (changes requested):

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/internal/processors/GlobalAttributeSpanProcessor.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    override fun isStartRequired(): Boolean {
+        return true
+    }
+
+    override fun onEnd(span: ReadableSpan) {
+    }
+
+    override fun isEndRequired(): Boolean {
+        return true
```
surbhiia, line 31: We can "return false" here as we're doing nothing in onEnd()

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/AgentConfiguration.kt
```diff
@@ -26,7 +26,7 @@ data class AgentConfiguration(
     var deploymentEnvironment: String? = null,
     var enableDebugLogging: Boolean = false,
     var sessionSamplingRate: Double = 1.0,
-    var globalAttributes: Attributes? = null,
+    var globalAttributes: Attributes? = Attributes.empty(),
```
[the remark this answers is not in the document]
tonzhan2: I dont think so. 

It makes more sense for customers to use the standard oTel Attributes class when configuring initial attributes. It keeps the API simple and familiar and also consistent with our `globalAttributes` API methods like `setAll`, where the customer passes in an Attributes object.

Also, in `SplunkRumAgentCore`, when the agent is initialized we are setting these attributes in `agentConfiguration` to become part of the `GlobalAttributes` class' attributes by calling `setAll`, and properly applies our internal key prefixing too:
`        agentConfiguration.globalAttributes?.let { GlobalAttributes.instance.setAll(it) }
`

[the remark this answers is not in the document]
SenNeonoveNoci: @tonzhan2 We can certainly make a change if you believe the proposed API is not ideal for the client. However, this change should not happen within this PR or during the PR review process but rather after a discussion with the team.  

I shared the documentation on Global Attributes with you and Stephen on **February 13th** and specifically informed you about `GlobalAttributes(...)` in `AgentConfiguration` on **March 20th**—both times without any proposals for a change.  

I don't agree that it is more straightforward for clients to use `Attributes` directly. We introduced a dedicated API for handling attributes to make the experience more user-friendly, leveraging `operator fun` to improve usability. Using `Attributes` directly would mean clients would have to interact with different APIs to set global attributes during the installation phase versus during the agent run.  

Feel free to start this discussion with the TO and the team on Slack if you believe the documented API should be changed. Thanks.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRumBuilder.kt
```diff
@@ -73,7 +74,7 @@ class SplunkRumBuilder {
         return this
     }
 
-    fun setGlobalAttributes(attributes: Attributes): SplunkRumBuilder {
+    fun setGlobalAttributes(attributes: GlobalAttributes): SplunkRumBuilder {
```
SenNeonoveNoci, line 77: The legacy API must remain unchanged so that clients using it do not have to make adjustments.  

The API must stay like this:  
```kotlin
fun setGlobalAttributes(attributes: GlobalAttributes): SplunkRumBuilder {
   ...
}
```
However, in the `build()` function, you need to map it to the new `AgentConfiguration`.

SenNeonoveNoci (changes requested):

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRumBuilder.kt
tonzhan2: Okay. I've reverted this method signature, marked it as deprecated and am calling `globalAttributes.setAll(attributes)` in it.

I believe the build() function should be good already as globalAttributes is being passed into the agentConfiguration field

SenNeonoveNoci: The whole class `SplunkRumBuilder` is marked as deprecated there is no need to mark specific functions as deprecated. Please remove the `@Deprecated` for this function.

SenNeonoveNoci (approved):
From the API perspective, it looks good. I haven’t checked the work on the SpanProcessors.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
```diff
@@ -0,0 +1,120 @@
...
+        builder.put(AttributeKey.doubleKey(prefixedKey), value)
+        attributes = builder.build()
+    }
+
+    operator fun set(key: String, value: Boolean) {
+        val prefixedKey = prefixKey(key)
+        val builder = attributes.toBuilder()
+        builder.put(AttributeKey.booleanKey(prefixedKey), value)
+        attributes = builder.build()
+    }
```
surbhiia, line 55: Let's collate all the above set methods into one. Also include the when statements for other allowed value types like StringArray etc:
 
```
operator` fun set(key: String, value: Any) {
        val prefixedKey = prefixKey(key)
        val builder = attributes.toBuilder()

        when (value) {
            is String -> builder.put(AttributeKey.stringKey(prefixedKey), value)
            is Long -> builder.put(AttributeKey.longKey(prefixedKey), value)
            is Double -> builder.put(AttributeKey.doubleKey(prefixedKey), value)
            is Boolean -> builder.put(AttributeKey.booleanKey(prefixedKey), value)
            //Extend when statement to include other allowed value types as well - StringArray, LongArray
            else -> {
                Logger.d(TAG, "Unsupported value type: ${value::class}")
            }
        }

        attributes = builder.build()
    }
```

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    operator fun <T> set(key: AttributeKey<T>, value: T & Any) {
+        val prefixedKey = createPrefixedKey(key)
+        val builder = attributes.toBuilder()
+        if (prefixedKey != null) {
+            builder.put(prefixedKey, value)
+        }
+        attributes = builder.build()
+    }
```
surbhiia, line 64: This can be optimized as follows:
```
  operator fun <T> set(key: AttributeKey<T>, value: T) {
        val prefixedKey = createPrefixedKey(key)
        prefixedKey?.let {
            attributes = attributes.toBuilder().apply { put(it, value) }.build()
        }

    }
```

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
```diff
@@ -0,0 +1,120 @@
+package com.splunk.rum.integration.agent.api.attributes
+
+import io.opentelemetry.api.common.AttributeKey
+import io.opentelemetry.api.common.AttributeType
+import io.opentelemetry.api.common.Attributes
+import com.cisco.android.common.logger.Logger
+
+class GlobalAttributes constructor(
+    @Volatile
+    private var attributes: Attributes = Attributes.empty()
```
surbhiia, line 10: We need to grant write access to this variable to only one thread at a time.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
```diff
@@ -0,0 +1,120 @@
+package com.splunk.rum.integration.agent.api.attributes
+
+import io.opentelemetry.api.common.AttributeKey
+import io.opentelemetry.api.common.AttributeType
+import io.opentelemetry.api.common.Attributes
+import com.cisco.android.common.logger.Logger
+
+class GlobalAttributes constructor(
```
surbhiia, line 8: Multiple instances of this class are created - In `AgentConfiguraton.kt`, `SplunkRumBuilder.kt` and then the one singleton instance, that we intended for, is getting created in `SplunkRumAgentCore`. That doesn't seem like a good pattern. 

I suggest we instead receive a `Attributes` type argument in `AgentConfiguraton.kt`, `SplunkRumBuilder.kt` and use that to call `setAll(Attributes)` in `SplunkRumAgentCore` on our singleton instance and make the `constructor` here `internal`. That way we ensure only one instance of this class is ever created. 

I will add suggested code changes at other places too that are referred in this comment for clarity. Here, we need to change to:
`class GlobalAttributes private constructor(`

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/AgentConfiguration.kt
```diff
@@ -26,7 +26,7 @@ data class AgentConfiguration(
     var deploymentEnvironment: String? = null,
     var enableDebugLogging: Boolean = false,
     var sessionSamplingRate: Double = 1.0,
-    var globalAttributes: Attributes? = null,
+    var globalAttributes: GlobalAttributes = GlobalAttributes(),
```
surbhiia, line 29: This relates to the first comment in `GlobalAttributes` class. Suggested change here is:
`var globalAttributes: Attributes = Attributes.empty(),`

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRumBuilder.kt
```diff
@@ -32,7 +33,7 @@ class SplunkRumBuilder {
     private var realm: String? = null
     private var beaconEndpoint: String? = null
     private var enableDebug: Boolean = false
-    private var globalAttributes: Attributes? = null
+    private var globalAttributes: GlobalAttributes = GlobalAttributes()
```
surbhiia, line 36: This relates to the first comment in `GlobalAttributes` class. Suggested change here is:
`private var globalAttributes: Attributes = Attributes.empty()`

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRumBuilder.kt
```diff
@@ -74,7 +75,7 @@ class SplunkRumBuilder {
     }
 
     fun setGlobalAttributes(attributes: Attributes): SplunkRumBuilder {
-        globalAttributes = attributes
+        globalAttributes.setAll(attributes)
```
surbhiia, line 78: This relates to the first comment in GlobalAttributes class. Suggested change here is:
```
 fun setGlobalAttributes(attributes: Attributes): SplunkRumBuilder {
        globalAttributes = attributes
        return this
    }
```

Additionally, we expect the customer to call this only once on `SplunkRumBuilder` instance while creating the `SplunkRum` via the `SplunkRumBuilder`. In my head, that's sort of understable, but if you think we need to call this out in the customer docs, please make a note of this. 

Otherwise if you think this should allow being called multiple times, please adjust the code to append additional attributes to the `globalAttributes` `var` instead of overriding it.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/internal/SplunkRumAgentCore.kt
```diff
@@ -88,6 +91,11 @@ internal object SplunkRumAgentCore {
         val stateManager = StateManager.obtainInstance(application)
         SessionStartEventManager.obtainInstance(agentIntegration.sessionManager)
 
+        agentConfiguration.globalAttributes.getAll().forEach { key, value ->
+            @Suppress("UNCHECKED_CAST")
+            GlobalAttributes.instance[key as AttributeKey<Any>] = value
+        }
```
surbhiia, line 97: This relates to the first comment in `GlobalAttributes` class.
`GlobalAttributes.instance.setAll(agentConfiguration.globalAttributes)`

app/src/main/java/com/splunk/app/ui/menu/MenuFragment.kt
```diff
@@ -152,6 +161,39 @@ class MenuFragment : BaseFragment<FragmentMenuBinding>() {
...
+            viewBinding.removeAllGlobalAttributes.id -> {
+                SplunkRum.instance.globalAttributes.removeAll()
+            }
+            viewBinding.getAllGlobalAttributes.id -> {
+                AlertDialog.Builder(context)
+                    .setTitle("All Global Attributes")
+                    .setMessage("All global attributes: " + SplunkRum.instance.globalAttributes.getAll().toString())
+                    .setPositiveButton("OK", null)
+                    .show()
+            }
```
surbhiia, line 196: Screenshot 2025-04-01 at 5 28 06 PM

Suggesting the following changes here:
- A separate heading/title for global attributes section (**`Global Attributes`**) to tell it apart from **`Custom Tracking`** section in the UI. 
- A toast after every button press like other buttons, for feedback to users after a user interaction that the button was pressed, so they can know and avoid multiple clicks. 
- Renaming them consistently -> **Get Global Attribute** instead of **Add Global Attribute** for the UI. (Also please rename in code).

surbhiia:
Perhaps a commented out test of adding global attributes via agent config in the App.kt. Basically after you test, leave it as commented out in the PR so others can quickly uncomment and test if needed.

Screenshot 2025-04-01 at 8 02 52 PM

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
```diff
@@ -0,0 +1,120 @@
...
+            set(key as AttributeKey<Any>, value)
+        }
+    }
+
+    fun getAll(): Attributes = attributes
+
+    private fun prefixKey(key: String): String = CUSTOM_PREFIX + key
+
+    @Suppress("UNCHECKED_CAST")
+    private fun <T> createPrefixedKey(key: AttributeKey<T>): AttributeKey<T>? {
```
surbhiia, line 98: nit: Can we please rename this to `createPrefixedAttributeKey` and perhaps use `prefixedAttriuteKey` as the `var` name in functions where we are calling it, instead of `prefixedKey`

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
TomasChladekSL: One function with `Any` is very bad and leads into issues.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/AgentConfiguration.kt
SenNeonoveNoci: This was discussed previously in this PR. The current implementation is in sync with the API specification.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRumBuilder.kt
SenNeonoveNoci: All places that allow setting or editing global attributes in the API will use GlobalAttributes according to the specification.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
aditi-s3: Agreed. Individual set methods help with type safety at compile time

tonzhan2: Yeah, Im inclined to keep it this way. Its more verbose and repetitive, but keeps things a more type safe and clarifies the API more

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/internal/SplunkRumAgentCore.kt
```diff
@@ -87,11 +88,14 @@ internal object SplunkRumAgentCore {
         SessionStartEventManager.obtainInstance(agentIntegration.sessionManager)
 
         val initializer = OpenTelemetryInitializer(application, agentConfiguration.spanFilter)
+            // The GlobalAttributeSpanProcessor must be registered first to ensure that global attributes
```
tonzhan2, line 91: Does the order in which span processors are added impact this?

I wasn't familiar with that, but you're saying that later span processors that are added will take precedent over the global attributes span processor?

SenNeonoveNoci: Yes, the span processors are stored in a list and applied in order. I tested this, and if I place the `GlobalAttributeSpanProcessor` last in the list, and then in `App` I set `globalAttributes["session.id"] = "wrong"`, the `session.id` attribute ends up being `"wrong"`. However, if the `GlobalAttributeSpanProcessor` is added first, the `session.id` attribute is correctly set.

keciman (approved):

surbhiia (approved):

PR #1205, opened 2025-04-02
Consolidate naming around build id field

tonzhan2:
Renamed all methods, variables and values that pertain to the optional unique identifier field that gets set by the user in the AndroidManifest.xml file and included in crash/error span attributes to splunk build id. Now the field that the user sets, as well as the attribute tag that gets sent to the backend are: `splunk.build_id` 

Also updated the method that retrieves this build ID to handle non string values more gracefully and better debug logging instead of error logging

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
@@ -28,6 +28,7 @@ import androidx.annotation.Nullable;
 public class ErrorIdentifierExtractor {
 
     private static final String SPLUNK_BUILD_ID = "splunk.build_id";
+
     private final Application application;
     private final PackageManager packageManager;
     @Nullable private final ApplicationInfo applicationInfo;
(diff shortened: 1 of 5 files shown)


splunk-otel-android/src/main/java/com/splunk/rum/ErrorIdentifierExtractor.java
```diff
@@ -76,16 +77,29 @@ private String retrieveVersionCode() {
...
-            Log.e(SplunkRum.LOG_TAG, "ApplicationInfo is null; cannot retrieve Custom UUID.");
+            Log.e(SplunkRum.LOG_TAG, "ApplicationInfo is null; cannot retrieve Splunk Build ID.");
             return null;
         }
+
         Bundle bundle = applicationInfo.metaData;
-        if (bundle != null) {
-            return bundle.getString(SPLUNK_UUID_MANIFEST_KEY);
+        if (bundle == null) {
+            Log.d(SplunkRum.LOG_TAG, "Application MetaData bundle is null - no metadata present");
```
SenNeonoveNoci, line 88: nit: You have `MetaData` here, but few lines below you have `metadata`. I would unify to `metadata`.

SenNeonoveNoci (approved):
I have left one "nit" comment, appart from that it looks good.

aditi-s3 (approved):
LGTM

PR #1207, opened 2025-04-04
A quick and dirty code example to show the approach

surbhiia:


integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
```diff
@@ -10,7 +10,17 @@ class GlobalAttributes constructor(
     private var attributes: Attributes = Attributes.empty()
 ) {
     companion object {
-        val instance: GlobalAttributes by lazy { GlobalAttributes() }
+        private var instance: GlobalAttributes? = null
+
+        internal fun initialize(attributes: Attributes){
```
surbhiia, line 15: `Singleton` instance of `GlobalAttributes` can only be `initialized` in this `internal` scoped method, so only from within the same module. Encapsulating the initialization logic and preventing customer from initializing the same. We will initialize it during Agent initialization in `SplunkRumAgentCore`.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/GlobalAttributes.kt
```diff
@@ -10,7 +10,17 @@ class GlobalAttributes constructor(
...
+            instance = GlobalAttributes()
+            instance?.setAll(attributes)
+        }
+
+        fun getInitializedInstance() : GlobalAttributes? {
```
surbhiia, line 20: A `public` method to obtain the `instance`. It will return `null` before agent initialized, and the actual instance after agent initialization

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/internal/SplunkRumAgentCore.kt
```diff
@@ -91,10 +92,13 @@ internal object SplunkRumAgentCore {
...
 
-        agentConfiguration.globalAttributes.getAll().forEach { key, value ->
+        /*agentConfiguration.globalAttributes.getAll().forEach { key, value ->
             @Suppress("UNCHECKED_CAST")
             GlobalAttributes.instance[key as AttributeKey<Any>] = value
-        }
+        }*/
+
+
+        GlobalAttributes.initialize(agentConfiguration.globalAttributes)
```
surbhiia, line 101: This is where the `singleton` `GlobalAttributes` instance is initialized.

app/src/main/java/com/splunk/app/App.kt
```diff
@@ -43,6 +47,13 @@ class App : Application() {
 //        globalAttributes["email"] = "john.doe@example.com"
 //        globalAttributes["isValid"] = true
 
+        val ga = SplunkRum.instance.globalAttributes //or GlobalAttributes.getInitializedInstance()
+        if(ga == null){
+            Log.d("Surbhi","ga is null") //it will be null here
+        } else {
+            Log.d("Surbhi","ga is not null")
+        }
+
```
surbhiia, line 56: Before agent initialization, customer gets a `null` `GlobalAttributes` `instance`. 
Screenshot 2025-04-03 at 9 00 55 PM

app/src/main/java/com/splunk/app/App.kt
```diff
@@ -52,7 +63,7 @@ class App : Application() {
                 ),
                 appName = "Android demo app",
                 enableDebugLogging = true,
-//                globalAttributes = globalAttributes
+                globalAttributes = Attributes.of(AttributeKey.stringKey("xyz"), "xyz")
```
surbhiia, line 66: Customer can pass in an `Attributes` types parameter, which is populated in the `GlobalAttributes` `singleton` after it is created in `SplunkRumAgentCore`.

app/src/main/java/com/splunk/app/App.kt
```diff
@@ -66,6 +77,13 @@ class App : Application() {
             )
         )
 
+        val ga2 = GlobalAttributes.getInitializedInstance()
+        if(ga2 == null){
+            Log.d("Surbhi","ga2 is null")
+        } else {
+            Log.d("Surbhi","ga2 is not null") //it will be not null here
+        }
+
```
surbhiia, line 86: After agent initialization, customer can obtain the `GlobalAttributes` `singleton` `instance` and use the get/set/remove APIs to update global attributes. 
Screenshot 2025-04-03 at 9 01 00 PM

PR #1208, opened 2025-04-04
Consolidate build id naming in feature/next-gen branch

tonzhan2:
Mirrors this commit in the main branch:
https://github.com/signalfx/splunk-otel-android/commit/f54871a5c6a43c2b298bd3487554355bf6a34cdf

Renamed all methods, variables and values that pertain to the optional unique identifier field that gets set by the user in the AndroidManifest.xml file and included in crash/error span attributes to splunk build id. Now the field that the user sets, as well as the attribute tag that gets sent to the backend are: splunk.build_id

Also updated the method that retrieves this build ID to handle non string values more gracefully and better debug logging instead of error logging

### DIFF
--- a/common/utils/src/main/java/com/splunk/sdk/utils/ApplicationInfoUtils.kt
+++ b/common/utils/src/main/java/com/splunk/sdk/utils/ApplicationInfoUtils.kt
@@ -56,14 +56,14 @@ class ApplicationInfoUtils {
                 return null
             }
 
-            val metaData = applicationInfo.metaData
-            if (metaData == null) {
+            val metadata = applicationInfo.metaData
+            if (metadata == null) {
                 Logger.d(TAG, "Application metadata bundle is null - no metadata present")
                 return null
             }
 
-            if (metaData.containsKey(SPLUNK_BUILD_ID)) {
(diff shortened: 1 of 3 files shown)


common/utils/src/main/java/com/splunk/sdk/utils/ApplicationInfoUtils.kt
```diff
@@ -56,11 +56,24 @@ class ApplicationInfoUtils {
                 return null
             }
 
-            return applicationInfo.metaData?.getString(SPLUNK_UUID_MANIFEST_KEY)?.takeIf {
-                it.isNotEmpty()
-            } ?: run {
-                Logger.e(TAG, "Application MetaData bundle is null or does not contain the UUID")
-                null
+            val metaData = applicationInfo.metaData
```
SenNeonoveNoci, line 59: Let’s choose a consistent naming convention for metadata. There’s still a mix of `metadata` and `metaData` being used.

SenNeonoveNoci (changes requested):

SenNeonoveNoci (approved):
lgtm

surbhiia (approved):

PR #1214, opened 2025-04-11
Fix duplicate crash + ANR span bug

tonzhan2:
**When a crash occurs in the application, the SDK was generating two almost identical spans with the same exception information:**

1) One span with basic attributes
2) Another span that also includes our custom attributes including component="crash", service.application_id, and service.version_code

**The problem was caused by having two separate crash reporters being registered:**

1) The default crash reporter from the version of OpenTelemetryRumBuilder we are using, which is enabled by default in the OtelRumConfig
2) Our custom crash reporter installed in the installCrashReporter method with the CrashComponentExtractor and other custom attributes appended

We do not want this one:
Screenshot 2025-04-11 at 11 40 49 AM

We only want this one:
https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java#L316

To resolve this, we must explicitly disable the default crash reporter by adding config.disableCrashReporting() in the RumInitializer while keeping our custom crash reporter implementation. This prevents the duplicate registration and only keeps the crash reporter the splunk sdk installs with the additional attributes we require

Testing:
Tested with sample app and confirm that crashes now generate a single span with all required attributes
The span correctly includes our custom component="crash" attribute and service.version_code etc 

Did the exact same with ANR which had the same issue

### DIFF
--- a/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
+++ b/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java
@@ -108,7 +108,7 @@ class RumInitializer {
         // We are installing the crash reporter ourselves in installCrashReporter.
         // Disabling this one to prevent duplicate spans
         config.disableCrashReporting();
-        config.disableAnrDetection();  // Same with ANR
+        config.disableAnrDetection(); // Same with ANR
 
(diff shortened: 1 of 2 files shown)


SenNeonoveNoci (approved):
lgtm

surbhiia (approved):

PR #1219, opened 2025-04-16
Add attribute extractors for server-timing header

surbhiia:
This PR covers the following:

- Adds an additional attributes extractor for `HttpURLConnection` and `OkHttp3` each to add:
-- `component = http`
-- `link.spanId` and `link.traceId` attributes if `server-timing` (not case sensitive) header is present in `response` (last valid header wins if multiple valid server-timing headers are present) 
- Adds tests for these in sample app.
- Cleanup `OkHttpFragment.kt` code. 
- Upgrade to latest version of `opentelemetry-android` i.e 0.11.0-alpha


common/utils/src/main/java/com/splunk/sdk/utils/ServerTimingHeaderParser.kt
tonzhan2, line 1: It would be good if we could get some unit tests for this utils function. Just sanity checks on valid and invalid inputs

tonzhan2 (approved):
Otherwise, this looks good to me

common/utils/src/main/java/com/splunk/sdk/utils/ServerTimingHeaderParser.kt
surbhiia: Sure! We do have all these covered via sample app for now. We can further determine if we should add unit tests as well. Merging this PR as it is not a blocker right now. I'll bring it up for discussion.

tonzhan2: sounds good

PR #1222, opened 2025-04-19
Integrating Otel Android ANR and Crash Instrumentations and map Crash and ANR legacy APIs

tonzhan2:
Integrated both otel crash and ANR modules. 

**Changes:**
- Added the dependencies to Dependencies.kt and the respective integration/crash or anr `build.gradle`
- Installing the instrumentations in crashIntegration and anrIntegration if config flag is on
- Added crash and anr attribute extractors to the crash and anr instrumentations so that the `component` attribute is added
- Added config blocks for crash and anr to test app setup
- In log exporter, check if log is a crash, and if so, force flush so the span gets exported before the app dies
- Implemented mapping of legacy disableCrashReporting() and disableANRReporting()

**Testing:**
- Tested by logging out spans + log records, and triggering crashes and ANRs in sample app and verifying
- Also verified no crashes and ANRs picked up when config flags set to false
- Tried both with legacy crash and ANR api and legacy splunkrum builder


integration/anr/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ dependencies {
     implementation(project(":integration:agent:internal"))
 
     implementation(Dependencies.Otel.androidInstrumentation)
```
surbhiia, line 30: nit: This dependency can now be removed as it is already part of the new dependency that you've added.

integration/anr/src/main/java/com/splunk/rum/integration/anr/AnrIntegration.kt
```diff
@@ -43,21 +42,30 @@ internal object AnrIntegration {
 
     fun attach(context: Context) {
         Logger.d(TAG, "attach()")
-        AgentIntegration.obtainInstance(context).listeners += installationListener
+        val integration = AgentIntegration.obtainInstance(context)
+        integration.moduleConfigurationManager.listeners += configManagerListener
```
surbhiia, line 46: Question - In other integration modules, we add configManagerListener in the onInstall method. Why did we move this here?

tonzhan2: Moved the configuration listener registration to the attach method because I discovered it was being registered too late in the initialization process. 

In the onInstall method, the configuration had already been processed, causing crash reporting to ignore the disabled setting if I set it to disabled in the test app. It would default to the default enabled setting.

By registering earlier, we ensure the module correctly captures and applies all configuration changes, in this case it was for disabling crash reporting.

Could be worth double checking for your modules too to see if thats an issue

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
```diff
@@ -0,0 +1,39 @@
+package com.splunk.rum.integration.agent.api.attributes
+
+import android.app.Application
+import com.splunk.sdk.common.otel.internal.RumConstants
+import com.splunk.sdk.utils.ApplicationInfoUtils
+import io.opentelemetry.context.Context
+import io.opentelemetry.sdk.logs.LogRecordProcessor
+import io.opentelemetry.sdk.logs.ReadWriteLogRecord
+import kotlin.math.log
```
SenNeonoveNoci, line 9: Please remove.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    private var applicationId: String? = null
+    private var versionCode: String? = null
+    private var splunkBuildId: String? = null
+
+    init {
+        applicationId = ApplicationInfoUtils.retrieveApplicationId(application)
+        versionCode = ApplicationInfoUtils.retrieveVersionCode(application)
+        splunkBuildId = ApplicationInfoUtils.retrieveSplunkBuildID(application)
+    }
```
SenNeonoveNoci, line 21: Nit: The `init` might be redundant:

```kotlin
private var applicationId: String? = ApplicationInfoUtils.retrieveApplicationId(application)
private var versionCode: String? = ApplicationInfoUtils.retrieveVersionCode(application)
private var splunkBuildId: String? = ApplicationInfoUtils.retrieveSplunkBuildID(application)
```

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
```diff
@@ -0,0 +1,39 @@
...
+                logRecord.setAttribute(RumConstants.APPLICATION_ID_KEY, it)
+            }
+            versionCode?.let {
+                logRecord.setAttribute(RumConstants.APP_VERSION_CODE_KEY, it)
+            }
+            splunkBuildId?.let {
+                logRecord.setAttribute(RumConstants.SPLUNK_BUILD_ID, it)
+            }
+        }
+        if (logRecord.instrumentationScopeInfo.name == "io.opentelemetry.crash") {
```
SenNeonoveNoci, line 35: The `"io.opentelemetry.crash"` should be a constant.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
```diff
@@ -0,0 +1,39 @@
...
+            logRecord.setAttribute(RumConstants.COMPONENT_KEY, "crash")
```
SenNeonoveNoci, line 36: The `"crash"` should be a constant.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesSpanProcessor.kt
```diff
@@ -49,6 +48,9 @@ internal class ErrorIdentifierAttributesSpanProcessor(application: Application)
                 span.setAttribute(RumConstants.SPLUNK_BUILD_ID, it)
             }
         }
+        if (span.instrumentationScopeInfo.name == "io.opentelemetry.anr") {
```
SenNeonoveNoci, line 51: The `"io.opentelemetry.anr"` should be a constant.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesSpanProcessor.kt
```diff
@@ -49,6 +48,9 @@ internal class ErrorIdentifierAttributesSpanProcessor(application: Application)
...
+            span.setAttribute(RumConstants.COMPONENT_KEY, "error")
```
SenNeonoveNoci, line 52: The `"error"` should be a constant.

integration/anr/src/main/java/com/splunk/rum/integration/anr/AnrIntegration.kt
```diff
@@ -43,21 +42,30 @@ internal object AnrIntegration {
...
+        integration.listeners += installationListener
     }
 
     private val configManagerListener = object : ModuleConfigurationManager.Listener {
         override fun onSetup(configurations: List<ModuleConfiguration>) {
+            moduleConfiguration = configurations.find<AnrModuleConfiguration>() ?: defaultModuleConfiguration
+            Logger.d(TAG, "onSetup(moduleConfiguration: ${moduleConfiguration})")
```
SenNeonoveNoci, line 53: Nit: Can be simplified to:
```kotlin
Logger.d(TAG, "onSetup(moduleConfiguration: $moduleConfiguration)")
```

integration/crash/src/main/java/com/splunk/rum/integration/crash/CrashIntegration.kt
```diff
@@ -40,21 +42,30 @@ internal object CrashIntegration {
...
-        AgentIntegration.obtainInstance(context).listeners += installationListener
+        val integration = AgentIntegration.obtainInstance(context)
+        integration.moduleConfigurationManager.listeners += configManagerListener
+        integration.listeners += installationListener
     }
 
     private val configManagerListener = object : ModuleConfigurationManager.Listener {
         override fun onSetup(configurations: List<ModuleConfiguration>) {
+            moduleConfiguration = configurations.find<CrashModuleConfiguration>() ?: defaultModuleConfiguration
+            Logger.d(TAG, "onSetup(moduleConfiguration: ${moduleConfiguration})")
```
SenNeonoveNoci, line 53: Nit: Can be simplified: `$moduleConfiguration`.

SenNeonoveNoci (dismissed):

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
```diff
@@ -0,0 +1,32 @@
...
+                logRecord.setAttribute(RumConstants.APPLICATION_ID_KEY, it)
+            }
+            versionCode?.let {
+                logRecord.setAttribute(RumConstants.APP_VERSION_CODE_KEY, it)
+            }
+            splunkBuildId?.let {
+                logRecord.setAttribute(RumConstants.SPLUNK_BUILD_ID, it)
+            }
+        }
+        if (logRecord.instrumentationScopeInfo.name == RumConstants.CRASH_INSTRUMENTATION_SCOPE_NAME) {
```
surbhiia, line 28: For crashes, we want to set component = "crash" for the first thread that crashes and component = "error" for subsequent threads in a crash involving multiple threads. 

Earlier [this](https://github.com/signalfx/splunk-otel-android/blob/57737e9f8a58fb257dc8947800e2265a715aed04/splunk-otel-android/src/main/java/com/splunk/rum/CrashComponentExtractor.java#L27) extractor was used to do so, we might need something similar.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesSpanProcessor.kt
```diff
@@ -49,6 +42,9 @@ internal class ErrorIdentifierAttributesSpanProcessor(application: Application)
                 span.setAttribute(RumConstants.SPLUNK_BUILD_ID, it)
             }
         }
+        if (span.instrumentationScopeInfo.name == RumConstants.ANR_INSTRUMENTATION_SCOPE_NAME) {
```
surbhiia, line 45: We should instead add a separate attribute extractor to the `ANR instrumentation` for setting `component` instead of using the common `ErrorIdentifierAttributesSpanProcessor` which is meant for adding de-obfuscation details required by server for telemetry with `exception.stacktrace` attribute.

tonzhan2: Done.

I was originally bit more inclined to keep that logic here, and maybe rename the span processor to something a bit more appropriate to prevent creating another entire class just for this which felt to me a little bit overkill. But since we added added an attributes extractor to crash instrumentation, I just mirrored it here and had it always set the component to "error"

common/otel/src/main/java/com/splunk/sdk/common/otel/logRecord/AndroidLogRecordExporter.kt
```diff
@@ -68,6 +68,10 @@ internal class AndroidLogRecordExporter : LogRecordExporter {
                 }
             } finally {
                 span.end(log.timestampEpochNanos, TimeUnit.NANOSECONDS)
+                if (log.instrumentationScopeInfo.name == RumConstants.CRASH_INSTRUMENTATION_SCOPE_NAME) {
+                   SplunkOpenTelemetrySdk.instance!!.sdkTracerProvider.forceFlush()
```
TomasChladekSL, line 72: Please do not use `!!`.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
```diff
@@ -0,0 +1,30 @@
+package com.splunk.rum.integration.agent.api.attributes
+
+import android.app.Application
+import com.splunk.sdk.common.otel.internal.RumConstants
+import com.splunk.sdk.utils.ApplicationInfoUtils
+import io.opentelemetry.context.Context
+import io.opentelemetry.sdk.logs.LogRecordProcessor
+import io.opentelemetry.sdk.logs.ReadWriteLogRecord
+
+internal class ErrorIdentifierAttributesLogProcessor(application: Application) : LogRecordProcessor {
```
TomasChladekSL, line 10: This looks more like application identification than error identification. Please consider renaming the class.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    private var applicationId: String? = ApplicationInfoUtils.retrieveApplicationId(application)
+    private var versionCode: String? = ApplicationInfoUtils.retrieveVersionCode(application)
+    private var splunkBuildId: String? = ApplicationInfoUtils.retrieveSplunkBuildID(application)
+
+    override fun onEmit(context: Context, logRecord: ReadWriteLogRecord) {
+        if (logRecord.getAttribute(RumConstants.COMPONENT_KEY) == RumConstants.COMPONENT_ERROR ||
+            logRecord.getAttribute(RumConstants.COMPONENT_KEY) == RumConstants.COMPONENT_CRASH) {
+            applicationId?.let {
```
TomasChladekSL, line 19: The null check is redundant, you can pass nullable value.

TomasChladekSL (changes requested):

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
tonzhan2: This class has been removed

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/attributes/ErrorIdentifierAttributesLogProcessor.kt
tonzhan2: These fields are only added to spans with component "crash" or "error" and are used to identify errors that the backend needs to potentially deobfuscate. Thats why it is named accordingly. But this is a duplicate and will be removed

surbhiia (approved):

PR #1223, opened 2025-04-19
Implemented mappings for legacy globalAttribute APIs

tonzhan2:
Implemented mappings for legacy global attribute APIs.

Tested with sample app (updates to sample app for testing not added to this PR)

### DIFF
--- a/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
+++ b/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
@@ -85,9 +85,6 @@ class SplunkRum private constructor(
     /**
      * Update the global set of attributes that will be appended to every span and event.
      *
-     * <p>Note: this operation performs an atomic update. The passed function should be free from
-     * side effects, since it may be called multiple times in case of thread contention.
-     *
      * @param attributesUpdater A function which will update the current set of attributes, by
      *     operating on a {@link AttributesBuilder} from the current set.
      */
--- a/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
+++ b/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
@@ -66,9 +66,6 @@ class SplunkRum private constructor(
      *
      * <p>If you attempt to set a value to null or use a null key, this call will be ignored.
      *
-     * <p>Note: this operation performs an atomic update. The passed function should be free from
-     * side effects, since it may be called multiple times in case of thread contention.
-     *
      * @param key The {@link AttributeKey} for the attribute.
      * @param value The value of the attribute, which must match the generic type of the key.
      * @param <T> The generic type of the value.
@@ -84,6 +81,9 @@ class SplunkRum private constructor(
 
     /**
      * Update the global set of attributes that will be appended to every span and event.
+     *
+     * <p>Note: this operation performs an atomic update. The passed function should be free from
+     * side effects, since it may be called multiple times in case of thread contention.
      *
      * @param attributesUpdater A function which will update the current set of attributes, by
      *     operating on a {@link AttributesBuilder} from the current set.
--- a/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
+++ b/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
@@ -60,10 +60,8 @@ class SplunkRum private constructor(
 
     @Deprecated("Use globalAttributes property")
     fun <T> setGlobalAttribute(key: AttributeKey<T>, value: T) {
-        if (value != null) {
-            @Suppress("UNCHECKED_CAST")
-            globalAttributes.set(key as AttributeKey<Any>, value as Any)
-        }
+        @Suppress("UNCHECKED_CAST")
+        value?.let { globalAttributes[key as AttributeKey<Any>] = it as Any }
     }
 
     @Deprecated("Use globalAttributes property")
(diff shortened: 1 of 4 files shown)


integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -60,12 +60,18 @@ class SplunkRum private constructor(
 
     @Deprecated("Use globalAttributes property")
     fun <T> setGlobalAttribute(key: AttributeKey<T>, value: T) {
-        // TODO separate task
+        if (value != null) {
```
SenNeonoveNoci, line 63: Nit: Consider using:
```kotlin
@Suppress("UNCHECKED_CAST")
value?.let { globalAttributes[key as AttributeKey<Any>] = it as Any }
```

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -60,12 +60,18 @@ class SplunkRum private constructor(
...
+        }
     }
 
     @Deprecated("Use globalAttributes property")
     fun updateGlobalAttributes(attributesUpdater: Consumer<AttributesBuilder>) {
-        // TODO separate task
+        val builder = Attributes.builder()
+        attributesUpdater.accept(builder)
+        globalAttributes.removeAll()
+        globalAttributes.setAll(builder.build())
```
SenNeonoveNoci, line 74: Can this be replaced by the following?
```kotlin
globalAttributes.update { attributesUpdater.accept(this) }
```

SenNeonoveNoci (changes requested):
I have one nitpick and one request for change.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
tonzhan2: I dont think so.

globalAttributes.update { attributesUpdater.accept(this) }

does not remove the already existing attributes, it just adds the new ones.

The original splunk agent update call is supposed to clear the existing attributes before adding whatever new ones are passed in. To replicate that behavior here we have to do it like this.

Screenshot 2025-04-21 at 12 27 29 PM

SenNeonoveNoci: > The original splunk agent update call is supposed to clear the existing attributes before adding whatever new ones are passed in.

I don’t think that’s the case. The API documentation states the following:

```
* @param attributesUpdater A function which will update the current set of attributes, by
* operating on a {@link AttributesBuilder} from the current set.
```

And in the legacy code, it’s implemented like this:

```java
public void update(Consumer<AttributesBuilder> attributesUpdater) {
    AttributesBuilder builder = attributes.toBuilder();
    attributesUpdater.accept(builder);
    attributes = builder.build();
}
```

So, the `AttributesBuilder` that the client uses to update should be populated with the current attributes in both the legacy and next-gen APIs.

surbhiia:
Can we add some good comments to these APIs, l[ike how it was there in the legacy agent](https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/SplunkRum.java#L280-L293), as these are public APIs, it's helpful for the customers. May be also the replaceWith argument in the @Deprecated annotation like [this](https://github.com/signalfx/splunk-otel-android/blob/feature/next-gen/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt#L56).

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
tonzhan2: Ah, my mistake. Thanks for clarifying! Just updated it

tonzhan2:
> Can we add some good comments to these APIs, l[ike how it was there in the legacy agent](https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/SplunkRum.java#L280-L293), as these are public APIs, it's helpful for the customers. May be also the replaceWith argument in the Deprecated annotation like [this](https://github.com/signalfx/splunk-otel-android/blob/feature/next-gen/integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt#L56).

I added the replaceWith argument in the annotation. 

But since it's an old API thats being deprecated, I dont think we need to add comments to it since:

- Existing customers already understand what these methods do
- The ReplaceWith annotation clearly guides users to the new API
- The new MutableAttributes class already has comments and documentation
- It's unnecessary maintenance overhead for code that will eventually be removed

SenNeonoveNoci (approved):
Thanks for the changes.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -58,6 +58,21 @@ class SplunkRum private constructor(
...
+    /**
+     * Set an attribute in the global attributes that will be appended to every span and event.
+     *
+     * <p>Note: If this key is the same as an existing key in the global attributes, it will replace
+     * the existing value.
+     *
+     * <p>If you attempt to set a value to null or use a null key, this call will be ignored.
+     *
+     * <p>Note: this operation performs an atomic update. The passed function should be free from
+     * side effects, since it may be called multiple times in case of thread contention.
```
surbhiia: This might not be correct for this method.

surbhiia (approved):

PR #1226, opened 2025-04-23
Web View Integration 

surbhiia:
This PR provides an API to integrate Android Rum with Browser Rum thereby providing a way to access Android RUM session Id from Browser RUM. 

Following was added in the PR:
- New API
- Legacy API
- Sample App tests for both


integration/webview/src/main/java/com/splunk/rum/integration/webview/extension/SplunkRumExt.kt
```diff
@@ -0,0 +1,40 @@
...
+ * This method will enable Splunk Browser-based RUM to integrate with the current Android RUM
+ * Session. It injects a javascript object named "SplunkRumNative" into your WebView which
+ * exposes the Android Session ID to the browser-based RUM javascript implementation.
+ *
+ * @param webView The WebView to inject the javascript object into.
+ */
+@Deprecated("Use SplunkRum.instance.webViewNativeBridge.integrateWithBrowserRum(webView)", ReplaceWith("SplunkRum.instance.webViewNativeBridge.integrateWithBrowserRum(webView)"))
+fun SplunkRum.integrateWithBrowserRum(webView: WebView) {
+    webViewNativeBridge.integrateWithBrowserRum(webView)
+}
```
SenNeonoveNoci, line 40: This will not work in Java—or more specifically, it cannot be called as before.

The legacy API could be invoked from Java like this:
```java
splunkRum.integrateWithBrowserRum(myWebView);
```

However, if you test it with the `SplunkRum.integrateWithBrowserRum(webView: WebView)` Kotlin extension and try to call it from Java, it will not be found:

image

This happens because Kotlin extension functions are not visible to Java in the same way—they don't work as expected in a Java context.

So for compatibility and API mapping, we need to retain the following in `SplunkRum`:
```
@Deprecated("Use webView.integrateWithBrowserRum()")
fun integrateWithBrowserRum(webView: WebView) {
    // TODO separate task
}
```

This allows the function to remain callable from Java the same way as in legacy API.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -68,11 +68,6 @@ class SplunkRum private constructor(
         // TODO separate task
     }
 
-    @Deprecated("Use webView.integrateWithBrowserRum(webView)")
-    fun integrateWithBrowserRum(webView: WebView) {
-        // TODO separate task
-    }
-
```
SenNeonoveNoci, line 75: This cannot be removed. See the latest comment in the review for details.

SenNeonoveNoci (changes requested):

integration/webview/src/main/java/com/splunk/rum/integration/webview/WebViewNativeBridge.kt
```diff
@@ -0,0 +1,42 @@
...
+    fun integrateWithBrowserRum(webView: WebView) {
+        webView.addJavascriptInterface(NativeRumSessionId(), "SplunkRumNative")
+    }
+
+    companion object {
+
+        /**
+         * The instance of [WebViewNativeBridge].
+         */
+        @JvmStatic
```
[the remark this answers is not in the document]
surbhiia: Converted to `object` for now. Can be called from both `kotlin` and `Java` directly instead of needing an `instance`. Keeping inline with `navigation` module. We will update this as needed based on our other pending decisions.

surbhiia:
For the unresolved comments, I'm maintaining parity with `navigation` module for now until we decide the path forward and will update as needed.

SenNeonoveNoci (approved):
Thank you for addressing the comments.

tonzhan2 (approved):

PR #1231, opened 2025-04-24
Make java compatible

surbhiia:
A non-complete quick example. Will update/optimize if we agree with higher level approach


integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/session/SplunkSessionManager.kt
```diff
@@ -152,8 +152,15 @@ class SplunkSessionManager internal constructor(
         }
     }
 
-    private companion object {
+    companion object {
         const val DEFAULT_SESSION_TIMEOUT = 15L * 60L * 1000L // 15m
```
TomasChladekSL, line 156: Constants should remain private.

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/session/SplunkSessionManager.kt
```diff
@@ -152,8 +152,15 @@ class SplunkSessionManager internal constructor(
...
         const val DEFAULT_SESSION_LENGTH = 4L * 60L * 60L * 1000L // 4h
+
+        var instance: SplunkSessionManager? = null
```
TomasChladekSL, line 159: You should not expose `instance` as property that can be changed.

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/session/SplunkSessionManager.kt
```diff
@@ -152,8 +152,15 @@ class SplunkSessionManager internal constructor(
...
+
+        fun initialize(agentStorage: IAgentStorage): SplunkSessionManager {
+            instance = SplunkSessionManager(agentStorage)
```
TomasChladekSL, line 162: You should not override `instance` every time the `initialize` is called.

TomasChladekSL (changes requested):

PR #1233, opened 2025-04-28
Session state, api adjustments

OndrejSynakSL:


integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -25,6 +25,10 @@ import com.splunk.rum.integration.agent.api.attributes.MutableAttributes
 import com.splunk.rum.integration.agent.api.internal.SplunkRumAgentCore
 import com.splunk.rum.integration.agent.api.subprocess.SubprocessDetector
 import com.splunk.rum.integration.agent.api.user.User
+import com.splunk.rum.integration.agent.internal.AgentIntegration
+import com.splunk.rum.integration.agent.internal.session.ISplunkSessionManager
+import com.splunk.rum.integration.agent.internal.session.NoOpSplunkSessionManager
+import com.splunk.rum.integration.agent.internal.session.SplunkSessionManager
```
SenNeonoveNoci, line 31: Nit: Not used.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -101,11 +107,13 @@ class SplunkRum private constructor(
     }
 
     companion object {
-        private val noop = SplunkRum(
+        @Deprecated("Instance is available if install is not called prior to fetching instance.")
+        val noop = SplunkRum(
```
SenNeonoveNoci, line 111: Why is this public? Since we are discontinuing the API, it should be enough to have:

```kotlin
@JvmStatic
@Deprecated("Use SplunkRum.instance without calling SplunkRum.install() to get noop instance")
fun noop(): SplunkRum {
    return noop
}
```

SenNeonoveNoci (changes requested):
Added 2 comments.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -171,13 +181,16 @@ class SplunkRum private constructor(
...
 
         @JvmStatic
-        @Deprecated("Use SplunkRum.instance.state.status == Status.Running", ReplaceWith("instance.state.status == Status.Running", "com.splunk.rum.integration.agent.api.SplunkRum.Companion.instance"))
+        @Deprecated(
+            "Use SplunkRum.instance.state.status == Status.Running",
+            ReplaceWith("instance.state.status == Status.Running", "com.splunk.rum.integration.agent.api.SplunkRum.Companion.instance")
+        )
         fun isInitialized(): Boolean {
             return instance.state.status == Status.Running
         }
```
surbhiia, line 191: Let's retain the comments documentation for all these deprecated APIs as well - https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/SplunkRum.java#L166

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/AgentConfiguration.kt
```diff
@@ -35,15 +35,14 @@ import io.opentelemetry.sdk.trace.data.SpanData
...
+    val appName: String,
+    val deploymentEnvironment: String,
+    val appVersion: String? = null,
     val enableDebugLogging: Boolean = false,
-    val sessionSamplingRate: Double = 1.0, // TODO move to session
     val globalAttributes: MutableAttributes = MutableAttributes(),
     val spanInterceptor: ((SpanData) -> SpanData?)? = null,
     val user: UserConfiguration = UserConfiguration(),
-    val session: Any? = null, // TODO,
+    val session: SessionConfiguration = SessionConfiguration(),
```
surbhiia, line 45: `sessionConfiguration` as the name of the `argument` would be better than using `session`

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/ISession.kt
```diff
@@ -16,15 +16,19 @@
 
 package com.splunk.rum.integration.agent.api
 
+import com.splunk.rum.integration.agent.internal.session.ISplunkSessionManager
+
 interface ISession {
     val state: State
```
surbhiia, line 22: There's a `session.state` which is this, and then there's agent state - The following class in this PR (`integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/IState.kt`)

Might be worth renaming the two - `SessionState`, `AgentState` or at least the `IState` class to `IAgentState`

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRumBuilder.kt
```diff
@@ -159,7 +159,7 @@ class SplunkRumBuilder {
                 appName = appName ?: throw IllegalStateException("applicationName was not set"),
                 deploymentEnvironment = deploymentEnvironment ?: throw IllegalStateException("deploymentEnvironment was not set"),
                 enableDebugLogging = enableDebug,
-                sessionSamplingRate = sessionBasedSampling,
+                session = SessionConfiguration(sessionBasedSampling),
```
surbhiia, line 162: Rename variable to `sessionConfiguration` here as well.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -42,9 +44,10 @@ import java.util.function.Consumer
 class SplunkRum private constructor(
     agentConfiguration: AgentConfiguration,
     userManager: IUserManager,
+    sessionManager: ISplunkSessionManager,
     val openTelemetry: OpenTelemetry,
     val state: IState = State(agentConfiguration),
-    val session: ISession = Session(ISession.State())
+    val session: ISession = Session(ISession.State(agentConfiguration.session, sessionManager))
 ) {
     val user: User = User(userManager)
```
surbhiia, line 53: Question: Is it worth moving `user` and `globalAttributes` `val` to the above private constructor as well, similar to `state` (agentState) and `session` present in there already? 

Both are immutable (`val`) and require simple initialization based on received constructor arguments. Keeping in constructor would help keep all such initializations/dependencies in the constructor itself offering better readability.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/AgentConfiguration.kt
OndrejSynakSL: it follows the same logic like user, endpoint and so on according to API document

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
OndrejSynakSL: Not scope of this PR but I will update it

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRumBuilder.kt
OndrejSynakSL: as mentioned above

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/session/SplunkSessionManager.kt
```diff
@@ -48,17 +64,17 @@ class SplunkSessionManager internal constructor(
...
         private set(value) = agentStorage.writePreviousSessionId(value)
 
-    var sessionTimeout: Long = DEFAULT_SESSION_TIMEOUT
+    override val sessionListeners: MutableSet<SessionListener> = HashSet()
 
-    var maxSessionLength: Long = DEFAULT_SESSION_LENGTH
+    private var sessionTimeout: Long = DEFAULT_SESSION_TIMEOUT
 
-    val sessionListeners: MutableSet<SessionListener> = HashSet()
+    private var maxSessionLength: Long = DEFAULT_SESSION_LENGTH
```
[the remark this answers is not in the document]
OndrejSynakSL: Not scope of this PR, but I think maxSessionLength is applied no matter the foreground vs background. sessionTimeout will be renamed to sessionBackgroundTimeout

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
surbhiia: How is it not in scope when you're altering the deprecated API annotation of the same in this PR?

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/session/SplunkSessionManager.kt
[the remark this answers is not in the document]
surbhiia: Same here. How is it not in scope of this PR when you're adjusting this class and the very same variables we're talking about.

SenNeonoveNoci (approved):
Thanks for resolving my comments.

surbhiia (approved):

PR #1236, opened 2025-04-28
Remove all log processors

SenNeonoveNoci:
Since the agent only emits spans, and logs are currently translated into spans, the log processors are redundant and have been removed.


surbhiia (approved):
LGTM!!

PR #1240, opened 2025-04-30
Rename auto http interception modules & artifacts to contain word auto

surbhiia:
This PR covers:
- Rename auto http interception modules & artifacts to contain word auto
- Remove a ToDo from `AgentIntegration.kt` as discussed in today's sync.


settings.gradle
```diff
@@ -14,29 +14,29 @@ include ':common:otel',
 
 // Instrumentation modules
 include ':instrumentation:runtime:startup',
-        ':instrumentation:buildtime:httpurlconnection:plugin',
-        ':instrumentation:buildtime:okhttp3:plugin'
+        ':instrumentation:buildtime:httpurlconnection-auto:plugin',
```
tonzhan2, line 17: Certainly won't block this from merging. But none of these other modules have a '-' in the name.
Would it make more sense to have these modules named such that these include statements are called like:         `':instrumentation:buildtime:httpurlconnectionauto:plugin' `instead of
`':instrumentation:buildtime:httpurlconnection-auto:plugin'`?

tonzhan2 (approved):

settings.gradle
surbhiia: @tonzhan2 We already discussed and decided in team chat before. Please bring it up there again if needed.

SenNeonoveNoci (approved):
lgtm

PR #1241, opened 2025-04-30
Refactor all span processors in a common internal package

tonzhan2:
- Create a processor package within the internal module of the agent
- Move all span processors to this common package
- Update usages of processor imports to this new package


SenNeonoveNoci (approved):
lgtm

PR #1245, opened 2025-05-07
Integrate Slow Rendering Instrumentation and mapped legacy APIs

tonzhan2:
- Added slow rendering integration module with new module config setup style
- Added module config with isEnabled and interval flag, both of which are used to determine how the instrumentation is installed 
- Added slow rendering integration module to agent api buildscript and settings.gradle
- Added legacy API mappings in SplunkRumBuilder for disableSlowRenderingDetection and setSlowRenderingDetectionPollInterval
- Added these module configs (new and legacy) to sdk setup in app.kt class

Tested with both new and legacy API:
- Triggered slow and frozen frames and verified spans are generated
- Interval can be set, and invalid intervals (eg negative) will be denied and default to 1second default
- No spans generated when disabled


integration/slowrendering/src/main/java/com/splunk/rum/integration/slowrendering/SlowRenderingModuleConfiguration.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package com.splunk.rum.integration.slowrendering
+
+import com.splunk.rum.integration.agent.module.ModuleConfiguration
+import java.time.Duration
+
+class SlowRenderingModuleConfiguration(
+    val isEnabled: Boolean = true,
+    val interval: Duration = Duration.ofSeconds(1)) : ModuleConfiguration {
```
SenNeonoveNoci, line 24: I’m not sure if `interval` is descriptive enough. This can probably be clarified with a Javadoc. I checked, and other module configurations are also missing Javadocs. I’ll take care of that, but in the meantime, please add documentation for this one. Thanks!

integration/slowrendering/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+    api(platform(Dependencies.Otel.androidBom))
+
+    implementation(project(":integration:agent:internal"))
+
+    implementation(Dependencies.Otel.instrumentationApi)
+    implementation(Dependencies.Otel.androidSlowRenderingInstrumentation)
+
+    implementation(Dependencies.SessionReplay.commonLogger)
+    implementation(Dependencies.SessionReplay.commonUtils)
+    implementation(project(":common:otel"))
```
SenNeonoveNoci, line 35: nit: I would keep the `project` dependencies together.

SenNeonoveNoci (changes requested):

integration/slowrendering/src/main/java/com/splunk/rum/integration/slowrendering/SlowRenderingModuleConfiguration.kt
tonzhan2: Added some docs to the top of this. Let me know if it looks good to you.

Should I also add to the legacy slow rendering module config class?

SenNeonoveNoci: > Added some docs to the top of this. Let me know if it looks good to you.

I think like this, its good. Thanks.

> Should I also add to the legacy slow rendering module config class?

I wouldn’t do that, because it’s not meant to be used by clients—it’s just a means for us to bridge the legacy API. So from my perspective, it’s not necessary.

SenNeonoveNoci (approved):
Thanks for the changes.

PR #1247, opened 2025-05-12
Added documentation

TomasChladekSL:
Added documentation


SenNeonoveNoci (approved):
lgtm

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRumBuilder.kt
```diff
@@ -122,12 +183,22 @@ class SplunkRumBuilder {
...
         crashReportingEnabled = false
         return this
     }
 
+    /**
+     * Disables the ANR reporting feature.
+     *
+     * This feature is enabled by default. You can disable it by calling this method.
+     */
     @Deprecated("ANRReporting is now controlled by the ANRModuleConfiguration")
```
surbhiia, line 202: May be add what it returns as well:
@return `this`

surbhiia:
May be add what API returns and params it takes as well, the documentation right now is not consistent with these, some mention some don't: 
@return
@param

TomasChladekSL:
> May be add what API returns and params it takes as well, the documentation right now is not consistent with these, some mention some don't: @return @param

I know, but I think the code is self-explanatory in many places. For example, there's a builder, and all functions return the instance itself except for build. This is a design pattern, and it should work consistently across all the code. Or consider parameter named `attributes` in function named `setGlobalAttributes`. I don't think there's much room for error if you simply look at the names.

But if needed, we can add any additional documentation later.

PR #1256, opened 2025-05-14
Add tests for legacy custom tracking APIs + some necessary cleanup

surbhiia:
- Moved Custom Tracking tests to its own screen. Will cleanup remaining in upcoming task for decluttering menu fragment. 
- Added tests for legacy custom tracking APIs + some necessary cleanup

This is how the custom tracking screen looks:
[
Screenshot 2025-05-14 at 2 44 35 PM
](url)

### DIFF
--- a/app/src/main/java/com/splunk/app/ui/customtracking/CustomTrackingFragment.kt
+++ b/app/src/main/java/com/splunk/app/ui/customtracking/CustomTrackingFragment.kt
@@ -116,9 +116,9 @@ class CustomTrackingFragment : BaseFragment<FragmentCustomTrackingBinding>() {
         }
 
         workflowSpan?.apply {
-            setAttribute("workflow.start.time", System.currentTimeMillis())
-            Thread.sleep(125) // Simulate processing time
-            setAttribute("workflow.end.time", System.currentTimeMillis())
+            val startTime = System.currentTimeMillis()
+            setAttribute("workflow.start.time", startTime)
+            setAttribute("workflow.end.time", startTime + 125) // Added 125ms processing time in end time.
             end()
         }
     }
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/splunk/app/ui/customtracking/CustomTrackingFragment.kt
```diff
@@ -0,0 +1,140 @@
...
+    private fun simulateWorkflow(implementationType: ApiVariant) {
+        val workflowSpan = when (implementationType) {
+            ApiVariant.NEXTGEN -> SplunkRum.instance.customTracking.trackWorkflow("Test Workflow")
+            ApiVariant.LEGACY -> SplunkRum.instance.startWorkflow("Test Workflow")
+        }
+
+        workflowSpan?.apply {
+            setAttribute("workflow.start.time", System.currentTimeMillis())
+            Thread.sleep(125) // Simulate processing time
+            setAttribute("workflow.end.time", System.currentTimeMillis())
```
tonzhan2, line 121: Question: can we just set end.time attribute with the previous system time + say 1000ms? That way we only need to get the current time once and can avoid putting the thread to sleep. And the span that gets sent can have a larger time interval

tonzhan2 (approved):

app/src/main/java/com/splunk/app/ui/customtracking/CustomTrackingFragment.kt
surbhiia: Yeah, That would be better! I will update. Thanks for the suggestion :)

OndrejSynakSL (approved):

PR #1260, opened 2025-05-21
Add network attributes to all spans

surbhiia:
Adds network attributes to all spans

### DIFF
--- a/integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/SplunkInternalGlobalAttributeSpanProcessor.kt
+++ b/integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/SplunkInternalGlobalAttributeSpanProcessor.kt
@@ -54,5 +54,16 @@ class SplunkInternalGlobalAttributeSpanProcessor : SpanProcessor {
 
     companion object {
         val attributes: MutableList<Attribute> = ArrayList()
+
+        fun updateStringAttribute(key: String, value: String?) {
+            attributes.removeIf { it.name == key }
+            if (value != null) {
+                attributes += Attribute.String(
+                    key, value
+                )
+            }
+        }
     }
+
+
 }
--- a/integration/networkmonitor/src/main/java/com/splunk/rum/integration/networkmonitor/NetworkMonitorModuleIntegration.kt
+++ b/integration/networkmonitor/src/main/java/com/splunk/rum/integration/networkmonitor/NetworkMonitorModuleIntegration.kt
@@ -20,10 +20,22 @@ import android.content.Context
 import com.cisco.android.common.logger.Logger
 import com.splunk.rum.integration.agent.internal.legacy.LegacyNetworkMonitorModuleConfiguration
 import com.splunk.rum.integration.agent.internal.module.ModuleIntegration
+import com.splunk.rum.integration.agent.internal.processor.SplunkInternalGlobalAttributeSpanProcessor
 import com.splunk.rum.integration.agent.module.ModuleConfiguration
 import com.splunk.rum.integration.agent.module.extension.find
+import io.opentelemetry.android.common.internal.features.networkattributes.data.CurrentNetwork
+import io.opentelemetry.android.common.internal.features.networkattributes.data.NetworkState
 import io.opentelemetry.android.instrumentation.InstallationContext
 import io.opentelemetry.android.instrumentation.network.NetworkChangeInstrumentation
+import io.opentelemetry.android.internal.services.Services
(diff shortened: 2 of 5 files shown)


integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/SplunkInternalGlobalAttributeSpanProcessor.kt
```diff
@@ -54,5 +54,12 @@ class SplunkInternalGlobalAttributeSpanProcessor : SpanProcessor {
 
     companion object {
         val attributes: MutableList<Attribute> = ArrayList()
+
+        fun updateStringAttribute(key: String, value: String?) {
```
SenNeonoveNoci, line 58: Nit: This might be useful in multiple places and use cases. Consider extracting it into a utility method.

Idea:
```koltin
fun MutableList<Attribute>.updateStringAttribute(key: String, value: String) {
    this.removeIf { it.name == key }
    this += Attribute.String(key, value)
}
```

SenNeonoveNoci: Nit: Also, so you don't have to always "convert" between `key: String` and `key: AttributeKey<String>`, it can be beneficial to have this conversion resolved by defining two functions. For example:

```
fun MutableList<Attribute>.updateStringAttribute(key: AttributeKey<String>, value: String) = 
    updateStringAttribute(key.key, value)

fun MutableList<Attribute>.updateStringAttribute(key: String, value: String) {
    this.removeIf { it.name == key }
    this += Attribute.String(key, value)
}
```

integration/networkmonitor/src/main/java/com/splunk/rum/integration/networkmonitor/NetworkMonitorModuleIntegration.kt
```diff
@@ -36,9 +48,37 @@ internal object NetworkMonitorModuleIntegration : ModuleIntegration<NetworkMonit
...
+
+                    SplunkInternalGlobalAttributeSpanProcessor.updateStringAttribute(NETWORK_CARRIER_NAME.key, currentNetwork.getCarrierName())
+
+                    SplunkInternalGlobalAttributeSpanProcessor.updateStringAttribute(NETWORK_CARRIER_MCC.key, currentNetwork.getCarrierCountryCode())
+
+                    SplunkInternalGlobalAttributeSpanProcessor.updateStringAttribute(NETWORK_CARRIER_MNC.key, currentNetwork.getCarrierNetworkCode())
+
+                    SplunkInternalGlobalAttributeSpanProcessor.updateStringAttribute(NETWORK_CARRIER_ICC.key, currentNetwork.getCarrierIsoCountryCode())
+                }
+            }
```
SenNeonoveNoci, line 73: This code does not seem idiomatic to Kotlin, lets try to fix this.

```kotlin
val listener = NetworkChangeListener { currentNetwork ->
    with(SplunkInternalGlobalAttributeSpanProcessor) {
        updateStringAttribute(NETWORK_CONNECTION_TYPE, currentNetwork.state.humanName)
        updateStringAttribute(NETWORK_CONNECTION_SUBTYPE, currentNetwork.subType)
        updateStringAttribute(NETWORK_CARRIER_NAME, currentNetwork.carrierName)
        updateStringAttribute(NETWORK_CARRIER_MCC, currentNetwork.carrierCountryCode)
        updateStringAttribute(NETWORK_CARRIER_MNC, currentNetwork.carrierNetworkCode)
        updateStringAttribute(NETWORK_CARRIER_ICC, currentNetwork.carrierIsoCountryCode)
    }
}
```

My suggested improvements:
- Please convert the anonymous object to a lambda expression to simplify the code.
- Replace all `getX()` method calls with Kotlin property access syntax (e.g., use `state.humanName` instead of `getState().getHumanName()`).
- Use a `with` block around `SplunkInternalGlobalAttributeSpanProcessor` to avoid repeating the object name multiple times.
- Remove explicit type declarations when Kotlin can infer the type automatically for cleaner code.

SenNeonoveNoci: Nit: Also, I don’t need to use `.key` because the above proposed function already has `key: AttributeKey<String>` in its signature.

SenNeonoveNoci (changes requested):
I have a small request for some changes. Also, let’s finish the debate on whether `MutableAttributes` can be used on Slack.

PR #1269, opened 2025-05-27
AndroidSpanExporter buffer from memory to storage

keciman:


common/otel/src/main/java/com/splunk/sdk/common/otel/span/AndroidSpanExporter.kt
```diff
@@ -48,43 +47,49 @@ internal class AndroidSpanExporter(
...
-    override fun export(spans: MutableCollection<SpanData>): CompletableResultCode = if (deferredUntilForeground &&
-        !isForeground
-    ) {
-        buffer.addAll(spans)
-        CompletableResultCode.ofSuccess()
-    } else {
-        flushBufferedSpans(spans)
-        CompletableResultCode.ofSuccess()
+    override fun export(spans: MutableCollection<SpanData>): CompletableResultCode {
+        if (spans.isEmpty()) return CompletableResultCode.ofSuccess()
```
OndrejSynakSL, line 51: Just a nitpick in general throughout this PR. You are saving list of spans and not one span so it is not spanID but spansID

PR #1283, opened 2025-06-06
Unify API classes (User, Session)

SenNeonoveNoci:
Unified the API classes for user and session from `Session.XyZ` to `Session.Xyz` and `User.XyZ` to `User.Xyz`.


app/src/main/java/com/splunk/app/App.kt
```diff
@@ -77,7 +77,7 @@ class App : Application() {
                     realm = realm,
                     rumAccessToken = rumAccessToken
                 ),
-                appName = "Android demo app",
+                appName = "StabilityTest",
```
keciman, line 80: is this intended ?

SenNeonoveNoci: Good catch, I have changed this back.

keciman (approved):

OndrejSynakSL (approved):

PR #1285, opened 2025-06-06
Update repos for sonatype central

breedx-splk:
See this annoucement: https://central.sonatype.org/news/20250326_ossrh_sunset/

and see this for the docs on this change https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-central

Once a new token is generated and put into GitLab releaser repo, we can merge this to get the new urls into place.


aditi-s3 (approved):
LGTM

PR #1291, opened 2025-06-13
NO-TICKET: Rename all package names to com.splunk.rum

SenNeonoveNoci:
Renamed all the package names to `com.splunk.rum`.


common/otel/src/main/java/com/splunk/rum/common/otel/OpenTelemetryInitializer.kt
```diff
@@ -54,7 +53,7 @@ class OpenTelemetryInitializer(
 
         resource = Resource.getDefault()
 
-        val spanExporter = SpanInterceptorExporter(
+        val spanExporter = com.splunk.rum.common.otel.span.SpanInterceptorExporter(
```
keciman, line 56: seems like wrong fix

OndrejSynakSL (approved):

PR #1294, opened 2025-06-13
Okhttp manual instrumentation

surbhiia:
This PR covers:

- Manual okhttp module
- A `OkHttpManualInstrumentation` class to host the public API so it is callable from java. And kotlin extension for same. 
- Renamed classes in auto okhttp module with `auto` word similar to manual okhttp module
- Added `capturedRequestHeaders` and `capturedResponseHeaders` config in all 3 Http instrumentation modules
- Tests for manual okhttp API latest and legacy - right now code needs to be uncommented if you're not using auto okhttp and want to test manual okhttp. Later, as part of the existing app update ticket, I'll introduce a dropdown or some other nice way of altering between the two manual options as needed in the app itself. 
- Extract the common attribute extractor in :okhttp:common module

What is not covered:
- Removal of the other manual okhttp instrumentation. Would like to leave it to the developer of same to remove it.


SenNeonoveNoci (changes requested):
Please remove the `LICENSE.txt` file. The other comments are optional (nits), but feel free to address them if you’d like.

PR #1296, opened 2025-06-16
DEMRUM-2439: Add event.name attribute where missing

tonzhan2:
`event.name` attribute was missing in spans converted from logs, because with otel logs:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-attributes

"eventName", not "event.name" identifies the class / type of event.

### DIFF
--- a/common/otel/src/main/java/com/splunk/rum/common/otel/logRecord/AndroidLogRecordExporter.kt
+++ b/common/otel/src/main/java/com/splunk/rum/common/otel/logRecord/AndroidLogRecordExporter.kt
@@ -40,7 +40,7 @@ internal class AndroidLogRecordExporter : LogRecordExporter {
 
             // traceId and spanId should be inside the context already from global OTel instance
             val spanBuilder = SplunkOpenTelemetrySdk.instance!!.sdkTracerProvider.get(RumConstants.RUM_TRACER_NAME)
-                .spanBuilder(log.attributes[AttributeKey.stringKey("event.name")] ?: "")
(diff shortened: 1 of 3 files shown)


common/otel/src/main/java/com/splunk/rum/common/otel/logRecord/AndroidLogRecordExporter.kt
```diff
@@ -40,7 +40,7 @@ internal class AndroidLogRecordExporter : LogRecordExporter {
 
             // traceId and spanId should be inside the context already from global OTel instance
             val spanBuilder = SplunkOpenTelemetrySdk.instance!!.sdkTracerProvider.get(RumConstants.RUM_TRACER_NAME)
-                .spanBuilder(log.attributes[AttributeKey.stringKey("event.name")] ?: "")
+                .spanBuilder(log.attributes[AttributeKey.stringKey("eventName")] ?: "")
```
OndrejSynakSL, line 43: I see that there is still some usage of event.name in the project. Would this change break the current code?

SenNeonoveNoci (changes requested):
Hi @tonzhan2, sorry for the late response (we were out yesterday). I believe this is not a robust solution for all cases since it won’t work for instrumentation implemented directly in the agent that uses the `event.name` attribute.

We can unify this throughout later, but for now, the solution is hopefully in this PR: https://github.com/signalfx/splunk-otel-android/pull/1302

Please take a look.

PR #1298, opened 2025-06-18
DEMRUM-2356: Add required fingerprint to enable stacktrace symbolication

tonzhan2:
The backend needs either one of the following conditions to be met:
- otel.status_code tag has value ERROR
- error tag exists
- http.response.status_code tag has a value greater or equal than 400

in order to enable symbolication.

In this PR, `error` = `true` attribute has been added to all ANR and crash spans, as well as manually reported errors


OndrejSynakSL (approved):

surbhiia:
Left a comment in slack to discuss if we should instead use `otel.status_code` as that is part of semantic conventions instead of `error` which is not. Also, checked the legacy crash span which also sets `otel.status_code = ERROR`whereas `error = ""`

PR #1300, opened 2025-06-18
NO-TICKET: Add default screen.name

SenNeonoveNoci:
Having a value for the `screen.name` attribute is **mandatory** for metrics to be derived on the platform. Using `RumConstants.DEFAULT_SCREEN_NAME` ensures that a `screen.name` is always present.


surbhiia:
Question: Have we evaluated if the derived metrics in the backend make sense in case when navigation instrumentation is not enabled and all spans contain the same default screen.name? 

Secondly, Based on how that evaluation goes, we should call this out if needed in our documentation - the impact navigation turned off vs on has on the metrics.

aditi-s3:
@surbhiia Great question! I did check this and it does not. We do not aggregate based on screen name but there are places we segment on it and that would be inaccurate anyway if screen name is not correctly populated.

PR #1302, opened 2025-06-18
NO-TICKET: Log2span - fix event name detection

SenNeonoveNoci:
Adds logic to set span names using:
	1.	`eventName` from `ExtendedLogRecordData`
	2.	`"event.name"` attribute
	3.	Defaults to "log" if neither is present


OndrejSynakSL (approved):

surbhiia:
Question: Why is `event.name` at all used? Is it or was it part of the log semantic conventions? Why can't all instrumentations stick to setting `eventName` as per semantic conventions. 

Have we evaluated the possibility of deciphering the span name from `instrumentationScopeInfo.name` attribute in case `eventName` is missing?

PR #1313, opened 2025-06-25
(WIP) DEMRUM-2067: Mapping file gradle plugin POC

tonzhan2:
Work in progress.

Gradle plugin that:
- automatically generates a `splunk build id` for all app build variants with minification enabled
- injects that id into the app's final AndroidManifest.XML file so the id is accessible by the sdk at runtime
- and then uploads the generated mapping files to splunk o11y cloud

To test:

- add access token and realm variables to `uploadMappingFileAfterBuild ` method in `MappingFilePlugin.kt`
- build sdk and publish to maven local
- uncomment `id("com.splunk.rum-mapping-file-plugin") version "2.0.0-beta"` in `app/build.gradle.kts`
- build app and verify successful build id generation and mapping file upload in logs

To do:

- split up code into separate files, clean up
- optimizations highlighted in `TODO` comments

### DIFF
--- a/instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
+++ b/instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
@@ -1,0 +1,93 @@
+import utils.artifactPrefix
+import utils.defaultGroupId
+
+plugins {
+    id("java-gradle-plugin")
+    kotlin("jvm")
+    id("maven-publish")
+    id("signing")
+}
+
+group = defaultGroupId
+version = Configurations.sdkVersionName
+
+val javadocJar by tasks.creating(Jar::class) {
+    archiveClassifier.set("javadoc")
+    from(tasks.named("javadoc"))
+    metaInf {
+        from("$projectDir/src/main/assets/LICENSE.txt")
+    }
+}
+
+val sourcesJar by tasks.creating(Jar::class) {
+    archiveClassifier.set("sources")
+    from(sourceSets["main"].allSource)
+    metaInf {
+        from("$projectDir/src/main/assets/LICENSE.txt")
+    }
+}
+
+tasks.jar {
+    manifest {
+        attributes(
+            "Implementation-Version" to Dependencies.Otel.otelAndroidBomVersion
+        )
+    }
+    metaInf {
+        from("$projectDir/src/main/assets/LICENSE.txt")
+    }
+}
+
+gradlePlugin {
+    plugins {
+        create("androidInstrumentationPlugin") {
+            id = "$defaultGroupId.${artifactPrefix}mapping-file-plugin"
+            implementationClass = "com.splunk.rum.mappingfile.plugin.MappingFilePlugin"
+            displayName = "Splunk Android Mapping File Plugin"
+        }
+    }
+}
+
+dependencies {
+    implementation(gradleApi())
+    implementation(Dependencies.gradle)
+}
+
+signing {
+    val secretKey: String? = System.getenv("SECRET_KEY")
+    val signingPassword: String? = System.getenv("SIGNING_PASSWORD")
+
+    if (secretKey != null && signingPassword != null) {
+        useInMemoryPgpKeys(secretKey, signingPassword)
+        sign(publishing.publications)
+    } else {
+        println("WARNING: Environment variables SECRET_KEY and/or SIGNING_PASSWORD not set. Skipping signing of artifacts.")
+    }
+}
+
+publishing {
+    publications {
+        withType(MavenPublication::class.java) {
+            pom.withXml { asNode().addCiscoInfo() }
+
+            artifactId = "${artifactPrefix}mapping-file-plugin"
+
+            artifact(javadocJar)
+            artifact(sourcesJar)
+        }
+        repositories {
+            maven {
+                name = "maven"
+                url = uri(Configurations.Artifactory.bareRepositoryURL)
+                credentials {
+                    username = System.getenv("ARTIFACT_REPO_USERNAME")
+                    password = System.getenv("ARTIFACT_REPO_PASSWORD")
+                }
+            }
+            maven {
+                name = "local"
+                url = uri("$projectDir/repo")
+            }
+        }
+    }
+}
--- a/settings.gradle
+++ b/settings.gradle
@@ -3,7 +3,8 @@ pluginManagement {
     repositories {
         gradlePluginPortal()
         //uncomment to look for plugins in local maven (.m2) folder
-        //mavenLocal()
+        mavenLocal()
+        google()
     }
 }
 
@@ -15,7 +16,8 @@ include ':common:otel',
 // Instrumentation modules
 include ':instrumentation:runtime:startup',
         ':instrumentation:buildtime:httpurlconnection-auto:plugin',
-        ':instrumentation:buildtime:okhttp3-auto:plugin'
+        ':instrumentation:buildtime:okhttp3-auto:plugin',
+        ':instrumentation:buildtime:mapping-file:plugin'
 
 // Integration
 include ':integration:agent:api',
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -8,6 +8,8 @@ plugins {
     // Uncomment this to test HttpURLConnection instrumentation
     //id("com.splunk.android.rum-okhttp3-auto-plugin") version "24.4.1"
     //id("com.splunk.android.rum-httpurlconnection-auto-plugin") version "24.4.1"
+    // Uncomment this to test mapping file plugin
+    //id("com.splunk.rum-mapping-file-plugin") version "2.0.0-beta"
 }
 
 apply<ConfigAndroidApp>()
--- a/instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/MappingFilePlugin.kt
+++ b/instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/MappingFilePlugin.kt
@@ -1,0 +1,285 @@
+/*
+ * Copyright 2024 Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.rum.mappingfile.plugin
+
+import com.android.build.gradle.AppExtension
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import java.io.File
+import java.util.*
+
+class MappingFilePlugin : Plugin<Project> {
+
+    private lateinit var project: Project
+
+    override fun apply(project: Project) {
+        this.project = project
+        // TODO: replace all println with more proper logging
+        println("Splunk RUM: Mapping File Plugin applied!")
+
+        project.afterEvaluate {
+            setupBuildIdInjection()
+        }
+    }
+
+    private fun setupBuildIdInjection() {
+        val android = project.extensions.findByType(AppExtension::class.java)
+        if (android == null) {
+            println("Splunk RUM: Android application plugin not found, skipping setup")
+            return
+        }
+
+        android.applicationVariants.configureEach { variant ->
+            if (variant.buildType.isMinifyEnabled) {
+                injectBuildIdForVariant(variant)
+            } else {
+                println("Splunk RUM: Skipping variant '${variant.name}' as minification not enabled")
+            }
+        }
+    }
+
+    private fun injectBuildIdForVariant(variant: com.android.build.gradle.api.ApplicationVariant) {
+        // TODO: check that customer does not have existing manual setup, if so, skip this process and log notice
+
+        val buildId = UUID.randomUUID().toString()
+        println("Splunk RUM: Generated build ID for variant '${variant.name}': $buildId")
+
+        injectViaManifestModification(variant, buildId)
+        setupUploadTask(variant, buildId)
+    }
+
+    private fun setupUploadTask(
+        variant: com.android.build.gradle.api.ApplicationVariant,
+        buildId: String
+    ) {
+        // Hook into the assemble task to upload after build completes
+        val assembleTaskName = "assemble${variant.name.capitalize(Locale.ROOT)}"
+        project.tasks.named(assembleTaskName).configure { assembleTask ->
+            assembleTask.doLast {
+                uploadMappingFileAfterBuild(variant, buildId)
+            }
+        }
+    }
+
+    private fun uploadMappingFileAfterBuild(
+        variant: com.android.build.gradle.api.ApplicationVariant,
+        buildId: String
+    ) {
+        // TODO: create gradle plugin extension for customer to put in these values instead
+        val accessToken = project.findProperty("splunk.accessToken") as String?
+            ?: System.getenv("SPLUNK_ACCESS_TOKEN")
+            ?: ""
+        val realm = project.findProperty("splunk.realm") as String?
+            ?: System.getenv("SPLUNK_REALM")
+            ?: "lab0"
+
+        // TODO: potentially recursively search for mapping file in case it isnt in this location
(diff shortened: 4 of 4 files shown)


instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/BuildIdInjector.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    private fun injectMetadataIntoMergedManifest(
+        variant: com.android.build.gradle.api.ApplicationVariant,
+        buildId: String
+    ) {
+        println("Splunk RUM: Searching for manifest files")
+
+        val buildDir = project.layout.buildDirectory.get().asFile
+        val manifestFiles = buildDir.walkTopDown()
+            .filter { it.name == "AndroidManifest.xml" && it.absolutePath.contains(variant.name) }
```
TomasChladekSL, line 54: Maybe `variant.name in it.name`?

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/BuildIdInjector.kt
```diff
@@ -0,0 +1,108 @@
...
+            println("Splunk RUM: No manifest files found for variant ${variant.name}")
+        }
+    }
+
+    private fun addMetadataToManifest(manifestFile: File, buildId: String): Boolean {
+        try {
+            var content = manifestFile.readText()
+
+            // Find <application> tag and inject metadata after it
+            val applicationPattern = Regex("<application([^>]*)>")
```
TomasChladekSL, line 88: Regex isn't safe here, it could cause the build to fail. Please consider using the built-in XML parser instead.

[https://regex101.com/r/K959s1/1](https://regex101.com/r/K959s1/1)

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/MappingFileUploader.kt
```diff
@@ -0,0 +1,75 @@
...
+package com.splunk.rum.mappingfile.plugin.utils
+
+import java.io.File
+import org.gradle.api.Project
+
+class MappingFileUploader(private val project: Project) {
+
+    private val httpClient = SplunkHttpClient()
+
+    fun uploadAfterBuild(variant: com.android.build.gradle.api.ApplicationVariant, buildId: String) {
```
TomasChladekSL, line 26: Class import

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/MappingFileUploader.kt
```diff
@@ -0,0 +1,75 @@
...
+        val buildDir = project.layout.buildDirectory.get().asFile
+        val mappingFile = File(buildDir, "outputs/mapping/${variant.name}/mapping.txt")
+
+        if (!mappingFile.exists()) {
+            println("Splunk RUM: Mapping file not found at ${mappingFile.absolutePath}, skipping upload")
+            return
+        }
+
+        if (accessToken.isBlank() || realm.isBlank()) {
+            val missingConfig = when {
```
TomasChladekSL, line 45: Could you please handle a case when both properties are missing?

settings.gradle
```diff
@@ -42,4 +44,4 @@ include ':integration:agent:api',
 include ':app'
```
TomasChladekSL, line 44: Could you please make a sample integration in the 'app'? Upload logic will be disabled by an user-accessible property.

TomasChladekSL (changes requested):

instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+    metaInf {
+        from("$projectDir/src/main/assets/LICENSE.txt")
+    }
+}
+
+tasks.jar {
+    manifest {
+        attributes(
+            "Implementation-Version" to Dependencies.Otel.otelAndroidBomVersion
+        )
```
surbhiia, line 34: This was needed in the other plugins as we add a library dependency with this version at compile time to host project. Why is this needed here?

instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+        sign(publishing.publications)
+    } else {
+        println("WARNING: Environment variables SECRET_KEY and/or SIGNING_PASSWORD not set. Skipping signing of artifacts.")
+    }
+}
+
+publishing {
+    publications {
+        withType(MavenPublication::class.java) {
+            pom.withXml { asNode().addCiscoInfo() }
```
surbhiia, line 71: use addSplunkInfo() here

instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+        }
+        repositories {
+            maven {
+                name = "maven"
+                url = uri(Configurations.Artifactory.bareRepositoryURL)
+                credentials {
+                    username = System.getenv("ARTIFACT_REPO_USERNAME")
+                    password = System.getenv("ARTIFACT_REPO_PASSWORD")
+                }
+            }
```
surbhiia, line 86: Needs to change to the new build configs we're using now.

settings.gradle
```diff
@@ -3,7 +3,8 @@ pluginManagement {
     repositories {
         gradlePluginPortal()
         //uncomment to look for plugins in local maven (.m2) folder
-        //mavenLocal()
+        mavenLocal()
+        google()
```
surbhiia, line 7: Comment mavenLocal() and why do we need to add google() here? Is the plugin not available via  gradlePluginPortal()?

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/SplunkHttpClient.kt
```diff
@@ -0,0 +1,117 @@
...
+ * limitations under the License.
+ */
+
+package com.splunk.rum.mappingfile.plugin.utils
+
+import java.io.File
+import java.net.HttpURLConnection
+import java.net.URL
+
+class SplunkHttpClient {
```
surbhiia, line 23: Class name doesn't seem apt. This class seems to be a mix of general Http methods + upload related methods.

instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
tonzhan2: Yes, this is simply a POC that was made before those changes. The official PR will be made off the latest feature/next-gen

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/BuildIdInjector.kt
tonzhan2: I dont think that works. Using variant.name instead of it.name wouldn't work because all Android manifest files are named AndroidManifest.xml regardless of variant. The variant info is in the directory path, not the filename. And using it.name instead of variant name wouldn't help us check for the correct path
For example:

build/intermediates/merged_manifests/release/AndroidManifest.xml
build/intermediates/merged_manifests/debug/AndroidManifest.xml

Both have it.name = "AndroidManifest.xml", so we need to check the path to distinguish between variants.

Unless Im misunderstanding what you're saying

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/BuildIdInjector.kt
```diff
@@ -0,0 +1,178 @@
...
+
+    private fun injectMetadataIntoMergedManifest(variant: ApplicationVariant, buildId: String) {
+        variant.outputs.forEach { output ->
+            try {
+                val processManifestTaskProvider = output.processManifestProvider
+                val processManifestTask = processManifestTaskProvider.get()
+                val outputFiles = processManifestTask.outputs.files.files
+
+                println("Splunk RUM: processManifest task outputs for variant '${variant.name}':")
+                println("Splunk RUM: Found ${outputFiles.size} output files:")
```
SenNeonoveNoci, line 70: Please use the `project.logger` instead.

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/BuildIdInjector.kt
```diff
@@ -0,0 +1,178 @@
...
+                outputFiles.forEach { file ->
+                    println("  - ${file.absolutePath} (exists: ${file.exists()}, isFile: ${file.isFile})")
+
+                    // If it's a directory, look for AndroidManifest.xml inside it
+                    if (file.isDirectory) {
+                        val manifestInDir = File(file, "AndroidManifest.xml")
+                        println(
+                            "    → Looking for AndroidManifest.xml: ${manifestInDir.absolutePath} (exists: ${manifestInDir.exists()})"
+                        )
```
SenNeonoveNoci, line 79: Please use the `project.logger` instead.

instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+
+signing {
+    val signingKey: String? = project.findProperty("signingKey") as String?
+    val signingPassword: String? = project.findProperty("signingPassword") as String?
+
+    if (signingKey != null && signingPassword != null) {
+        useInMemoryPgpKeys(signingKey, signingPassword)
+        sign(publishing.publications)
+    } else {
+        println("WARNING: Environment variables signingKey and/or signingPassword not set. Skipping signing of artifacts.")
```
SenNeonoveNoci, line 70: Please use the `project.logger` instead.

instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+
+group = defaultGroupId
+version = Configurations.sdkVersionName
+
+val javadocJar by tasks.creating(Jar::class) {
+    archiveClassifier.set("javadoc")
+    from(tasks.named("javadoc"))
+    metaInf {
+        from("$projectDir/src/main/assets/LICENSE.txt")
+    }
```
SenNeonoveNoci, line 21: This license file inclusion is probably leftover logic from AppD, and I believe it should not be here for the Splunk Agents. Also, I think there is no `LICENSE.txt` in `"$projectDir/src/main/assets/LICENSE.txt"`. If my assumptions are true, please remove it.

instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+val sourcesJar by tasks.creating(Jar::class) {
+    archiveClassifier.set("sources")
+    from(sourceSets["main"].allSource)
+    metaInf {
+        from("$projectDir/src/main/assets/LICENSE.txt")
+    }
```
SenNeonoveNoci, line 29: Not needed, license file (see above comment).

instrumentation/buildtime/mapping-file/plugin/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+
+tasks.jar {
+    manifest {
+        attributes(
+            "Implementation-Version" to Dependencies.Otel.otelAndroidBomVersion
+        )
+    }
+    metaInf {
+        from("$projectDir/src/main/assets/LICENSE.txt")
+    }
```
SenNeonoveNoci, line 40: Not needed, license file (see above comment).

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/SplunkRumExtension.kt
```diff
@@ -0,0 +1,53 @@
...
+
+open class SplunkRumExtension @Inject constructor(objectFactory: ObjectFactory) {
+
+    /**
+     * Enable or disable the mapping file plugin functionality. Will generate an unique Splunk Build ID for each variant
+     * to accurately correlate stacktraces with mapping files. The gradle plugin will also inject it into the intermediate
+     * AndroidManifest.xml files of your app during the build process, and also uploaded with each mapping file
+     * Default: true
+     */
+
```
SenNeonoveNoci, line 31: Nit: Remove the space.

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/MappingFileUploadClient.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    private fun sendMultipartData(connection: HttpURLConnection, mappingFile: File) {
+        logger.info("Splunk RUM: Sending multipart data")
+        val boundary = connection.getRequestProperty("X-Boundary")
+
+        connection.outputStream.use { outputStream ->
+            val writer = java.io.PrintWriter(java.io.OutputStreamWriter(outputStream, "UTF-8"), true)
+
+            // Add filename parameter
+            writer.append("--$boundary").append("\r\n")
```
SenNeonoveNoci, line 87: Nit: Might be a good candidate for a constant: 

```kotlin
const val CRLF = "\r\n"
```

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/MappingFileUploadClient.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private fun handleResponse(connection: HttpURLConnection) {
+        logger.debug("Splunk RUM: Reading HTTP response")
+
+        val responseCode = connection.responseCode
+        logger.info("Splunk RUM: Response code: $responseCode")
+        logger.info("Splunk RUM: Response message: ${connection.responseMessage}")
+
+        if (responseCode in 200..299) {
```
SenNeonoveNoci, line 117: Nit: Good candidate for ext function to make this super clear:

```kotlin
fun Int.isSuccessfulHttpCode(): Boolean {
    return this in 200..299
}
```

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/MappingFileUploader.kt
```diff
@@ -0,0 +1,156 @@
...
+            // Don't fail the build, let app build finish
+        }
+    }
+
+    private fun resolveAccessToken(extension: SplunkRumExtension): String {
+        project.logger.debug("Splunk RUM: Resolving access token")
+        return extension.apiAccessToken.orNull
+            ?: project.findProperty("splunk.accessToken") as String?
+            ?: System.getenv("SPLUNK_ACCESS_TOKEN")
+            ?: ""
```
SenNeonoveNoci, line 114: Nit: It seems to me that returning null and then checking for it is more standard than using an empty string, but that is a matter of preference.

instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/utils/MappingFileUploader.kt
```diff
@@ -0,0 +1,156 @@
...
+            )
+            return
+        }
+
+        project.logger.info(
+            "Splunk RUM: Found mapping file (${mappingFile.length()} bytes) at: ${mappingFile.absolutePath}"
+        )
+
+        project.logger.lifecycle("Splunk RUM: Uploading mapping file for variant ${variant.name}")
+        project.logger.info("Splunk RUM: Upload details:")
```
SenNeonoveNoci, line 84: Nit: Replace with a single log and a multiline string, as you did previously for the logs above.

SenNeonoveNoci (approved):
LGTM

TomasChladekSL (approved):

PR #1328, opened 2025-07-02
Migrate publishing from Artifactory to Nexus Plugin, and add Snapshot publish support to feature/next-gen branch

tonzhan2:


SenNeonoveNoci (approved):
@tonzhan2 Thanks for working on this. It looks good from my perspective, but please note that I’m not an expert on Gradle and publishing. I have left one small nit comment for your consideration.

tonzhan2:
> @tonzhan2 I checked the PR and it all makes sense to me. I am just curious about the release branches which I know Aditi mentioned. Will we test it strictly from next-gen or from the release branches as well ? That is just an idea for future so we are possibly able to test from release branches as well once we have those if needed

So in the `build` job in the gitlab script, you can see that there is a `rules` block which specifies if the commit ref name is `feature/next-gen`. This was changed from `main` in this PR. Basically what this means is that gitlab will only run this build job if it's in the feature/next-gen branch. Functionally this means that this build job, and thus snapshot publishing, will only be run upon merges into the feature/next-gen branch. 

This is what we want by design. Previously this build job, and snapshot publishing, were run only on merges into the main branch. With the unified agent, feature/next-gen is effectively our new main branch.

In terms of testing release, we can always configure a dry run or snapshot job in a pinch in another branch if needed. But this should automate snapshot publishing from the next gen branch. So we can always test the latest snapshot of the feature/next-gen branch before we go ahead with an official release if that makes sense.

PR #1339, opened 2025-07-10
NO-TICKET: Fix last screen name logic

keciman:
This PR solves wrong displaying in the Web player of different screens. Also it aligns more with iOS agent mainly on creating the spans in AppStart

### DIFF
--- a/integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
+++ b/integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
@@ -18,7 +18,6 @@ package com.splunk.rum.integration.agent.internal.processor
 
 import android.os.SystemClock
 import com.splunk.rum.common.otel.SplunkOpenTelemetrySdk
-import com.splunk.rum.common.otel.internal.RumConstants
 import com.splunk.rum.integration.agent.common.module.toSplunkString
 import com.splunk.rum.integration.agent.internal.AgentIntegration.Companion.modules
 import io.opentelemetry.api.trace.Span
@@ -50,7 +49,7 @@ class AppStartSpanProcessor : SpanProcessor {
         val tracer = SplunkOpenTelemetrySdk.instance
             ?.sdkTracerProvider
             ?.get("splunk-app-start")
-                ?: throw IllegalStateException("unable to report initialization")
+            ?: throw IllegalStateException("unable to report initialization")
         val modules = modules.values
 
         val firstInitialization =
@@ -66,7 +65,7 @@ class AppStartSpanProcessor : SpanProcessor {
             .startSpan()
 
         span.setAttribute("component", "appstart")
-            .setAttribute("screen.name","unknown")
+            .setAttribute("screen.name", "unknown")
 
         val resources = modules.joinToString(",", "[", "]") {
             it.configuration?.toSplunkString()
--- a/integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
+++ b/integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
@@ -26,6 +26,7 @@ import io.opentelemetry.context.Context
 import io.opentelemetry.sdk.trace.ReadWriteSpan
 import io.opentelemetry.sdk.trace.ReadableSpan
 import io.opentelemetry.sdk.trace.SpanProcessor
+import java.time.Instant
 import java.util.concurrent.TimeUnit
 
 class AppStartSpanProcessor : SpanProcessor {
@@ -46,8 +47,9 @@ class AppStartSpanProcessor : SpanProcessor {
     override fun isEndRequired(): Boolean = true
 
     private fun reportInitialization(appStartSpan: Span) {
-        val provider =
-            SplunkOpenTelemetrySdk.instance?.sdkTracerProvider
+        val tracer = SplunkOpenTelemetrySdk.instance
+            ?.sdkTracerProvider
+            ?.get("splunk-app-start")
                 ?: throw IllegalStateException("unable to report initialization")
         val modules = modules.values
 
@@ -57,12 +59,14 @@ class AppStartSpanProcessor : SpanProcessor {
         val startTimestamp =
             firstInitialization.startTimestamp + SystemClock.elapsedRealtime() - firstInitialization.startElapsed
 
-        val span = provider.get(RumConstants.RUM_TRACER_NAME)
+        val span = tracer
             .spanBuilder("SplunkRum.initialize")
             .setParent(Context.current().with(appStartSpan))
             .setStartTimestamp(startTimestamp, TimeUnit.MILLISECONDS)
-            .setAttribute("component", "appstart")
             .startSpan()
+
+        span.setAttribute("component", "appstart")
+            .setAttribute("screen.name","unknown")
 
         val resources = modules.joinToString(",", "[", "]") {
             it.configuration?.toSplunkString()
@@ -89,6 +93,6 @@ class AppStartSpanProcessor : SpanProcessor {
             )
         }
 
-        span.end()
+        span.end(Instant.now())
     }
 }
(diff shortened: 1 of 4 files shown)


OndrejSynakSL (approved):

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
```diff
@@ -46,9 +46,10 @@ class AppStartSpanProcessor : SpanProcessor {
     override fun isEndRequired(): Boolean = true
 
     private fun reportInitialization(appStartSpan: Span) {
-        val provider =
-            SplunkOpenTelemetrySdk.instance?.sdkTracerProvider
-                ?: throw IllegalStateException("unable to report initialization")
+        val tracer = SplunkOpenTelemetrySdk.instance
+            ?.sdkTracerProvider
+            ?.get("splunk-app-start")
```
surbhiia, line 51: Why are we setting different instrumentation scope names in this PR instead of using "SplunkRum" which we were using earlier in its place?

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
```diff
@@ -57,13 +58,15 @@ class AppStartSpanProcessor : SpanProcessor {
...
 
-        val span = provider.get(RumConstants.RUM_TRACER_NAME)
+        val span = tracer
             .spanBuilder("SplunkRum.initialize")
             .setParent(Context.current().with(appStartSpan))
             .setStartTimestamp(startTimestamp, TimeUnit.MILLISECONDS)
-            .setAttribute("component", "appstart")
             .startSpan()
 
+        span.setAttribute("component", "appstart")
```
surbhiia, line 67: Please use constants defined in RumConstants for the strings here and define any new needed constants.

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
```diff
@@ -89,6 +92,6 @@ class AppStartSpanProcessor : SpanProcessor {
             )
         }
 
-        span.end()
+        span.end(Instant.now())
```
surbhiia, line 95: Why do we need to explicitly pass Instant.now() here? Span.end should automatically take the current time as the ending time

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
```diff
@@ -46,9 +46,10 @@ class AppStartSpanProcessor : SpanProcessor {
     override fun isEndRequired(): Boolean = true
 
     private fun reportInitialization(appStartSpan: Span) {
-        val provider =
-            SplunkOpenTelemetrySdk.instance?.sdkTracerProvider
-                ?: throw IllegalStateException("unable to report initialization")
+        val tracer = SplunkOpenTelemetrySdk.instance
+            ?.sdkTracerProvider
+            ?.get("splunk-app-start")
+            ?: throw IllegalStateException("unable to report initialization")
```
surbhiia, line 52: The exception message here and all other places doesn't seem correct. Here may be we can use - "Unable to capture app startup span"?

surbhiia (changes requested):

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
keciman: I can put it back to previous state this is the same way it is done on iOS I thought it might be good idea to keep it as similar as we can and also this way it is more descriptive

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
keciman: I do not know OTel specific this much but again this was a try to make it same as on iOS implementation.

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
keciman: Adjusted

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
keciman: I can convert it back I had already this part done in the code because I was trying to see what differs on our side compared to iOS.

integration/agent/internal/src/main/java/com/splunk/rum/integration/agent/internal/processor/AppStartSpanProcessor.kt
```diff
@@ -61,9 +63,11 @@ class AppStartSpanProcessor : SpanProcessor {
             .spanBuilder("SplunkRum.initialize")
             .setParent(Context.current().with(appStartSpan))
             .setStartTimestamp(startTimestamp, TimeUnit.MILLISECONDS)
-            .setAttribute("component", "appstart")
             .startSpan()
 
+        span.setAttribute(RumConstants.COMPONENT_KEY, "appstart")
+            .setAttribute(RumConstants.SCREEN_NAME_KEY, RumConstants.DEFAULT_SCREEN_NAME)
```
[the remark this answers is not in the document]
keciman: From what I understand there should be no navigation event before App Start as mentioned in the thread before on slack. And there will be fix for that in the future. AppStart is meant just for initialization before UI. So yes it does overwrite it but because that logic is already not correct I believe. Again this is taken from iOS agent and I believe it makes sense. This code is quite important for this PR. If really necessary you can open this discussion on slack.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -55,10 +55,18 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
 
             SplunkInternalGlobalAttributeSpanProcessor.attributes[RumConstants.SCREEN_NAME_KEY] = screenName
 
-            provider.get(RumConstants.RUM_TRACER_NAME)
-                .spanBuilder("Created")
-                .setAttribute("component", "ui")
-                .createZeroLengthSpan()
```
surbhiia, line 61: Is it still a zero length span? If not, why so? If yes, Why not use `createZeroLengthSpan()` method?

integration/startup/src/main/java/com/splunk/rum/integration/startup/StartupModuleIntegration.kt
```diff
@@ -83,12 +84,15 @@ internal object StartupModuleIntegration : ModuleIntegration<StartupModuleConfig
...
-            .setAttribute("component", "appstart")
-            .setAttribute("start.type", name)
+            .setStartTimestamp(startTimestamp, TimeUnit.MILLISECONDS)
             .startSpan()
+
+        span
+            .setAttribute(RumConstants.COMPONENT_KEY, "appstart")
+            .setAttribute(RumConstants.SCREEN_NAME_KEY, RumConstants.DEFAULT_SCREEN_NAME)
+            .setAttribute("start.type", name)
             .end(endTimestamp.toInstant())
```
surbhiia, line 96: You're changing the startTimestamp that goes here. You do not need to change anything here but just add one line to add `screen.name = "unknown"` using: 

`.setAttribute(RumConstants.SCREEN_NAME_KEY, RumConstants.DEFAULT_SCREEN_NAME)`

integration/startup/src/main/java/com/splunk/rum/integration/startup/StartupModuleIntegration.kt
```diff
@@ -83,12 +84,15 @@ internal object StartupModuleIntegration : ModuleIntegration<StartupModuleConfig
             return
         }
 
-        provider.get(RumConstants.RUM_TRACER_NAME)
+        val span = provider.get(RumConstants.RUM_TRACER_NAME)
             .spanBuilder("AppStart")
-            .setStartTimestamp(startTimestamp.toInstant())
-            .setAttribute("component", "appstart")
-            .setAttribute("start.type", name)
+            .setStartTimestamp(startTimestamp, TimeUnit.MILLISECONDS)
```
surbhiia, line 89: Don't change timestamp here. Keep it same as before - converting to Epoch instant.

surbhiia (approved):

PR #1354, opened 2025-07-25
DEMRUM-1983: Cleanup style + present okhttp manual options on the UI

surbhiia:
- Style cleanup + remove the extra header showing up in all screens (as also shown in the screenshot below)
- For OkHttp3, Introduce radio buttons that show up only when auto-instrumentation is not enabled, so customers can choose between latest/legacy kotlin API usage for manual instrumentation, earlier they needed to uncomment lines of code. 

This one


tonzhan2 (approved):
lgtm

PR #1357, opened 2025-07-30
DEMRUM-2788: Add BOMs again

surbhiia:
Add BOMs again after correcting any errors. 

We need to ensure that all dependencies directly mentioned in the `build.gradle.kts` of a `module` for which the `version` comes from BOM, have the corresponding platform BOM added also in the same `build.gradle.kts`. Even though things work without it as some other project dependency might be bringing in the relevant BOM, the created POM is not deemed correct and SONATYPE new rule validation fails the POM.


tonzhan2 (approved):

OndrejSynakSL (commented):
Does it have to be api dependency for all the modules?

surbhiia:
> Does it have to be api dependency for all the modules?

Yes, api helps ensure the consuming app also get this consistent versioning propagated to them. Also needed for multi-module projects like ours!

PR #1363, opened 2025-08-13
DEMRUM-2939: Add global logic for propagating lastScreenName as an attribute to Spans

OndrejSynakSL:
Remove local logic and adjust the FragmentTracer, ActivityTracer


SenNeonoveNoci (approved):
LGTM

PR #1377, opened 2025-08-25
NO-TICKET: remove event.name attribute from span

keciman:


common/otel/src/main/java/com/splunk/rum/common/otel/logRecord/AndroidLogRecordExporter.kt
```diff
@@ -101,7 +101,11 @@ internal class AndroidLogRecordExporter(
                     spanBuilder.setAttribute("body", log.bodyValue.toString())
                 }
 
-                log.attributes.asMap().forEach { (key, value) ->
+                log.attributes.asMap().forEach attrs@{ (key, value) ->
+                    if (key.key == "event.name") {
```
surbhiia, line 105: nit: We can add a constant for string `event.name` in RumConstants, nearby the other attribute key for it to keeps things at one place. :)

surbhiia (approved):

PR #1428, opened 2025-09-26
DEMRUM-3499: Fix Mapping File Gradle Plugin does not work with configuration-cache=true

tonzhan2:
Fixed configuration cache compatibility by eliminating Project reference captures in task actions

### Description

Customer was running into issues building their app with the gradle plugin when configuration caching was enabled.
 
The configuration cache requires that all task actions be serializable, but our plugin was inadvertently capturing Project references through lambda closures. This occurred because task action lambdas captured `this` (the plugin instance), which contained a `project` field, violating configuration cache requirements.

#### The code contained several potentially problematic areas:
- Plugin instance (this) captured in task action lambdas due to calling instance methods
- Extension objects passed to task actions containing potential Project references
- FileCollection objects (containing Project references) instead of extracted File objects
- Logger instances created from project.logger at configuration time

#### Changes made:
- Moved all task action logic to static `TaskActions` companion object methods to eliminate plugin instance capture
- Extract primitive values (buildDir, variantName, etc.) at configuration time before task creation
- Create fresh `SplunkLogger` instances inside task actions using `task.logger` instead of captured loggers
- Pass individual configuration values instead of extension objects to avoid hidden references
- Convert `BuildIdInjector` to object pattern with static methods requiring no Project access
- Extract `Set<File>` from `FileCollection.files` at configuration time to avoid API object capture

This resolves configuration cache serialization failures when `org.gradle.configuration-cache=true` is enabled and eliminates `"cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject'"` and similar errors.

#### Testing:

Reproduced the issue with the original mapping file plugin in three different apps:
- Sample app
- Wikipedia app
- Cloudkart app

In the Wikipedia app, also reproduced customers exact tool versioning - AGP 8.13.0, Kotlin 2.2.20, KSP 2.2.20-2.0.3

Verified that the fix works with all three apps and plugin still functions correctly. Build apps both using the emulator run button, and via command line with multiple build type + flavor combinations

### Checklist

- [x] (If applicable) I have added unit tests for my changes.
- [x] (If applicable) I have updated the sample app for integration testing.
- [x] (If applicable) I have updated any relevant documentation.

### How to Test These Changes

Build any app with the gradle plugin as usual with the following set in your `gradle.properties` file:
`org.gradle.configuration-cache=true
`
and verify that build is successful:


instrumentation/buildtime/mapping-file/plugin/src/main/java/com/splunk/rum/mappingfile/plugin/MappingFilePlugin.kt
```diff
@@ -66,44 +63,199 @@ class MappingFilePlugin : Plugin<Project> {
...
+                    accessToken = accessToken,
+                    realm = realm,
+                    failBuildOnUploadFailure = failBuildOnUploadFailure
+                )
+            }
+        }
+        logger.info("Upload", "Successfully configured upload task for variant '$variantName'")
+    }
+
+    companion object {
```
OndrejSynakSL, line 188: Why there is object in companion object? What does it solve?

OndrejSynakSL (changes requested):
LGTM, just one small comment

PR #1467, opened 2025-10-21
DEMRUM-3330: Decouple endpointConfiguration from Agent Installation

tonzhan2:
Decouple endpointConfiguration from Agent Installation

### Description

- Made EndpointConfiguration an optional field in AgentConfiguration and other files
- Implemented preferences/state pattern for endpoint configuration following the existing API architecture:
     - Created `AgentPreferences` class with mutable `endpointConfiguration` property accessible via        `SplunkRum.instance.preferences.endpointConfiguration`
     - Modified State class to expose read-only `endpointConfiguration` via `SplunkRum.instance.state.endpointConfiguration`
    - Both properties share a thread-safe backing store (`AtomicReference<EndpointConfiguration?>`) ensuring consistency
- Write the endpoint to `AgentStorage` trace/log URLs when the endpoint is set via preferences
- Only set write the `config.endpoint` in `ConfigurationManager` if the endpoint is passed into the `AgentConfiguration` on install
- In `AndroidSpanExporter`, check that the endpoint baseURL exists, if not, we let it behave exactly as it did when the app is not in foreground and uploads are deferred until it is in the foreground (store span ID for deferred upload using `addBufferedSpanId` API)
- Added equivalent APIs in `AgentStorage` to store session replay IDs for deferred upload `addBufferedSessionReplayId ` `getBufferedSessionReplayIds ` `clearBufferedSessionReplayIds `
- In `AndroidLogRecordExporter`, for the session replay logs, we check if the endpoint baseURL exists. If not, we defer uploads using `addBufferedSessionReplayId`. If it exists, we flush all as usual
- Removed calls to `clearBufferedIds` in `shutdown` methods of both log and span exporter to ensure buffer ids are always persisted in case of app restarts/deferred uploads

### API Architecture

Following the agent's established preferences/state patterns, endpoint configuration is now exposed through two properties:
Preferences (Mutable): `SplunkRum.instance.preferences.endpointConfiguration`
- Allows runtime configuration of the endpoint
- Set method persists URLs to AgentStorage and triggers flush of buffered data

State (Read Only): `SplunkRum.instance.state.endpointConfiguration`
- Returns the actual endpoint configuration being used by the agent

Currently, both properties return the same value (reading from the shared endpointRef backing store). This pattern is extensible to support potential future remote configuration support, where preferences would reflect what the developer manually set, while state would show the effective value (potentially overridden by remote config).

This mirrors the existing pattern used for other functionality groups like `User`

### Implementation Notes

In order to be able to save the endpoint baseURL to AgentStorage when the endpoint is set, we had to add the `Application` field to the the `SplunkRum` constructor. Thats because in the endpoint setter, we need to call `AgentStorage.attach(application)` to persist the endpoint URLs, and that method requires the Application/Context parameter. Without storing the application reference in the constructor, we have no way to access storage when the endpoint setter is called later.

This shouldn't be a problem for other use cases since this constructor is private, and only the companion object's `install` method creates the `SplunkRum` instances. All the real instances already have the `Application` object so it can just be passed through. And the No op instance was handled by making the `Application` field nullable. So this change should be fairly benign.

The endpoint configuration backing store (`endpointRef: AtomicReference<EndpointConfiguration?>`) was also added as a constructor parameter with a default value to maintain the existing constructor pattern while enabling the shared backing store between `preferences` and `state`.

### Checklist

### Generative AI usage

### How to Test These Changes

- Build Agent
- Use any button in the Sample app to set endpoint config using new API: 
```
SplunkRum.instance.preferences.endpointConfiguration = EndpointConfiguration(
    realm = "us0",
    rumAccessToken = "abc123"
)

```
- Do not pass in `endpointConfiguration` upon app launch
- Run app, verify that no data is sent out of the agent. (Check Controller and also device logs)
- Press button that sets `endpointConfiguration`
- Verify that auto and manually generated data (session replay, custom tracking, ANR etc) are all sent out in the device logs, and that it shows up in the controller
- Optional - verify that calling `SplunkRum.instance.state.endpointConfiguration` returns correct value

### DIFF
--- a/app/src/main/java/com/splunk/app/ui/endpointconfiguration/EndpointConfigurationFragment.kt
+++ b/app/src/main/java/com/splunk/app/ui/endpointconfiguration/EndpointConfigurationFragment.kt
@@ -1,0 +1,159 @@
+package com.splunk.app.ui.endpointconfiguration
+
+/*
+ * Copyright 2025 Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import android.os.Bundle
+import android.util.Log
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.Toast
+import com.splunk.app.R
+import com.splunk.app.databinding.FragmentEndpointConfigurationBinding
+import com.splunk.app.ui.BaseFragment
+import com.splunk.rum.integration.agent.api.EndpointConfiguration
+import com.splunk.rum.integration.agent.api.SplunkRum
+import java.net.URL
+
+/**
+ * A fragment for testing endpoint configuration APIs.
+ */
+class EndpointConfigurationFragment : BaseFragment<FragmentEndpointConfigurationBinding>() {
+
+    override val titleRes: Int = R.string.endpoint_configuration_title
+
+    override val viewBindingCreator: (LayoutInflater, ViewGroup?, Boolean) -> FragmentEndpointConfigurationBinding
+        get() = FragmentEndpointConfigurationBinding::inflate
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        with(viewBinding) {
+            setRealmToken.setOnClickListener { setEndpointWithRealmToken() }
+            setTracesOnly.setOnClickListener { setEndpointWithTracesOnly() }
+            setTracesAndLogs.setOnClickListener { setEndpointWithTracesAndLogs() }
+            clearEndpoint.setOnClickListener { clearEndpoint() }
+            logEndpoints.setOnClickListener { logCurrentEndpoints() }
+        }
+    }
+
+    /**
+     * Sets endpoint configuration using realm and token.
+     * Please replace the fields with the actual realm and token you want to test this API with
+     */
+    private fun setEndpointWithRealmToken() {
+        Log.i(TAG, "Setting endpoint with realm and token")
+        try {
+            SplunkRum.instance.preferences.endpointConfiguration =
+                EndpointConfiguration(realm = "us0", rumAccessToken = "abc123")
+            showToast("Endpoint set with realm and token")
+            logCurrentEndpoints()
+        } catch (e: Exception) {
+            Log.e(TAG, "Error setting endpoint with realm/token", e)
+            showToast("Error: ${e.message}")
+        }
+    }
+
+    /**
+     * Sets endpoint configuration with traces URL only.
+     * Please replace the fields with the actual URL you want to test with
+     */
+    private fun setEndpointWithTracesOnly() {
+        Log.i(TAG, "Setting endpoint with traces only")
+        try {
+            SplunkRum.instance.preferences.endpointConfiguration =
+                EndpointConfiguration(URL("https://custom-traces.example.com/v1/traces"))
+            showToast("Endpoint set with traces only")
+            logCurrentEndpoints()
+        } catch (e: Exception) {
+            Log.e(TAG, "Error setting endpoint with traces only", e)
+            showToast("Error: ${e.message}")
+        }
+    }
+
+    /**
+     * Sets endpoint configuration with both traces and logs URLs.
+     * Please replace the fields with the actual URLs you want to test with
+     */
+    private fun setEndpointWithTracesAndLogs() {
+        Log.i(TAG, "Setting endpoint with traces and logs")
+        try {
+            SplunkRum.instance.preferences.endpointConfiguration =
+                EndpointConfiguration(
+                    URL("https://custom-traces.example.com/v1/traces"),
+                    URL("https://custom-logs.example.com/v1/logs")
+                )
+            showToast("Endpoint set with traces and logs")
+            logCurrentEndpoints()
+        } catch (e: Exception) {
+            Log.e(TAG, "Error setting endpoint with traces and logs", e)
+            showToast("Error: ${e.message}")
+        }
+    }
+
+    /**
+     * Clears the endpoint configuration.
+     */
+    private fun clearEndpoint() {
+        Log.i(TAG, "Clearing endpoint")
+        try {
+            SplunkRum.instance.preferences.endpointConfiguration = null
+            showToast("Endpoint cleared")
+            logCurrentEndpoints()
+        } catch (e: Exception) {
+            Log.e(TAG, "Error clearing endpoint", e)
+            showToast("Error: ${e.message}")
+        }
+    }
+
+    /**
+     * Logs the current endpoint configuration from both State and Preferences.
+     */
+    private fun logCurrentEndpoints() {
+        Log.i(TAG, "=== CURRENT ENDPOINT CONFIGURATION ===")
+
+        Log.i(TAG, "--- state.endpointConfiguration ---")
+        val stateEndpoint = SplunkRum.instance.state.endpointConfiguration
+        logEndpointDetails(stateEndpoint, "State")
+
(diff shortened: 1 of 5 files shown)


integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -65,6 +69,37 @@ class SplunkRum private constructor(
...
+                    val storage = AgentStorage.attach(application)
+                    storage.writeTracesBaseUrl(newEndpoint.tracesEndpoint!!.toExternalForm())
+                    storage.writeLogsBaseUrl(newEndpoint.logsEndpoint!!.toExternalForm())
+
+                    endpointRef.set(newEndpoint)
+
+                    Logger.d(TAG, "Endpoint configured, flushing cached data")
+
+                    // Cast to SDK to access forceFlush
+                    (openTelemetry as? OpenTelemetrySdk)?.sdkTracerProvider?.forceFlush()
```
OndrejSynakSL: I think we need to force flush both tracer and logs provider

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/configuration/ConfigurationManager.kt
```diff
@@ -29,9 +29,10 @@ internal class ConfigurationManager private constructor(private val agentStorage
             config = config.copy(appVersion = context.packageManager.getPackageInfo(context.packageName, 0).versionName)
         }
 
-        agentStorage.writeLogsBaseUrl(config.endpoint.logsEndpoint!!.toExternalForm())
-        agentStorage.writeTracesBaseUrl(config.endpoint.tracesEndpoint!!.toExternalForm())
-
+        config.endpoint?.let { endpoint ->
+            agentStorage.writeTracesBaseUrl(endpoint.tracesEndpoint!!.toExternalForm())
+            agentStorage.writeLogsBaseUrl(endpoint.logsEndpoint!!.toExternalForm())
```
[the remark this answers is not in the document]
tonzhan2: Im not following why the log endpoint doesn't need to be setup here? And even if it doesn't need to be setup here, why wouldn't we just do it?

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -65,6 +69,37 @@ class SplunkRum private constructor(
...
+            value?.let { newEndpoint: EndpointConfiguration ->
+                synchronized(endpointLock) {
+                    val application = application ?: run {
+                        Logger.w(TAG, "Cannot set endpoint: application not available (noop instance?)")
+                        return@synchronized
+                    }
+
+                    val storage = AgentStorage.attach(application)
+                    storage.writeTracesBaseUrl(newEndpoint.tracesEndpoint!!.toExternalForm())
+                    storage.writeLogsBaseUrl(newEndpoint.logsEndpoint!!.toExternalForm())
```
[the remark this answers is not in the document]
SenNeonoveNoci: @tonzhan2 What Ondra means is that `newEndpoint.tracesEndpoint!!` can cause a crash, because at this point you are not guaranteed to have `newEndpoint.tracesEndpoint` defined.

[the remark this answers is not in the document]
SenNeonoveNoci: Right, the `tracesEndpoint` should be fine. But I would still prefer to never use `!!` in the code, except in some really “hackish” scenarios like using reflection where it might be unavoidable. But you’re right, this is a nit, and I’ll leave it up to you how you want to handle it. Feel free to mark this convo as resolved.

common/storage/src/main/java/com/splunk/rum/common/storage/AgentStorage.kt
```diff
@@ -244,6 +244,33 @@ class AgentStorage(context: Context) : IAgentStorage {
...
 
+    override fun addBufferedSessionReplayId(id: String) {
+        val ids = getBufferedSessionReplayIds().toMutableSet()
+        if (ids.add(id)) {
+            val array = JSONArray(ids)
+            preferences.putString(SESSION_REPLAY_IDS_KEY, array.toString()).commit()
+        }
+    }
+
+    override fun getBufferedSessionReplayIds(): List<String> {
```
TomasChladekSL: Please consider replacing the implementation with the following code.

```
        val json = preferences.getString(SESSION_REPLAY_IDS_KEY)

        if (json != null)
            try {
                return JSONArray(json).map { array, index -> array.getString(index) }
            } catch (e: JSONException) {
                Logger.e(TAG, "getBufferedSessionReplayIds()", e)
                clearBufferedSessionReplayIds()
            }

        return emptyList()
```

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/SplunkRum.kt
```diff
@@ -249,6 +263,8 @@ class SplunkRum private constructor(
             agentConfiguration: AgentConfiguration,
             vararg moduleConfigurations: ModuleConfiguration
         ): SplunkRum {
+            val storage = AgentStorage.attach(application)
```
surbhiia: Instead of changing the SplunkRum constructor to be able to pass AgentStorage to be able to initialize AgentPreferences, why don't you do this work in SplunkRumAgentCore instead? If you do it there, you won't have to do unnecessary initializations if anyway splunkRum is going to be noop. Also, SplunkRumAgentCore is the place we initialize needed stuff.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/AgentConfiguration.kt
```diff
@@ -47,7 +47,7 @@ import io.opentelemetry.sdk.trace.data.SpanData
  * @property deferredUntilForeground Whether to defer tracing until the app is brought to the foreground.
  */
 data class AgentConfiguration @JvmOverloads constructor(
-    val endpoint: EndpointConfiguration,
+    val endpoint: EndpointConfiguration? = null,
```
surbhiia: Below in this class where no-op object is created, we no longer need to create an empty EndpointConfiguration object. It will be null by default.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/AgentPreferences.kt
```diff
@@ -0,0 +1,74 @@
...
+
+class AgentPreferences internal constructor(
+    private val agentStorage: IAgentStorage?,
+    private val endpointRef: AtomicReference<EndpointConfiguration?>,
+    private val openTelemetry: OpenTelemetry
+) {
+    private val endpointLock = Any()
+
+    var endpointConfiguration: EndpointConfiguration?
+        get() = endpointRef.get()
```
surbhiia: Why do we need a getter in AgentPreferences when endpoint can be read via State. Should AgentPrefernces just provide a setter to set and if they need to read it is from state as per the API Architecture mentioned in description.

surbhiia:
Please add buttons in sample app to edit endpoint configuration using `preferences`. Add all test cases. 
Also add button to read endpointConfiguration via `state`.

integration/agent/api/src/main/java/com/splunk/rum/integration/agent/api/configuration/ConfigurationManager.kt
```diff
@@ -29,8 +29,12 @@ internal class ConfigurationManager private constructor(private val agentStorage
             config = config.copy(appVersion = context.packageManager.getPackageInfo(context.packageName, 0).versionName)
         }
 
-        agentStorage.writeLogsBaseUrl(config.endpoint.sessionReplayEndpoint!!.toExternalForm())
-        agentStorage.writeTracesBaseUrl(config.endpoint.traceEndpoint!!.toExternalForm())
+        config.endpoint?.let { endpoint ->
+            agentStorage.writeTracesBaseUrl(endpoint.traceEndpoint!!.toExternalForm())
```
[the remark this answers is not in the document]
tonzhan2: This is expected behavior. The traceEndpoint is required for core functionality and sending traces. The sessionReplayEndpoint is optional because session replay is an optional feature. If traceEndpoint is null, we log an error and abort the operation to prevent invalid state.
 
 But yes, can make field in EndpointConfiguration non nullable

surbhiia (approved):
Looks Good, Thanks for the changes! Just address this comment before you merge - https://github.com/signalfx/splunk-otel-android/pull/1467#discussion_r2519429121

PR #1569, opened 2026-03-13
DEMRUM-4248: Navigation implementation

tonzhan2:
### Description

**Replace legacy navigation with an event-based model**
- Removed the previous screen/tracer-based implementation (screen/ and tracer/ packages: `ScreenTrackerIntegration`, `VisibleScreenTracker`, `ActivityTracer`, `FragmentTracer`, and their Activity/Fragment callbacks).
- Navigation is now driven by a single path: automatic (lifecycle) and manual (`Navigation.track`) both feed the same emitter.

**Automatic tracking**
- `ScreenChangeDetector` observes Activity and Fragment lifecycle (resumed/paused) and derives the visible screen (Fragment over Activity when both exist).
- Emits **only on resumed** (onFragmentResumed / onActivityResumed) as the trigger
- `NavigationEventEmitter` emits OpenTelemetry log records for `device.app.ui.navigation` with `screen.name` and `last.screen.name`.
- Events are cached until the SDK is installed, for hybrid SDK (Flutter/RN) support.

**Manual tracking API**
- `Navigation.track(screenName)` and `Navigation.track(screenName, attributes)` record a navigation to a screen with optional OTel attributes; the module’s listener forwards to the same emitter and keeps automatic and manual tracking in sync via `recordEmittedScreen` / `lastEmittedScreenName`.
- Listener uses current screen (`ScreenNameTracker.screenName`) as the “from” screen so transitions (e.g. Menu → OkHttp) are correct.

**Screen name resolution**
- `ScreenNameDescriptor` resolves the display name for Activity/Fragment in order: `@NavigationElement(name)` → `@RumScreenName(name)` → class simple name. `@NavigationElement` also supports `isIgnored` to exclude a screen from automatic tracking.
- **NavHostFragment** is ignored by default (Jetpack Navigation container; not a user-facing screen).

**Configuration**
- `NavigationModuleConfiguration` has `isEnabled` and `isAutomatedTrackingEnabled`. Customers pass it via the existing module configuration array when installing the SDK.

### Checklist

### Generative AI usage

### How to Test These Changes

- Enable automated tracking via `NavigationModuleConfiguration(isAutomatedTrackingEnabled = true)` (as in the sample app). Optionally add manual `Navigation.track(...)` calls and `@NavigationElement` annotations.
- Open the app and navigate between screens (e.g. Main → Menu → Crash Reports → back to Menu → OkHttp → back to Menu, etc.).
- In logcat filter by `NavigationEventEmitter` and confirm emitted events match the path (e.g. `null -> Main`, `Main -> Menu`, `Menu -> CrashReportsFragment`, …) and that annotations take effect (e.g. OkHttp screen shows as `ok` when using `@NavigationElement(name = "ok")`).
- Confirm exported spans have correct `screen.name` and `last.screen.name`.
- Confirm log events appear correctly in session view (e.g. mon0/lab0).

### Future Considerations

**NavigationEventProcessor**
- A `NavigationEventProcessor` (or similar) abstraction was not implemented in this PR. That will be done in the jetpack compose support work

Data model is subject to change!


integration/navigation/src/main/java/com/splunk/rum/integration/navigation/Navigation.kt
```diff
@@ -16,16 +16,28 @@
...
+import io.opentelemetry.api.common.Attributes
+
 class Navigation internal constructor() {
 
     internal var listener: Listener? = null
 
+    /**
+     * Record a navigation to [screenName] (manual tracking).
+     */
     fun track(screenName: String) {
```
TomasChladekSL: Please use default argument 'attributes: Attributes = Attributes.empty()'

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -58,35 +70,73 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
-            val screenSpan = provider.get(GlobalRumConstants.RUM_TRACER_NAME)
-                .spanBuilder(GlobalRumConstants.NAVIGATION_SPAN_NAME)
-                .setAttribute(GlobalRumConstants.COMPONENT_KEY, RumConstant.COMPONENT_NAVIGATION)
-                .setStartTimestamp(timeNow)
-                .startSpan()
+        application.registerActivityLifecycleCallbacks(fragmentActivityCallback)
+    }
+
+    private val navigationListener = object : Navigation.Listener {
+        override fun onScreenNameChanged(screenName: String, attributes: io.opentelemetry.api.common.Attributes) {
```
TomasChladekSL: import

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -58,35 +70,73 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
+            Logger.d(TAG, "onScreenNameChanged(screenName: $screenName)")
```
TomasChladekSL: Please log 'Attributes' too

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -58,35 +70,73 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
 
-            screenSpan.end(timeNow)
+            val navigationEmitter = emitter ?: return
```
TomasChladekSL: You should cache the events here.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -58,35 +70,73 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
+            val previousScreenName = ScreenNameTracker.screenName
+            navigationEmitter.emitNavigationEvent(screenName, previousScreenName, attributes)
```
TomasChladekSL: 'previousScreenName' is a technical detail that should be inside 'emitNavigationEvent' function. (Right now, every caller have to do 'emitNavigationEvent(screenName, ScreenNameTracker.screenName, attributes)')

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -19,25 +19,37 @@ package com.splunk.rum.integration.navigation
...
+ */
 internal object NavigationModuleIntegration : ModuleIntegration<NavigationModuleConfiguration>(
     defaultModuleConfiguration = NavigationModuleConfiguration()
 ) {
 
-    private const val TAG = "NavigationIntegration"
+    private const val TAG = "NavigationModuleIntegration"
     private var currentActivityReference: WeakReference<Activity>? = null
 
+    private var emitter: NavigationEventEmitter? = null
```
TomasChladekSL: Please set an instance here, then remove '!!' nad '?.'.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -19,25 +19,37 @@ package com.splunk.rum.integration.navigation
...
+    private var screenChangeDetector: ScreenChangeDetector? = null
```
TomasChladekSL, line 51: Please set an instance here, then remove '!!' nad '?.'.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -58,35 +70,73 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
+
+                val application = context.applicationContext as Application
+                screenChangeDetector = ScreenChangeDetector(navigationEmitter)
+
+                registerActivityLifecycle(application, screenChangeDetector!!)
+                registerFragmentLifecycle(application, screenChangeDetector!!)
+
+                // Seed detector with already visible activity for late/hybrid installs.
+                currentActivityReference?.get()?.let { activity ->
+                    screenChangeDetector!!.onActivityResumed(activity)
```
TomasChladekSL: You are listening for 'onActivityStarted' on line 55, but you are calling 'onActivityResumed'. Is that correct?

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/callback/NavigationFragmentActivityCallback29.kt
```diff
@@ -14,21 +14,22 @@
...
-import com.splunk.android.common.logger.Logger
+import androidx.fragment.app.FragmentManager
 
-internal class FragmentActivityCallback29(override val fragmentCallback: FragmentLifecycleCallbacks) :
-    FragmentActivityCallback {
+/**
+ * Registers [NavigationFragmentCallback] for API 29+ (uses onActivityPreCreated).
+ */
+internal class NavigationFragmentActivityCallback29(private val fragmentCallback: NavigationFragmentCallback) :
+    com.splunk.android.common.utils.adapters.ActivityLifecycleCallbacksAdapter {
```
TomasChladekSL: import

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/callback/NavigationFragmentActivityCallback29.kt
```diff
@@ -14,21 +14,22 @@
...
-import androidx.fragment.app.FragmentManager.FragmentLifecycleCallbacks
-import com.splunk.android.common.logger.Logger
+import androidx.fragment.app.FragmentManager
 
-internal class FragmentActivityCallback29(override val fragmentCallback: FragmentLifecycleCallbacks) :
-    FragmentActivityCallback {
+/**
+ * Registers [NavigationFragmentCallback] for API 29+ (uses onActivityPreCreated).
+ */
+internal class NavigationFragmentActivityCallback29(private val fragmentCallback: NavigationFragmentCallback) :
```
TomasChladekSL, line 30: Please use '@RequiresApi(Build.VERSION_CODES.Q)' annotation.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/callback/NavigationFragmentActivityCallback29.kt
```diff
@@ -14,21 +14,22 @@
...
+    com.splunk.android.common.utils.adapters.ActivityLifecycleCallbacksAdapter {
 
     override fun onActivityPreCreated(activity: Activity, savedInstanceState: Bundle?) {
-        Logger.d("FragmentActivityCallback29", "onActivityPreCreated")
-        if (activity is FragmentActivity) {
-            Logger.d("FragmentActivityCallback29", "onActivityPreCreated FragmentActivity")
+        if (Build.VERSION.SDK_INT >= 29 && activity is FragmentActivity) {
```
TomasChladekSL: SDK check is not necessary

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/NavigationEventEmitter.kt
```diff
@@ -0,0 +1,124 @@
...
+     * Updates [ScreenNameTracker.screenName] and emits a log record.
+     * Caches the event if installation is not complete.
+     */
+    fun emitNavigationEvent(
+        screenName: String,
+        previousScreenName: String?,
+        attributes: Attributes = Attributes.empty()
+    ) {
+        val timestamp = System.currentTimeMillis()
+        synchronized(lock) {
```
TomasChladekSL: Navigation through the application is synchronous, so thread synchronization is unnecessary.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/NavigationEventEmitter.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        emitEventInternal(screenName, previousScreenName, attributes, timestamp)
+    }
+
+    /**
+     * Actually emit the event to OpenTelemetry.
+     */
+    private fun emitEventInternal(
+        screenName: String,
+        previousScreenName: String?,
```
TomasChladekSL, line 65: As mentioned in 'NavigationModuleIntegration', please handle 'previousScreenName' internally.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/NavigationEventEmitter.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        val builder = logger.get(GlobalRumConstants.RUM_TRACER_NAME)
+            .logRecordBuilder()
+            .setTimestamp(timestamp, TimeUnit.MILLISECONDS)
+            .setAttribute(GlobalRumConstants.LOG_EVENT_NAME_KEY, RumConstant.NAVIGATION_LOG_EVENT_NAME)
+            .setAttribute(GlobalRumConstants.COMPONENT_KEY, RumConstant.COMPONENT_NAVIGATION)
+            .setAttribute(GlobalRumConstants.SCREEN_NAME_KEY, screenName)
+
+        previousScreenName?.let {
+            builder.setAttribute(GlobalRumConstants.LAST_SCREEN_NAME_KEY, it)
```
TomasChladekSL: Feel free to use 'setAttribute(GlobalRumConstants.LAST_SCREEN_NAME_KEY, previousScreenName)', the implementation ignores null values.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/NavigationEventEmitter.kt
```diff
@@ -0,0 +1,124 @@
...
+        builder.setAllAttributes(attributes)
+        builder.emit()
+    }
+
+    /**
+     * Process all cached events. Called when installation is complete.
+     */
+    fun processCachedEvents() {
+        val cachedEvents: List<CachedEvent>
+        synchronized(lock) {
```
TomasChladekSL: Because navigation is a synchronous process, thread synchronization is not required.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/NavigationEventEmitter.kt
```diff
@@ -0,0 +1,124 @@
...
+            cachedEvents = cache.toList()
+            cache.clear()
+            isInstallComplete = true
+        }
+        if (cachedEvents.isNotEmpty()) {
+            Logger.d(TAG) { "Processing cached navigation events (size: ${cachedEvents.size})" }
+            cachedEvents.forEach { event ->
```
TomasChladekSL: Use 'forEachFast'

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
```diff
@@ -0,0 +1,111 @@
...
+    private val handler = Handler(Looper.getMainLooper())
+    private var lastResumedActivityName: String? = null
+    private var previouslyLastResumedActivityName: String? = null
+    private var lastResumedFragmentName: String? = null
+    private var previouslyLastResumedFragmentName: String? = null
+
+    /**
+     * Current visible screen name: fragment if any, else activity.
+     */
+    private fun currentVisibleScreenName(): String? {
```
TomasChladekSL: 'getCurrentVisibleScreenName()'

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/descriptor/ScreenNameDescriptor.kt
```diff
@@ -30,12 +31,26 @@ internal object ScreenNameDescriptor {
...
+        if (element is Fragment && element.javaClass.name == NAV_HOST_FRAGMENT_CLASS_NAME) {
+            return true
+        }
+        return getNavigationElement(element)?.isIgnored ?: getRumScreenName(element)?.isIgnored ?: false
+    }
+
+    private const val NAV_HOST_FRAGMENT_CLASS_NAME = "androidx.navigation.fragment.NavHostFragment"
 
-    private fun getAnnotation(element: Any): RumScreenName? {
+    private fun getRumScreenName(element: Any): RumScreenName? {
```
TomasChladekSL, line 49: Do we still support 'RumScreenName'? I think it should be replaced with 'NavigationElement'.

TomasChladekSL (changes requested):

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/Navigation.kt
tonzhan2: merged into one with a default argument. Added @JvmOverloads so Java callers don't have to pass Attributes.empty() explicitly.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
tonzhan2: Sounds good. I moved the emitter to an eagerly created val and set the listener in` onAttach()`, so manual `track()` calls before` onInstall() `now go straight into the emitter's existing cache.` onPostInstall() `drains them via` processCachedEvents()` as before

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/NavigationEventEmitter.kt
tonzhan2: gives a kotlin compiler warning, so will leave it as is

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
tonzhan2: `screenChangeDetector` needs to stay nullable at the field level since it's only created when automated tracking is enabled. 

I introduced a local val detector within `onInstall` to eliminate all the !! usages. The remaining ?. in the listener is intentional as its a no op when only manual tracking is active.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
tonzhan2: `onActivityStarted` worked in practice since the activity is always resumed by the time `onInstall` runs, but I changed the adapter to capture on `onActivityResumed` instead, which is consistent with what we pass to the detector. Functionally the same but this makes the intent clearer

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
tonzhan2: cool, `emitNavigationEvent` now resolves `previousScreenName` internally from `ScreenNameTracker.screenName`, and callers no longer pass it. The default screen name ("unknown") is filtered to null via `takeIf` so the first event omits` last.screen.name` rather than emitting "unknown".

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/descriptor/ScreenNameDescriptor.kt
tonzhan2: `NavigationElement` takes precedence. But keeping `RumScreenName` for now for backwards compatibility

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/NavigationEventEmitter.kt
```diff
@@ -0,0 +1,113 @@
...
+        builder.emit()
+    }
+
+    /**
+     * Process all cached events. Called when installation is complete.
+     */
+    fun processCachedEvents() {
+        val cachedEvents = cache.toList()
+        cache.clear()
+        isInstallComplete = true
```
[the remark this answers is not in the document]
tonzhan2: only gets called from main thread

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -19,33 +19,46 @@ package com.splunk.rum.integration.navigation
...
     private val activityLifecycleCallbacksAdapter = object : ActivityLifecycleCallbacksAdapter {
-        override fun onActivityStarted(activity: Activity) {
+        override fun onActivityResumed(activity: Activity) {
             currentActivityReference = WeakReference(activity)
         }
     }
 
     override fun onAttach(context: Context) {
         super.onAttach(context)
+        Navigation.instance.listener = navigationListener
```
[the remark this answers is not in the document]
tonzhan2: screen.name belongs to all spans anyway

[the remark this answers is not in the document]
tonzhan2: the window between onAttach and onInstall is milliseconds during SDK init, and navigation events require user interaction so growth is at most one entry per screen change. Even if install is short circuited, a handful of cached events is negligible memory. Not worth the added complexity.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -57,36 +70,80 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
+            // Seed detector with already visible activity/fragment for late/hybrid installs.
+            currentActivityReference?.get()?.let { activity ->
+                if (activity is FragmentActivity) {
+                    val fragmentCallback = NavigationFragmentCallback(detector)
+                    activity.supportFragmentManager.registerFragmentLifecycleCallbacks(fragmentCallback, true)
+                    activity.supportFragmentManager.fragments.forEach { fragment ->
+                        if (fragment.isResumed) {
+                            detector.onFragmentResumed(fragment)
+                        }
+                    }
```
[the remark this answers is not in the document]
tonzhan2: this only affects the first navigation event in late install scenarios with nested fragments (eg NavHostFragment), it self corrects on the next fragment resume since registerFragmentLifecycleCallbacks(callback, true) handles nested fragments going forward. The practical impact is minimal and not worth adding recursive traversal complexity for the seed path.

TomasChladekSL (approved):

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/NavigationEventEmitter.kt
```diff
@@ -0,0 +1,113 @@
...
+        emitEventInternal(screenName, previousScreenName, attributes, timestamp)
+    }
+
+    private fun emitEventInternal(
+        screenName: String,
+        previousScreenName: String?,
+        attributes: Attributes,
+        timestamp: Long
+    ) {
+        ScreenNameTracker.screenName = screenName
```
surbhiia: You should not set the screen name in tracker here. You've already set it on line 51 correctly. For scenario when navigation is not installed and this data is cached and this method is called from cache processing, it'll screw up the current running app's screen name in the tracker.

tonzhan2: This is a good catch thank you. I think it was accidentally left there after we added the caching layer, removed it now

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
```diff
@@ -0,0 +1,114 @@
...
+        if (fragment is DialogFragment) {
+            previouslyLastResumedFragmentName = lastResumedFragmentName
+        }
+
+        val name = ScreenNameDescriptor.getName(fragment)
+        lastResumedFragmentName = name
+        tryEmitIfChanged()
+    }
+
+    fun onFragmentPaused(fragment: Fragment) {
```
surbhiia, line 78: The logic here is wrong because of `previouslyLastResumedFragmentName` being set on line 89.

It should be set inside each if/else case separately as needed: 

```

       if (fragment is DialogFragment) {
            lastResumedFragmentName = previouslyLastResumedFragmentName
            previouslyLastResumedFragmentName = null
        } else if (lastResumedFragmentName == ScreenNameDescriptor.getName(fragment)) {
             lastResumedFragmentName = null
        }

```

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
```diff
@@ -0,0 +1,114 @@
...
+        // Defer so fragment callbacks, which run synchronously during the same resume, execute
+        // first thus avoiding emitting an intermediate activity only event when a fragment is present
+        handler.post { tryEmitIfChanged() }
+    }
+
+    fun onActivityPaused(activity: Activity) {
+        if (ScreenNameDescriptor.isIgnored(activity)) return
+
+        val name = ScreenNameDescriptor.getName(activity)
+        previouslyLastResumedActivityName = name
```
surbhiia: `previouslyLastResumedActivityName` is written here but never read. 

If we do not have a similar concept like **Dialog Fragments** in activities, we do not need this variable.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
```diff
@@ -0,0 +1,114 @@
...
+import androidx.fragment.app.Fragment
+import com.splunk.rum.integration.navigation.descriptor.ScreenNameDescriptor
+
+/**
+ * Detects visible screen changes (Activity/Fragment) and notifies [NavigationEventEmitter].
+ * Fragment takes precedence over Activity when both are present.
+ *
+ * Emits only on resumed (onFragmentResumed / onActivityResumed) as the trigger.
+ */
+internal class ScreenChangeDetector(private val eventEmitter: NavigationEventEmitter) {
```
surbhiia, line 37: Question - Are we not catering to multiple dialogs stacked on top of each other and they could happen on top of an activity or on top of another fragment?  

Should we ignore DialogFragments totally as these are overlays and not actual screens, this helps us prevent the following issues:
- DialogFragment save/restore is broken for stacked dialogs (only one level of backup).
- DialogFragment dismiss does not emit a return event to the underlying screen (no tryEmitIfChanged in onFragmentPaused).

If it's a dialog covering entire screen, they can manually track it.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -57,36 +70,80 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
+        (context as Application).unregisterActivityLifecycleCallbacks(activityLifecycleCallbacksAdapter)
+        currentActivityReference = null
     }
 
-    private val navigationListener = object : Navigation.Listener {
-        override fun onScreenNameChanged(screenName: String) {
-            Logger.d(TAG) { "onScreenNameChanged(screenName: $screenName)" }
+    override fun onPostInstall() {
+        super.onPostInstall()
+        if (!moduleConfiguration.isEnabled) return
```
surbhiia: Cache should be cleared **when module is disabled** and **also when ONLY manual tracking is used instead**, to prevent holding up all this memory for unnecessary data we tracked in hopes of having automatic instrumentation enabled. You also do not need to wait till `onPostInstall` to clear cache, you can do that in `onInstall` itself. 

Another issue is you're processing the cache and emitting events even when **only manual tracking is enabled** (i.e automatic tracking is disabled explicitly).

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -57,36 +70,80 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
+            screenChangeDetector = detector
+
+            registerActivityLifecycle(application, detector)
+            registerFragmentLifecycle(application, detector)
+
+            // Seed detector with already visible activity/fragment for late/hybrid installs.
+            currentActivityReference?.get()?.let { activity ->
+                if (activity is FragmentActivity) {
+                    val fragmentCallback = NavigationFragmentCallback(detector)
+                    activity.supportFragmentManager.registerFragmentLifecycleCallbacks(fragmentCallback, true)
```
surbhiia, line 95: You already registered fragment callbacks on line 88, if you do it here again for this activity, it exists twice for this activity and would be fired twice next time this activity is resumed? 

Also, you do not use the SDK_INT > 29 filter here if this is actually needed?

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
surbhiia, line 78: So, I went back and forth with AI to determine the missed edge cases and here's a summary:

**Edge cases failing**:

- Fragment removed while activity stays resumed: the transition back to the activity (or parent fragment) is not emitted.
- Nested fragment pauses doesn't correctly restore to the parent fragment's name instead falls to the underlying activity.

**Solution**: 
Add deferred tryEmitIfChanged() in `onFragmentPaused` and ancestor walk to reflect the currently shown fragment in case of nested fragments (here it is assumed that we'll go with ignoring dialog fragments):

``` 
fun onFragmentPaused(fragment: Fragment) {
  if (ScreenNameDescriptor.isIgnored(fragment)) return
  if (fragment is DialogFragment) return
  val name = ScreenNameDescriptor.getName(fragment)
  if (lastResumedFragmentName == name) {
      lastResumedFragmentName = findResumedAncestorName(fragment)
  }
  handler.post { tryEmitIfChanged() }
}

private fun findResumedAncestorName(fragment: Fragment): String? {
  var parent = fragment.parentFragment
  while (parent != null) {
      if (parent.isResumed && !ScreenNameDescriptor.isIgnored(parent) && parent !is DialogFragment) {
          return ScreenNameDescriptor.getName(parent)
      }
      parent = parent.parentFragment
  }
  return null
}

```

 Scenario | Before | After | Preserved? |
|---|---|---|---|
| Fragment A → Fragment B | Emits "B" | Emits "B" (deferred post is deduped) | ✅ Yes |
| Activity A → Activity B (no fragments) | Emits "B" | Emits "B" (no fragment pause involved) | ✅ Yes |
| Activity + Fragment resume together | Fragment takes precedence, one event | Same (deferred post deduped) | ✅ Yes |
| Same screen name re-resumed | Deduped, no event | Same | ✅ Yes |
| App goes to background | No event | No event (both null when post drains) | ✅ Yes |
| Fragment removed, no replacement | ❌ No event (bug) | Emits activity/parent | 🔧 Fixed |
| Nested child pauses, parent visible | ❌ Emits wrong activity name | Emits correct parent fragment | 🔧 Fixed |
| DialogFragment open | Emits dialog as screen | Skipped (dialog is overlay) | ⚠️ Changed |
| DialogFragment dismiss | No return event (bug) | Skipped (dialog is overlay) | ⚠️ Changed |


We should also ensure that our tests cover all possible scenarios we want to cover.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
tonzhan2: Sure, moved the cache clearing into onInstall and clearing too when auto tracking is off.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
tonzhan2: I agree we should ignore them. They are technically overlays, not screen navigations, and it remains a manually trackable option nonetheless. 

Going to skip them entirely in onFragmentResumed and paused. So should simplify the logic quite a bit for those edge cases

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
tonzhan2: It won't be fired twice. `registerFragmentLifecycle` (line 89 now) registers an Application level `ActivityLifecycleCallbacks` that hooks into `onActivityPreCreated` (for API 29+) or `onActivityCreated` (API <29). 

When a new activity is created in the future, it registers `NavigationFragmentCallback` on that activitys supportFragmentManager. But for the current already visible activity, these hooks will never fire since its already past `onCreate`.

Line 94-95  directly registers `NavigationFragmentCallback` on the current activity's supportFragmentManager for the late/hybrid install case.

Pretty much, line 89 covers future activities, line 94 and 95 cover current activities. So there's no overlap.

We don't need the SDK_INT>29 filter 

SDK_INT check is needed on line 94-95 because supportFragmentManager.registerFragmentLifecycleCallbacks is an AndroidX API available on all API levels. The SDK_INT >= 29 distinction in `registerFragmentLifecycle` is only for the application level `onActivityPreCreated` vs `onActivityCreated` hook, not for the fragment callback registration itself

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
tonzhan2: This logic is sound to me. Implemented it. Basically added deferred` tryEmitIfChanged()` in `onFragmentPaused` and a `findResumedAncestorName` ancestor walk. In the normal case (Fragment A pauses, Fragment B resumes), the deferred post is deduped by `lastEmittedScreenName`. For the "fragment removed, no replacement" case, the deferred post correctly falls back to the parent fragment or activity.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
surbhiia: The logic is not clean, please take a look again:
- Cache is cleared once when module is disabled in onInstall
- Cache is cleared again for scenarios when either module is disabled or automatic instrumentation is disabled in postInstall. 

Let's do it at one place where it makes the most sense, I think the sooner the better i.e in `onInstall`. 

- In onPostInstall, you'll still need the condition checks to ensure you process only for automatic scenario. 

But before you do the above changes, another thing that clicked to me just now is that in manual scenario as well, the same method is called to emit event i.e `emitNavigationEvent` which has the cache component. So, can manual track be called earlier to when opentelemetry instance and logger is initialized? If so, in manual scenario too we should set the install complete variable and put stuff in cache until we're ready to process. Sorry, I missed thinking of this scenario before but please dig into it further and confirm what it should be.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
```diff
@@ -59,34 +63,40 @@ internal class ScreenChangeDetector(private val eventEmitter: NavigationEventEmi
...
 
         val name = ScreenNameDescriptor.getName(activity)
-        previouslyLastResumedActivityName = name
         if (lastResumedActivityName == name) {
             lastResumedActivityName = null
         }
     }
 
     fun onFragmentResumed(fragment: Fragment) {
+        if (fragment is DialogFragment) return
```
surbhiia: Let's push this logic to `isIgnored` function itself like the `NavHostFragment` to keep it all in one place?

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
tonzhan2: It's true, manual track calls can technically happen between when the listener is registered and `onPostInstall` since it goes through the same path. 

This can all be simplified by consolidating the logic. We can simply just clear the cache in `onInstall` as you suggested, but only when the module is fully disabled. 

`onPostInstall`  now processes the cache for both auto and manual events (since manual calls are now properly processed from the cache once things are ready).

And since `screenChangeDetector` is only created when auto tracking is enabled, auto events wont end up in the cache if only manual track is enabled

just made the change

surbhiia (approved):

PR #1591, opened 2026-04-08
DEMRUM-4549: Implemented Jetpack Compose Navigation Support

tonzhan2:
### Description

- New public API: `Navigation.registerNavController(navController)` and `Navigation.unregisterNavController(navController)` for tracking Jetpack Compose navigation routes via the Navigation library's NavController

- New `ComposeNavigationTracker` (internal): Manages `NavController` registration, attaches an `OnDestinationChangedListener`, filters out NavGraph containers and dialog destinations, extracts route templates as screen names, and separates resolved arguments into event attributes

-  `ComposeNavigationTracker` internals: Holds a WeakReference to the registered `NavController` and attaches an `OnDestinationChangedListener`. On each destination change it: 
(1) filters out `NavGraph` containers and dialog destinations via `navigatorName`
(2) uses the route template (e.g. `profile/{userId}`) as the screen name to avoid cardinality explosion
(3) extracts resolved arguments into OTel attributes while skipping internal android-support-nav: keys
(4) runs the event through `NavigationEventProcessor` if configured
(5) routes the result through `ScreenChangeDetector.onComposeRouteChanged()` for deduplication and priority handling. Re registering the same NavController (eg on activity resume), Reregistering the same `NavController` (eg on activity resume) routes the current destination through the full `handleDestinationChanged` pipeline so processor logic and filtering apply consistently.

- New `NavigationEvent` and `NavigationEventProcessor`: Allows customers to transform screen names, filter events, or modify attributes before they are emitted. Configured via `NavigationModuleConfiguration.navigationEventProcessor`. only applies to `COMPOSE_ROUTE`, not `ACTIVITY` or `FRAGMENT`

- Updated `ScreenChangeDetector`: Implements a priority system (Compose route > Fragment > Activity) so Compose routes take precedence when active, with proper fallback to Fragment/Activity tracking when the Compose activity is paused or the `NavController` is unregistered

- `compileOnly` dependency on `androidx.navigation:navigation-runtime:2.4.0`: Provides compile time type safety for `NavController` without bundling into consuming apps. Version `2.4.0` is the minimum that supports Compose Navigation

- Sample app: Added `ComposeNavigationActivity` with test scenarios for simple routes, parameterized routes `(profile/{userId})`, optional query parameters `(search?query={query})`, dialog destinations, nested NavHost (unregistered secondary controller), and back navigation to Fragment based screens

- `ScreenChangeDetector` tracks which activity owns the compose route via `composeRouteActivityName`. The compose route is preserved through background/foreground cycles (same activity resumes) and only cleared when a different activity resumes (navigated away), preventing spurious activity name events.

### Checklist

### Generative AI usage

### How to Test These Changes

- build and run sample app
- open compose navigation menu
- navigate through compose navigation part of sample app
- filter logcat by `NavigationEventEmitter` | `ComposeNavigationTracker` to observe events and verify they are correct
- confirm in session that they are correct
- optionally: configure navigationEventProcessor to tailor the navigation events generated for compose elements.

A session involving the compose navigation part of the sample app may look something like this (in this example, screen name "search" was renamed to "Search!!!" in the navigationEventProcessor:

Screenshot 2026-04-08 at 2 00 25 PM

### Future Considerations (Optional)

- Support registration of multiple navControllers
- Support 100% automatic instrumentation via reflection

### DIFF
--- a/app/src/main/java/com/splunk/app/ui/compose/ComposeNavigationActivity.kt
+++ b/app/src/main/java/com/splunk/app/ui/compose/ComposeNavigationActivity.kt
@@ -1,0 +1,309 @@
+@file:Suppress("ktlint:standard:function-naming")
+
+/*
+ * Copyright 2026 Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.app.ui.compose
+
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.AlertDialog
+import androidx.compose.material.Button
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.OutlinedButton
+import androidx.compose.material.Scaffold
+import androidx.compose.material.Text
+import androidx.compose.material.TextButton
+import androidx.compose.material.TopAppBar
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import androidx.navigation.NavController
+import androidx.navigation.NavType
+import androidx.navigation.compose.NavHost
+import androidx.navigation.compose.composable
+import androidx.navigation.compose.currentBackStackEntryAsState
+import androidx.navigation.compose.dialog
+import androidx.navigation.compose.rememberNavController
+import androidx.navigation.navArgument
+import com.splunk.rum.integration.agent.api.SplunkRum
+import com.splunk.rum.integration.navigation.extension.navigation
+
+class ComposeNavigationActivity : ComponentActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContent {
+            MaterialTheme {
+                ComposeNavigationApp(onBack = { finish() })
+            }
+        }
+    }
+}
+
+@Composable
+fun ComposeNavigationApp(onBack: () -> Unit = {}) {
+    val navController = rememberNavController()
+
+    SplunkRum.instance.navigation.registerNavController(navController)
+
+    Scaffold(
+        topBar = { TopAppBar(title = { Text("Compose Navigation") }) }
+    ) { padding ->
+        NavHost(
+            navController = navController,
+            startDestination = "home",
+            modifier = Modifier.padding(padding)
+        ) {
+            composable("home") { HomeScreen(navController, onBack) }
+            composable("settings") { SettingsScreen(navController) }
+            composable(
+                route = "profile/{userId}",
+                arguments = listOf(navArgument("userId") { type = NavType.StringType })
+            ) { backStackEntry ->
+                ProfileScreen(
+                    userId = backStackEntry.arguments?.getString("userId") ?: "unknown",
+                    navController = navController
+                )
+            }
+            composable(
+                route = "search?query={query}",
+                arguments = listOf(
+                    navArgument("query") {
+                        type = NavType.StringType
+                        defaultValue = ""
+                    }
+                )
+            ) { backStackEntry ->
+                SearchScreen(
+                    query = backStackEntry.arguments?.getString("query") ?: "",
+                    navController = navController
+                )
+            }
+            composable("nested_tabs") { NestedTabsScreen(navController) }
+            dialog("confirm_dialog") {
+                ConfirmDialog(navController)
+            }
+        }
+    }
+}
+
+@Composable
+fun HomeScreen(navController: NavController, onBack: () -> Unit = {}) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp),
+        verticalArrangement = Arrangement.spacedBy(12.dp),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Text("Home", style = MaterialTheme.typography.h5)
+        Spacer(modifier = Modifier.height(16.dp))
+
+        Button(
+            onClick = { navController.navigate("settings") },
+            modifier = Modifier.fillMaxWidth()
+        ) { Text("Go to Settings") }
+
+        Button(
+            onClick = { navController.navigate("profile/user_12345") },
+            modifier = Modifier.fillMaxWidth()
+        ) { Text("Profile (user_12345)") }
+
+        Button(
+            onClick = { navController.navigate("profile/user_67890") },
+            modifier = Modifier.fillMaxWidth()
+        ) { Text("Profile (user_67890)") }
+
+        Button(
+            onClick = { navController.navigate("search?query=opentelemetry") },
+            modifier = Modifier.fillMaxWidth()
+        ) { Text("Search: opentelemetry") }
+
+        Button(
+            onClick = { navController.navigate("search") },
+            modifier = Modifier.fillMaxWidth()
+        ) { Text("Search (no query)") }
+
+        Button(
+            onClick = { navController.navigate("confirm_dialog") },
+            modifier = Modifier.fillMaxWidth()
+        ) { Text("Show Dialog") }
+
+        Button(
+            onClick = { navController.navigate("nested_tabs") },
+            modifier = Modifier.fillMaxWidth()
+        ) { Text("Nested Tabs (own NavHost)") }
+
+        OutlinedButton(
+            onClick = onBack,
+            modifier = Modifier.fillMaxWidth()
+        ) { Text("Back to Main Menu") }
+    }
+}
+
+@Composable
+fun SettingsScreen(navController: NavController) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp),
+        verticalArrangement = Arrangement.Center,
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Text("Settings", style = MaterialTheme.typography.h5)
+        Spacer(modifier = Modifier.height(16.dp))
+        Button(onClick = { navController.popBackStack() }) { Text("Back") }
+    }
+}
+
+@Composable
+fun ProfileScreen(userId: String, navController: NavController) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp),
+        verticalArrangement = Arrangement.Center,
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Text("Profile", style = MaterialTheme.typography.h5)
+        Text("User ID: $userId", style = MaterialTheme.typography.body1)
+        Spacer(modifier = Modifier.height(16.dp))
+        Button(onClick = { navController.popBackStack() }) { Text("Back") }
+    }
+}
+
+@Composable
+fun SearchScreen(query: String, navController: NavController) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp),
+        verticalArrangement = Arrangement.Center,
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Text("Search", style = MaterialTheme.typography.h5)
+        Text(
+            text = if (query.isNotEmpty()) "Query: $query" else "No query provided",
+            style = MaterialTheme.typography.body1
+        )
+        Spacer(modifier = Modifier.height(16.dp))
+        Button(onClick = { navController.popBackStack() }) { Text("Back") }
+    }
+}
+
+@Composable
+fun NestedTabsScreen(parentNavController: NavController) {
+    val nestedNavController = rememberNavController()
+
+    Column(modifier = Modifier.fillMaxSize()) {
+        Text(
+            "Nested Tabs (separate NavController)",
+            style = MaterialTheme.typography.h6,
+            modifier = Modifier.padding(16.dp)
+        )
+
+        val currentRoute = nestedNavController.currentBackStackEntryAsState().value
+            ?.destination?.route
+
+        Row(
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(horizontal = 16.dp),
+            horizontalArrangement = Arrangement.spacedBy(8.dp)
+        ) {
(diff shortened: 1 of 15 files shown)


integration/navigation/build.gradle.kts
```diff
@@ -34,6 +34,8 @@ dependencies {
 
     implementation(Dependencies.Android.fragmentKtx)
 
+    compileOnly(Dependencies.Android.navigationRuntime)
```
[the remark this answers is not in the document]
tonzhan2: It wouldn't crash in practice, Android runtime has lazy class resolution.

Added a check anyways though, `ComposeNavigationTracker` is now only created if the navigation library is on the classpath. Apps without it get a debug log and Compose tracking is simply disabled. `Navigation.registerNavController() `already handles the null tracker case gracefully with a warning.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
```diff
@@ -107,6 +117,34 @@ internal class ScreenChangeDetector(private val eventEmitter: NavigationEventEmi
...
+     * highest-priority screen name and emits the event with optional [attributes].
+     * Associates the route with the currently resumed activity so it can be cleared
+     * when a different activity resumes.
+     *
+     * Deduplicates by both screen name and attributes so navigating the same route template
+     * with different arguments (e.g. profile/user_123 → profile/user_456) still emits.
+     */
+    fun onComposeRouteChanged(screenName: String, attributes: Attributes = Attributes.empty()) {
+        lastComposeRouteName = screenName
+        composeRouteActivityName = lastResumedActivityName
```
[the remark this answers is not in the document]
tonzhan2: Added an orphan adoption check: if onActivityResumed sees an active compose route with no owner, it assigns the current activity. This ensures the cleanup condition works correctly when a different activity resumes later.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -141,6 +155,13 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
         application.registerActivityLifecycleCallbacks(fragmentActivityCallback)
     }
 
+    private fun isNavigationAvailable(): Boolean = try {
+        Class.forName("androidx.navigation.NavController")
+        true
```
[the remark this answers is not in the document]
tonzhan2: Updated the guard to check for NavDestination.getRoute() (the route property getter, added in 2.4.0) instead of just NavController class presence. This catches both "no navigation library" and "pre 2.4.0 navigation library" cases. Uses a broad Exception catch to handle both ClassNotFoundException and NoSuchMethodException.

tonzhan2:
@codex review

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/Navigation.kt
```diff
@@ -30,11 +34,37 @@ class Navigation internal constructor() {
...
+    /**
+     * Register a [NavController] for automatic Compose navigation tracking.
+     *
+     * The SDK attaches an [NavController.OnDestinationChangedListener] to track route changes.
+     * Only one NavController can be tracked at a time; registering a new one replaces the previous.
+     * Passing the same instance again is a no-op.
+     *
+     * Call this after [NavController] is created (e.g. in your Activity's `setContent` block).
+     */
+    fun registerNavController(navController: NavController) {
```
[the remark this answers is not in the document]
tonzhan2: This is safe on android runtime, method parameter types are resolved lazily, not at class load time

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationEventProcessor.kt
```diff
@@ -0,0 +1,30 @@
...
+ * Processes navigation events before they are emitted.
+ *
+ * Use to transform screen names, filter events, or modify attributes.
+ * The [NavigationEvent] passed to [process] is mutable: modify [NavigationEvent.screenName]
+ * or [NavigationEvent.attributes] as needed.
+ *
+ * Return `true` to emit the event, or `false` to suppress it.
+ */
+fun interface NavigationEventProcessor {
+    fun process(event: NavigationEvent): Boolean
```
TomasChladekSL: Function should looks like `fun process(event: NavigationEvent): NavigationEvent?`

app/src/main/java/com/splunk/app/ui/compose/ComposeNavigationActivity.kt
```diff
@@ -0,0 +1,309 @@
...
+import androidx.navigation.compose.NavHost
+import androidx.navigation.compose.composable
+import androidx.navigation.compose.currentBackStackEntryAsState
+import androidx.navigation.compose.dialog
+import androidx.navigation.compose.rememberNavController
+import androidx.navigation.navArgument
+import com.splunk.rum.integration.agent.api.SplunkRum
+import com.splunk.rum.integration.navigation.extension.navigation
+
+class ComposeNavigationActivity : ComponentActivity() {
```
TomasChladekSL, line 55: The formatting of the AI-generated code is often inconsistent. Please conduct a line-by-line review to format the code appropriately. These two lines serve as an example of the current state:

`topBar = { TopAppBar(title = { Text("Compose Navigation") }) }`

`Button(onClick = { navController.popBackStack() }) { Text("Back") }`

Definition of Done: Standardized and consistent code formatting is applied across the entire file.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetector.kt
```diff
@@ -38,13 +39,15 @@ internal class ScreenChangeDetector(private val eventEmitter: NavigationEventEmi
...
     private var lastResumedActivityName: String? = null
     private var lastResumedFragmentName: String? = null
+    private var lastComposeRouteName: String? = null
+    private var composeRouteActivityName: String? = null
 
     /**
-     * Current visible screen name: fragment if any, else activity.
+     * Current visible screen name: Compose route > Fragment > Activity.
      */
     private fun getCurrentVisibleScreenName(): String? {
```
TomasChladekSL: `return lastComposeRouteName ?: lastResumedFragmentName ?: lastResumedActivityName`

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationEvent.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Represents a navigation event that can be inspected and modified by a [NavigationEventProcessor]
+ * before it is emitted.
+ *
+ * @property screenName The screen name for this event. Modify to rename the screen.
+ * @property attributes Mutable map of event attributes. Modify to add, remove, or change attributes.
+ * @property sourceType The origin of this navigation event.
+ */
+class NavigationEvent(var screenName: String, val attributes: MutableMap<String, String>, val sourceType: SourceType) {
```
TomasChladekSL: Should not be `screenName`, but only `name`.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -72,18 +73,21 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
         Logger.d(TAG, "onInstall")
         if (!moduleConfiguration.isEnabled) {
             Navigation.instance.listener = null
+            Navigation.instance.composeTracker?.unregisterCurrent()
```
TomasChladekSL: Does this perform any action? No `NavController` will be registered during the installation phase.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -53,9 +53,11 @@ object Dependencies {
 
         private const val annotationVersion = "1.9.1"
         private const val fragmentKtxVersion = "1.3.3"
+        private const val navigationVersion = "2.4.0"
```
TomasChladekSL, line 56: We should ensure compatibility with all NavController versions. Please verify that your solution works.

TomasChladekSL (changes requested):

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
tonzhan2: It was a defensive cleanup, but can be removed. I dont think there are cases where the module could get reinstalled while a previous NavController is still active

buildSrc/src/main/kotlin/Dependencies.kt
tonzhan2: I responded on slack earlier. But the sample app is has ben tested with versions `2.5.3` `2.4.0` and `2.8.9`. So I believe it should cover things pretty well. More detailed comprehensive testing for this, and all other lifecycle/navigation related features will be done as part of the testing ticket and bug bash

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTracker.kt
```diff
@@ -0,0 +1,154 @@
...
+
+    fun unregisterCurrent() {
+        registeredController?.get()?.let { controller ->
+            activeListener?.let { controller.removeOnDestinationChangedListener(it) }
+        }
+        registeredController = null
+        activeListener = null
+        screenChangeDetector.clearComposeRoute()
+    }
+
```
surbhiia, line 74: These two functions perform similar operations, can a third method be extracted which contains the common lines of code and called from both these places?

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -103,6 +106,16 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
...
         }
 
+        if (isNavigationAvailable()) {
+            Navigation.instance.composeTracker = ComposeNavigationTracker(
+                screenChangeDetector = detector,
+                processor = moduleConfiguration.navigationEventProcessor
+            )
+            Logger.d(TAG, "ComposeNavigationTracker initialized")
+        } else {
+            Logger.d(TAG, "androidx.navigation not found on classpath, Compose navigation tracking disabled")
```
surbhiia: The log could be confusing, we should provide the full detail:

"androidx.navigation.NavDestination.getRoute not found on classpath, Compose navigation tracking disabled")

integration/navigation/build.gradle.kts
surbhiia, line 37: @tonzhan2 Do ensure we have this scenario as one of our test cases for bug bash/e2e tests - We should make sure that if an app does not already have `androidx.navigation:navigation-runtime dependency` - we do not cause any crashes. Thanks!

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/NavigationModuleIntegration.kt
```diff
@@ -141,6 +154,14 @@ internal object NavigationModuleIntegration : ModuleIntegration<NavigationModule
         application.registerActivityLifecycleCallbacks(fragmentActivityCallback)
     }
 
+    private fun isNavigationAvailable(): Boolean = try {
+        Class.forName("androidx.navigation.NavDestination")
+            .getMethod("getRoute")
+        true
```
surbhiia, line 163: Question - @tonzhan2 would we not support customers using earlier versions of `androidx.navigation:navigation-runtime` than 2.4.0? Shouldn't we have a fallback code path for customers using earlier versions for wherever `NavDestination.route` code is used?

tonzhan2: No we shouldn't, `navigation-runtime:2.4.0` is the minimum version that supports the route property on `NavDestination`.

this covers every app that uses compose navigation since `androidx.navigation:navigation-compose` itself was introduced in 2.4.0.

There is no such thing as compose navigation on an app before `2.4.0` as it didn’t exist yet. Supporting `2.4.0+`means we support all compose navigation apps

integration/navigation/build.gradle.kts
surbhiia, line 37: One more thing to test for `compileOnly` here is that - if R8 shrinking is used and `minifyEnabled = true` and app has NO `androidx.navigation:navigation-runtime` dependency - there are no missing class errors and if there are we might need consumer R8/proguard rules like `-dontwarn` for the missing class errors.

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTracker.kt
```diff
@@ -0,0 +1,151 @@
...
+                if (value != null) {
+                    builder.put(AttributeKey.stringKey(key), value.toString())
+                }
+            }
+        }
+        destination.parent?.route?.let {
+            builder.put(AttributeKey.stringKey(ATTR_NAV_GRAPH), it)
+        }
+        return builder.build()
+    }
```
surbhiia: 1. Can these two functions be made concise too by extracting a common function if possible? 

2. The arguments Bundle can hold primitives, strings, and sometimes custom values. We currently stringify everything with `.toString()`. That’s simple, but for types that still use the default `Object.toString()` we’d emit values like `com.example.Foo@1a2b3c`, which isn’t useful in logs and can be noisy or misleading. 

There isn't a better way also to handle this and keep things simple. Perhaps we can add it to our documentation so we're pro-actively making customers aware:

Option A — short “Important”

> Important: For automatic Compose navigation capture, route arguments are taken from the NavController Bundle and recorded as strings (via string conversion). For predictable, useful screen attributes, use primitive or string arguments (for example stable IDs). Complex or custom types may not stringify to meaningful values and may be unsuitable for telemetry unless you transform or remove them with a NavigationEventProcessor.

Option B — tighter “Note”

> Note: Navigation arguments are not sent as typed values. The SDK stringifies each argument for RUM. Treat route args as string-safe telemetry inputs (IDs, flags), or use NavigationEventProcessor to control what is emitted.

buildSrc/src/main/kotlin/Dependencies.kt
surbhiia, line 56: @tonzhan2 I recently learnt that `androidx.navigation-runtime` has a dependency on `androidx.fragment` , so here as `androidx.navigation-runtime` has the higher version, gradle will automatically resolve `androidx.fragment`  to whatever is compatible with `androidx.navigation-runtime`. You can as well look at what gradle resolves` androidx.fragment`  finally too and use that here specifically.

tonzhan2: This isnt a concern for us. `navigation-runtime` is `compileOnly`, so its transitive dependencies including fragment are not propagated to consuming apps

Our module already declares its own `implementation(Dependencies.Android.fragmentKtx)`. At runtime, whatever fragment version the consumer's app resolves to is what gets used. Gradles dependency resolution handles this on the consumer side, not ours so we don't need to change anything here

integration/navigation/src/main/java/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTracker.kt
tonzhan2: 1) done
2) I'd go with option B, its more concise and actionable

Its probably not a huge deal as in practice, route arguments in Compose Navigation are almost always primitives/strings anyway. Pretty sure complex types in navigation args are an anti pattern even by googles own guidelines

surbhiia (approved):
Approving the high level code here without going much into how the instrumentation functions. Leaving that to @TomasChladekSL.

TomasChladekSL (approved):

PR #1599, opened 2026-04-15
NO-TICKET: Fix screen.name override

tonzhan2:
Navigation events set `screen.name `on the log record at emit time, but the global attribute processor (`SplunkInternalGlobalAttributeSpanProcessor`) was unconditionally overwriting it at export time with the current global value. When manual and automatic tracking fired in rapid succession, both exported spans would show the latest` screen.name `, when the manual one should have screen.name of whatever the customer set.

This is a problem when the customer calls a manual track API passing in a DIFFERENT screenName argument, causing both screens in the session view to have the same `screen.name`

The fix adds a guard in` onStart()` that skips setting` screen.name` when all three conditions are met:
- The attribute being applied is` screen.name`
- The span is a navigation span `app.ui.navigation`
- The span already has a `screen.name` value carried from the log record

This only affects navigation spans

[Bullet list of the changes or enhancements made in this PR.]

[Provide any necessary context or background information that reviewers should be aware of.]

[Link to any relevant issues or discussions, if applicable.]

### Checklist

### Generative AI usage

### How to Test These Changes

build sample app
add/uncomment existing manual track call, make sure screenName argument is different to the actual screen.name
run app
verify both manual and auto tracked screens show up in session view with correct screen names

### DIFF
--- a/common/otel/src/main/java/com/splunk/rum/common/otel/internal/GlobalRumConstants.kt
+++ b/common/otel/src/main/java/com/splunk/rum/common/otel/internal/GlobalRumConstants.kt
@@ -60,6 +60,11 @@ object GlobalRumConstants {
     val PREVIOUS_SESSION_ID_KEY: AttributeKey<String> = AttributeKey.stringKey("session.previous_id")
 
     /**
+     * Log event names.
+     */
+    const val NAVIGATION_LOG_EVENT_NAME = "app.ui.navigation"
+
+    /**
      * Screen values and attribute keys.
      */
     const val DEFAULT_SCREEN_NAME = "unknown"
--- a/common/otel/src/main/java/com/splunk/rum/common/otel/internal/GlobalRumConstants.kt
+++ b/common/otel/src/main/java/com/splunk/rum/common/otel/internal/GlobalRumConstants.kt
@@ -62,7 +62,7 @@ object GlobalRumConstants {
     /**
      * Log event names.
      */
-    const val NAVIGATION_LOG_EVENT_NAME = "app.ui.navigation"
+    const val NAVIGATION_EVENT_NAME = "app.ui.navigation"
 
     /**
      * Screen values and attribute keys.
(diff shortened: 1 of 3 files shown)


aditi-s3:
@codex review

common/otel/src/main/java/com/splunk/rum/common/otel/internal/GlobalRumConstants.kt
```diff
@@ -59,6 +59,11 @@ object GlobalRumConstants {
     val SESSION_ID_KEY: AttributeKey<String> = AttributeKey.stringKey("session.id")
     val PREVIOUS_SESSION_ID_KEY: AttributeKey<String> = AttributeKey.stringKey("session.previous_id")
 
+    /**
+     * Log event names.
+     */
+    const val NAVIGATION_LOG_EVENT_NAME = "app.ui.navigation"
+
```
aditi-s3, line 66: Nit: NAVIGATION_EVENT_NAME is better

surbhiia:
Question - Does this mean the `screen.name` set via the span processor is not always reliable? 

When you have already set `screen.name `in the global context when navigation occurs and before emitting the navigation event, shouldn't the global `screen.name ` set via the span processor be reliable. You already have `navigation.name` in the navigation signal set from the navigation instrumentation directly.

Even when navigation is happening in quick succession like the case you're testing, global `screen.name` should represent correct state.

tonzhan2:
> Question - Does this mean the `screen.name` set via the span processor is not always reliable?
> 
> When you have already set `screen.name `in the global context when navigation occurs and before emitting the navigation event, shouldn't the global `screen.name ` set via the span processor be reliable. You already have `navigation.name` in the navigation signal set from the navigation instrumentation directly.
> 
> Even when navigation is happening in quick succession like the case you're testing, global `screen.name` should represent correct state.

Yes, technically it's not always reliable as a representation of "what screen was the user on when this event occurred." It's reliable as a representation of "what screen is the user on when this span is created during batch conversion." Hence this PR

Youre right that` navigation.name `already captures the correct per event value, and `screen.name` is intended as a global state attribute. The issue is that the global state at batch conversion time doesnt match the state at emit time when events fire back to back, the emitter explicitly sets` screen.name` on the log record, but the span processor overwrites it during the deferred log to span conversion process

In practice this only happens in the back to back manual + auto tracking case.

If the backend uses `navigation.name` for session grouping on `app.ui.navigation` events, this SDK change isn't needed. But currently the backend groups by `screen.name` when it differs from` last.screen.name`, which makes the manual event indistinguishable from the auto tracked one in the session view.

PR #1604, opened 2026-04-17
NO-TICKET: Fix ignored fragment behaviour

tonzhan2:
When a tracked fragment paused and an ignored fragment resumed (e.g. @NavigationElement(isIgnored = true)), the deferred emit from onFragmentPaused would fire and fall back to the Activity name, producing a false navigation event. 

This fix cancels the pending deferred emit whenever any fragment resumes, so ignored fragments no longer trigger an Activity fallback. The "fragment removed without replacement" fallback is unaffected since no replacement fragment resumes to cancel it.

### Checklist

### Generative AI usage

### How to Test These Changes

- add isIgnored annotation to a fragment, ie crashReportsFragment
- Build app
- navigate to and back
- observe zero navigation events when tapping "Crash Reports" and pressing back

### DIFF
--- a/integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
+++ b/integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
@@ -170,6 +170,15 @@ class ScreenChangeDetectorTest {
     }
 
     @Test
+    fun `recordEmittedScreen prevents duplicate emission from automatic Compose tracking`() {
+        detector.recordEmittedScreen("home")
+
+        detector.onComposeRouteChanged("home")
+
+        assertTrue(exportedLogs.isEmpty())
+    }
+
+    @Test
     fun `activity becomes visible screen when fragment is paused`() {
         val activity = activityController.create().get()
         val menu = MenuFragment()
--- a/integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
+++ b/integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
@@ -323,7 +323,7 @@ class ScreenChangeDetectorTest {
     }
 
     @Test
-    fun `navigating to ignored fragment does not emit spurious activity event`() {
+    fun `navigating to ignored fragment does not emit false activity event`() {
         val activity = activityController.create().get()
         val menu = MenuFragment()
         val ignored = IgnoredFragment()
--- a/integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
+++ b/integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
@@ -323,7 +323,7 @@ class ScreenChangeDetectorTest {
     }
 
     @Test
-    fun `navigating to ignored fragment does not emit false activity event`() {
+    fun `navigating to ignored fragment does not emit a false activity event`() {
         val activity = activityController.create().get()
         val menu = MenuFragment()
         val ignored = IgnoredFragment()
@@ -363,6 +363,43 @@ class ScreenChangeDetectorTest {
 
         assertEquals(1, exportedLogs.size)
         assertEquals("Menu", exportedLogs[0].attributes.get(GlobalRumConstants.SCREEN_NAME_KEY))
+    }
+
+    @Test
+    fun `back to back fragment pauses do not emit duplicate events`() {
+        val activity = activityController.create().get()
+        val parent = MenuFragment()
+        val child = CrashReportsFragment()
+
+        detector.onActivityResumed(activity)
+        detector.onFragmentResumed(parent)
+        detector.onFragmentResumed(child)
+        shadowOf(Looper.getMainLooper()).idle()
+        assertEquals(2, exportedLogs.size)
+
+        detector.onFragmentPaused(child)
+        detector.onFragmentPaused(parent)
+        shadowOf(Looper.getMainLooper()).idle()
+
+        assertEquals(3, exportedLogs.size)
+        assertEquals("Main", exportedLogs[2].attributes.get(GlobalRumConstants.SCREEN_NAME_KEY))
+    }
+
+    @Test
+    fun `clearComposeRoute does not re-emit when fragment has same name`() {
+        val activity = activityController.create().get()
+
+        detector.onActivityResumed(activity)
+        shadowOf(Looper.getMainLooper()).idle()
+        assertEquals(1, exportedLogs.size)
+
+        detector.onComposeRouteChanged("Main")
+        assertEquals(2, exportedLogs.size)
+
+        detector.clearComposeRoute()
+        shadowOf(Looper.getMainLooper()).idle()
+
+        assertEquals(2, exportedLogs.size)
     }
 
     @NavigationElement(name = "Main")
--- a/integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
+++ b/integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
@@ -313,6 +313,49 @@ class ScreenChangeDetectorTest {
         assertEquals("Menu", exportedLogs[0].attributes.get(GlobalRumConstants.SCREEN_NAME_KEY))
     }
 
+    @Test
+    fun `navigating to ignored fragment does not emit spurious activity event`() {
+        val activity = activityController.create().get()
+        val menu = MenuFragment()
+        val ignored = IgnoredFragment()
+
+        detector.onActivityResumed(activity)
+        detector.onFragmentResumed(menu)
+        shadowOf(Looper.getMainLooper()).idle()
+        assertEquals(1, exportedLogs.size)
+        assertEquals("Menu", exportedLogs[0].attributes.get(GlobalRumConstants.SCREEN_NAME_KEY))
+
+        detector.onFragmentPaused(menu)
+        detector.onFragmentResumed(ignored)
+        shadowOf(Looper.getMainLooper()).idle()
+
+        assertEquals(1, exportedLogs.size)
+    }
+
+    @Test
+    fun `returning from ignored fragment to tracked fragment emits correctly`() {
+        val activity = activityController.create().get()
+        val menu = MenuFragment()
+        val ignored = IgnoredFragment()
+
+        detector.onActivityResumed(activity)
+        detector.onFragmentResumed(menu)
+        shadowOf(Looper.getMainLooper()).idle()
+        assertEquals(1, exportedLogs.size)
+
+        detector.onFragmentPaused(menu)
+        detector.onFragmentResumed(ignored)
+        shadowOf(Looper.getMainLooper()).idle()
+        assertEquals(1, exportedLogs.size)
+
+        detector.onFragmentPaused(ignored)
+        detector.onFragmentResumed(menu)
+        shadowOf(Looper.getMainLooper()).idle()
+
+        assertEquals(1, exportedLogs.size)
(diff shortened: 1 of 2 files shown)


integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
```diff
@@ -313,6 +322,49 @@ class ScreenChangeDetectorTest {
...
+        detector.onFragmentPaused(menu)
+        detector.onFragmentResumed(ignored)
+        shadowOf(Looper.getMainLooper()).idle()
+        assertEquals(1, exportedLogs.size)
+
+        detector.onFragmentPaused(ignored)
+        detector.onFragmentResumed(menu)
+        shadowOf(Looper.getMainLooper()).idle()
+
+        assertEquals(1, exportedLogs.size)
```
surbhiia, line 364: Exported log size here should be two? Two events - First visit to Menu and second visit to Menu after coming back from ignored fragment

integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
```diff
@@ -313,6 +322,49 @@ class ScreenChangeDetectorTest {
         assertEquals("Menu", exportedLogs[0].attributes.get(GlobalRumConstants.SCREEN_NAME_KEY))
     }
 
+    @Test
+    fun `navigating to ignored fragment does not emit spurious activity event`() {
```
surbhiia: nit - can we use a generic word like "false" instead of "spurious" here?

surbhiia:
We need to add the following tests to `ScreenCahngeDetectorTest` to guage the accuracy of the logic as I doubt some of these are also places where current logic fails:

1. `lastEmittedScreenName` is shared for Compose and Activity/Fragment — a real navigation can be skipped if the name string is unchanged when the source changes (e.g. Compose "X" → fragment "X" after clearComposeRoute()).
2. Test for `clearComposeRoute()` function. Should it also reset `lastEmittedScreenName` to prevent the issue in test  case 1 above. 
3. Back to back fragment pauses, no onResumed called - `onFragmentPaused` posts a deferred `tryEmitIfChanged`() to the main Handler but only stores the latest runnable in `pendingPauseEmit`; it does not remove previously posted runnables. When nested fragments pause back-to-back (e.g. child then parent) with no intervening `onFragmentResumed`, multiple deferred checks can run, which can produce duplicate or intermediate navigation emissions compared to the intended single deferred emit after a pause.
4. `NavigationEventProcessor` based suppression can cause issues - Suppose the user is on Compose route Home, then navigates to an internal Loading route that your NavigationEventProcessor suppresses (null). The UI is now Loading, but we never call `onComposeRouteChanged`, so the detector still thinks the top screen is Home. If the app then shows a full-screen ErrorRetry fragment, `tryEmitIfChanged()` still prefers the stale Compose name Home over the fragment, so you can miss the ErrorRetry navigation event (or attribute activity to the wrong screen) until some other Compose update or clear happens. 

DOCS Note: 
Our logic depends on name string values. We should ensure that we document it clearly for the customer that how names are derived and that they should keep them unique. Otherwise any two screen navigations (activity -> compose route, activity -> fragment , activity -> activity) in quick succession with same name will not produce two navigation events as they should.

surbhiia:
There's another larger question I have - Should we track ignored activities/fragments/compose routes with `screen.name=ignored` as user is actually on a fully visible screen which customer has chosen to not track but all our agent signals will wrongly report the old `screen.name` until a new non-ignored screen shows up. 

I also think we should do that to prevent issues like this edge case - 

**Emit a navigation event for something that is not what the user sees**
Example: User was on Main + Menu; Main pauses and clears lastResumedActivityName; fragments may still be mid-transition; Compose route might still be non-null. Something then calls tryEmitIfChanged() (fragment resumed, or a deferred Runnable from an earlier onFragmentPaused). The function uses getCurrentVisibleScreenName() right now, not “what Window is focused.” It might resolve to Menu or an old Compose route while the user is actually staring at the ignored full-screen activity. If that string also differs from lastEmittedScreenName, you emit a navigation event that reads like a screen change to Menu/Home, even though the visible window is the ignored activity.

integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
tonzhan2: It should be 1. When the user returns from the ignored fragment, `MenuFragment` resumes. But `lastEmittedScreenName` is still "Menu" as it was never changed since the ignored fragment didn't trigger any events. 

So `tryEmitIfChanged()` sees "Menu" == "Menu" and dedupes it. The exported log size correctly stays at 1. There's no "second visit" since the user never left Menu from the SDKs perspective with this ignore flag

tonzhan2:
> There's another larger question I have - Should we track ignored activities/fragments/compose routes with `screen.name=ignored` as user is actually on a fully visible screen which customer has chosen to not track but all our agent signals will wrongly report the old `screen.name` until a new non-ignored screen shows up.
> 
> I also think we should do that to prevent issues like this edge case -
> 
> **Emit a navigation event for something that is not what the user sees** Example: User was on Main + Menu; Main pauses and clears lastResumedActivityName; fragments may still be mid-transition; Compose route might still be non-null. Something then calls tryEmitIfChanged() (fragment resumed, or a deferred Runnable from an earlier onFragmentPaused). The function uses getCurrentVisibleScreenName() right now, not “what Window is focused.” It might resolve to Menu or an old Compose route while the user is actually staring at the ignored full-screen activity. If that string also differs from lastEmittedScreenName, you emit a navigation event that reads like a screen change to Menu/Home, even though the visible window is the ignored activity.


The current design is that when we mark something ignored, its invisible to the SDK entirely. 

The fact that spans will report the old `screen.name` while the user is staring at the ignored screen is part of the design and consistent with the previous agent behaviour, the customer chose isIgnored to opt out of tracking for that screen so we do not update the screen.

Changing this to emit `screen.name`=ignored would be a public behavior change. Not opposed to the idea but we'd need to think about it some more. Definitely a larger design discussion that goes much beyond the scope of this PR fix.

tonzhan2:
> We need to add the following tests to `ScreenCahngeDetectorTest` to guage the accuracy of the logic as I doubt some of these are also places where current logic fails:
> 
> 1. `lastEmittedScreenName` is shared for Compose and Activity/Fragment — a real navigation can be skipped if the name string is unchanged when the source changes (e.g. Compose "X" → fragment "X" after clearComposeRoute()).
> 2. Test for `clearComposeRoute()` function. Should it also reset `lastEmittedScreenName` to prevent the issue in test  case 1 above.
> 3. Back to back fragment pauses, no onResumed called - `onFragmentPaused` posts a deferred `tryEmitIfChanged`() to the main Handler but only stores the latest runnable in `pendingPauseEmit`; it does not remove previously posted runnables. When nested fragments pause back-to-back (e.g. child then parent) with no intervening `onFragmentResumed`, multiple deferred checks can run, which can produce duplicate or intermediate navigation emissions compared to the intended single deferred emit after a pause.
> 4. `NavigationEventProcessor` based suppression can cause issues - Suppose the user is on Compose route Home, then navigates to an internal Loading route that your NavigationEventProcessor suppresses (null). The UI is now Loading, but we never call `onComposeRouteChanged`, so the detector still thinks the top screen is Home. If the app then shows a full-screen ErrorRetry fragment, `tryEmitIfChanged()` still prefers the stale Compose name Home over the fragment, so you can miss the ErrorRetry navigation event (or attribute activity to the wrong screen) until some other Compose update or clear happens.
> 
> DOCS Note: Our logic depends on name string values. We should ensure that we document it clearly for the customer that how names are derived and that they should keep them unique. Otherwise any two screen navigations (activity -> compose route, activity -> fragment , activity -> activity) in quick succession with same name will not produce two navigation events as they should.

1) Not really relevant. This requires a Compose route and a Fragment to have the exact same name string with a clearComposeRoute() at the right time. Compose routes look like `profile/{userId}` or `home/feed` while fragment/activity names are simple class names like MenuFragment or MainActivity. There pretty much zero chance of a real world collision. 

2 & 3) Sure, added

4) Yeah but this is by design, suppression means that we're pretending this navigation event didn't happen. If the user wants different behavior, they should rename rather than suppress. It's working as intended. If we want to change this behavior it would warrant deeper, discussion. Beyond the scope of this PR 

docs note: Sure, we can ask docs to make note of this

SenNeonoveNoci:
@tonzhan2 `pendingPauseEmit` is overwritten on every `onFragmentPaused()`, and `cancelPendingPauseEmit()` removes only the most recent runnable. Can this mean the fix misses nested multi-pause transitions?

Example from AI: _`MainActivity` shows `HomeFragment`, and `HomeFragment` hosts `FeedFragment` via `childFragmentManager` (common for tabs, pagers, wizard flows, or shell + nested content). If the user opens a full-screen ignored fragment such as `CheckoutFragment` or `LoginGateFragment`, Android can dispatch `FeedFragment.onPause()` and then `HomeFragment.onPause()` synchronously in the same transaction, before any `Handler.post { ... }` work runs._

In `ScreenChangeDetector.onFragmentPaused()`, the first pause posts runnable A and the second posts runnable B. When the ignored fragment resumes, `cancelPendingPauseEmit()` removes only B because `pendingPauseEmit` has already been overwritten. Runnable A is still queued and can later emit the activity fallback.

// EDIT: I see that @surbhiia already mentioned this. However, Im not sure it was addressed.

integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
```diff
@@ -313,6 +322,86 @@ class ScreenChangeDetectorTest {
         assertEquals("Menu", exportedLogs[0].attributes.get(GlobalRumConstants.SCREEN_NAME_KEY))
     }
 
+    @Test
+    fun `navigating to ignored fragment does not emit a false activity event`() {
```
TomasChladekSL, line 326: For these tests, it is better to use Espresso tests instead of hardcoded behavior. You can take inspiration from our `SmartCinema` project (`smartlook-mobile-demo-android`).

tonzhan2: We can add espresso/instrumented tests, could be valuable for e2e validation of the navigation pipeline. For this PR the unit tests cover the specific callback ordering edge cases (nested pauses, ignored fragments) that are easier to reproduce deterministically with robolectric than with espresso navigation.

Dont want to have too much scope creep before the incoming release. Should be tracked as an additional item

surbhiia:
Regarding your below response in [this comment](https://github.com/signalfx/splunk-otel-android/pull/1604#issuecomment-4272195126): 
> The current design is that when we mark something ignored, it's invisible to the SDK entirely.
>
> The fact that spans will report the old screen.name while the user is staring at the ignored screen is part of the design and consistent with the previous agent behaviour, the customer chose isIgnored to opt out of tracking for that screen so we do not update the screen.
> 
> Changing this to emit screen.name=ignored would be a public behavior change. Not opposed to the idea but we'd need to think about it some more. Definitely a larger design discussion that goes much beyond the scope of this PR fix.

Sure, I'd leave that upto you to bring up this discussion and validate if there are no edge case bugs that need this as a solution. 

--------------------------------------------------
Regarding your response to point 4 mentioned in [this comment](https://github.com/signalfx/splunk-otel-android/pull/1604#issuecomment-4271796759) - 

> Yeah but this is by design, suppression means that we're pretending this navigation event didn't happen. If the user wants different behavior, they should rename rather than suppress. It's working as intended. If we want to change this behavior it would warrant deeper, discussion. Beyond the scope of this PR

I'd again leave it upto you to ensure suppression doesn't lead to any bugs.

surbhiia (approved):

integration/navigation/src/test/java/com/splunk/rum/integration/navigation/automatic/ScreenChangeDetectorTest.kt
SenNeonoveNoci: I have create a followup Jira ticket.

PR #1620, opened 2026-05-05
DEMRUM-4857: Resolve decoupled endpoint config 

tonzhan2:
## Title: implement atomic endpoint configuration storage to prevent rumAccessToken crash

**Description**

This is the root cause fix for the `IllegalStateException` in `AuthHeaderBuilder`. The hotfix in PR #1560 replaced the throw with a `Logger.w()` to prevent the crash, but left the underlying storage inconsistency unresolved. 

If the URL in storage is left without the token, causing background upload jobs to send unauthenticated requests. With this atomic storage, the config is either fully present or fully absent such that no partial state is possible.

- Store `tracesBaseUrl`, `logsBaseUrl`, and `rumAccessToken` as a single atomic JSON blob (`StoredEndpointConfig`) instead of three separate preference keys, eliminating the inconsistent state that caused the original `IllegalStateException` crash
- Make all upload jobs (`UploadOtelSpanDataJob`, `UploadOtelLogRecordDataJob`, `UploadSessionReplayDataJob`) to read the atomic config once and gracefully exit if absent
- Refactor `AuthHeaderBuilder` to accept `StoredEndpointConfig` directly so the URL and token are guaranteed to come from the same atomic read. Empty tokens now treated as absent
- Update `ConfigurationManager` and `AgentPreferences` to write endpoint config atomically in a single call
- Add transparent migration from legacy separate keys to the new atomic format on first read (idempotent and safe under concurrent access) to prevent potential data loss on first launch in case of upgrade
- Removed the old individual `readTracesBaseUrl()`, `readLogsBaseUrl()`, `readRumAccessToken() `methods in `IAgentStorage`- Add and update unit tests

### Checklist

### Generative AI usage

### How to Test These Changes

-Run unit tests
- Install the sample app with the current agent (without these changes) and trigger some telemetry so endpoint config is written to storage using the old separate keys, update to this branch and relaunch the app, confirm telemetry uploads still succeed


common/storage/src/main/java/com/splunk/rum/common/storage/AgentStorage.kt
```diff
@@ -93,36 +93,52 @@ class AgentStorage(context: Context) : IAgentStorage {
...
+    /**
+     * One-time migration from separate legacy keys to atomic ENDPOINT_CONFIG.
+     * Concurrent calls are safe: migration is idempotent (same config written, legacy keys cleared).
+     */
+    private fun migrateFromLegacyKeys(): StoredEndpointConfig? {
+        val tracesUrl = preferences.getString(TRACES_BASE_URL) ?: return null
+        val config = StoredEndpointConfig(
+            tracesBaseUrl = tracesUrl,
+            logsBaseUrl = preferences.getString(LOGS_BASE_URL),
+            rumAccessToken = preferences.getString(RUM_ACCESS_TOKEN)
```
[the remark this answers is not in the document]
tonzhan2: Extremely narrow and unlikely race condition that could only happen on first launch after upgrade, inconsequential

surbhiia:
We should consider renaming `logsBaseUrl` to `sessionReplayLogsBaseUrl` so in future we don't confuse it with a general `logsBaseUrl`

common/storage/src/main/java/com/splunk/rum/common/storage/StoredEndpointConfig.kt
```diff
@@ -0,0 +1,51 @@
...
+
+        fun fromJson(json: String): StoredEndpointConfig? = try {
+            val obj = JSONObject(json)
+            StoredEndpointConfig(
+                tracesBaseUrl = obj.getString(KEY_TRACES_URL),
+                logsBaseUrl = obj.optString(KEY_LOGS_URL).takeIf { !obj.isNull(KEY_LOGS_URL) },
+                rumAccessToken = obj.optString(KEY_TOKEN).takeIf { !obj.isNull(KEY_TOKEN) }
+            )
+        } catch (_: Exception) {
+            null
```
surbhiia, line 57: We should log at least the class of exception that occurred, useful for debugging purposes.

common/storage/src/main/java/com/splunk/rum/common/storage/AgentStorage.kt
```diff
@@ -93,36 +93,52 @@ class AgentStorage(context: Context) : IAgentStorage {
...
+        preferences.putString(ENDPOINT_CONFIG, config.toJson())
+        clearLegacyEndpointKeys()
+    }
+
+    override fun readEndpointConfig(): StoredEndpointConfig? {
+        val json = preferences.getString(ENDPOINT_CONFIG)
+        if (json != null) {
+            val config = StoredEndpointConfig.fromJson(json)
+            if (config == null) {
+                Logger.w(TAG, "readEndpointConfig() stored JSON is corrupted, discarding: $json")
```
surbhiia: We should not log entire `$json` as it can expose the `rumAccessToken`. Just mentioning corrupted is fine - as we'd be logging class of exception in `StoredEndpointConfig.fromJson` (as requested in another review comment), these two logs will help.

surbhiia:
Add a test case to `AgentStorageTest` to test the migration of legacy configs.

common/storage/src/main/java/com/splunk/rum/common/storage/AgentStorage.kt
```diff
@@ -93,36 +93,52 @@ class AgentStorage(context: Context) : IAgentStorage {
...
         }
 
-    override fun writeTracesBaseUrl(value: String) {
-        preferences.putString(TRACES_BASE_URL, value)
+    override fun writeEndpointConfig(config: StoredEndpointConfig) {
+        preferences.putString(ENDPOINT_CONFIG, config.toJson())
+        clearLegacyEndpointKeys()
+    }
+
+    override fun readEndpointConfig(): StoredEndpointConfig? {
```
surbhiia, line 101: I have doubts about concurrency here, can we add test cases to ensure that code is apt to cater to these scenarios and also update the code to guarantee no concurrency issues - as test cases might not be able to produce the rare race conditions that might happen in production:

- If suppose migration hasn't happened - it's the first launch after code update - both the span and log exporter call this method, call from span exporter goes in migrate function and is at the point where trace url is removed, and log exporter call now goes into the migrate function as config was not set before it read it and it sees an empty trace url and returns null. This would be wrong behavior as this log won't be exported as it received null config. 

So, can we do something to guarantee that code will perform correctly even if any rare race condition were to occur.

tonzhan2:
> Add a test case to `AgentStorageTest` to test the migration of legacy configs.

Originally had them but decided to remove them since they required an extra test helper method and its just for a one time temporary migration case so wasn't sure it was needed in the code long term. But I put them back for now. This and the migration logic are probably things we can remove for simplicity much further down the line once the latest agents are more widely adopted

common/storage/src/main/java/com/splunk/rum/common/storage/AgentStorage.kt
tonzhan2: sure, i wrapped that method in a synchronized block and so its more robust now. 

But even without it, this is an extremely rare case. It requires the first launch after an agent upgrade (so only one time per device), and two background threads hitting `readEndpointConfig()` at the same time before `SplunkRum.install()` writes the new atomic config, and also interleaving at the exact window between reads and clears. 

Even if it happened, the `install()` overwrites it with correct values and things are corrected very shortly after. Theres no chance of crash or any real damage, just maybe an upload attempt failing in that specific window, if this were all to line up and happen

surbhiia: Looks good! :)

surbhiia (approved):

PR #1624, opened 2026-05-13
DEMRUM-5293: Fix instrumented tests job hanging on emulator shutdown

tonzhan2:
### Description
Re enable the instrumented_tests job that was disabled due to the emulator runner hanging indefinitely during cleanup. 

The root cause is some orphaned crashpad_handler processes that prevent the process tree from exiting (ReactiveCircus/android-emulator-runner#385).

Changes:
- Kill crashpad_handler processes inside the script before the action's cleanup phase runs to avoid the deadlock
- Add timeout-minutes: 30 as a fail fast safety net (tests take about 14min)

note: The pkill uses -f to match the full command line (the process name exceeds Linux's 15 character comm limit) and a [c] character class trick to avoid matching its own shell process

### Checklist

### Generative AI usage

- [x] GAI was not used (or, no additional notation is required)

### How to Test These Changes

- Verify tests succeed

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -44,7 +44,7 @@ jobs:
         run: ./gradlew testDebugUnitTest
   instrumented_tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     strategy:
       matrix:
         api-level: [ 29 ]
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -47,7 +47,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        api-level: [ 29 ]
+        api-level: [ 24, 29 ]
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - name: Set up JDK 17 for running Gradle
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -42,51 +42,59 @@ jobs:
           java-version: 17
       - name: Run Unit Tests
         run: ./gradlew testDebugUnitTest
-  # Temporarily disabled because instrumented tests are not working correctly in the PR workflow.
-  # instrumented_tests:
-  #   runs-on: ubuntu-latest
-  #   strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -42,51 +42,59 @@ jobs:
...
-  #       uses: reactivecircus/android-emulator-runner@v2
-  #       with:
-  #         api-level: ${{ matrix.api-level }}
-  #         force-avd-creation: false
-  #         emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-  #         disable-animations: true
-  #         script: ./gradlew connectedDebugAndroidTest
+  instrumented_tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
OndrejSynakSL: I would increase it to at least hour

.github/workflows/pr.yaml
```diff
@@ -42,51 +42,59 @@ jobs:
...
-  #         force-avd-creation: false
-  #         emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-  #         disable-animations: true
-  #         script: ./gradlew connectedDebugAndroidTest
+  instrumented_tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        api-level: [ 29 ]
```
surbhiia: Do we want to also run for another API level between 21 - 26? Perhaps 24 (that will work without having to set `forceEnableOnLowerApi = true`)?

tonzhan2: Sure, it will double the emulator runtime though

surbhiia: Do you not need to setup desugaring for API 24 for these tests?

tonzhan2: Desugaring is already enabled for the app module via `ConfigAndroidApp`  (which sets isCoreLibraryDesugaringEnabled = true and adds the coreLibraryDesugaring dependency).

The other module that runs tests (integration:agent:internal) doesn't use any APIs that need it. If anything breaks on API 24, the CI run will catch it.

surbhiia (approved):

PR #1673, opened 2026-08-05
NO-TICKET: Streamline and defer navigation module init

tonzhan2:
## Summary

Removes all reflection from the navigation module's install path to reduce main thread activity upon install

### Context

The original `isNavigationAvailable()` ran `Class.forName().getMethod()` synchronously on the main thread during `SplunkRum.install()`. On pressured devices this can trigger a dex I/O, `NavDestination.<clinit>`, and a reflective method hierarchy walk, contributing to main thread activity and potential ANRs for heavily loaded, non performant apps

### Details

- **`NavigationModuleIntegration.onInstall()`**: Removed the `isNavigationAvailable()` check entirely (`Class.forName("NavDestination").getMethod("getRoute")`) and the eager `ComposeNavigationTracker` creation. Instead, stores the tracker's dependencies (`ScreenChangeDetector`, `NavigationEventProcessor`) via `Navigation.setTrackerConfig()` for deferred use.

- **`Navigation.registerNavController()`**: Creates the `ComposeNavigationTracker` lazily on the customer's first call. Since this happens in UI code (e.g. `setContent`) after `SplunkRum.install()` completes, the work is no longer competing for the ANR budget during startup. No reflection is performed here as the tracker is just an object allocation

- **`ComposeNavigationTracker.handleDestinationChanged()`**: Added a `LinkageError` catch around `destination.route` access. If the app is on `androidx.navigation` < 2.4.0 (where `getRoute()` doesn't exist), the tracker disables itself on first destination change and logs a warning instead of crashing. This replaces the removed `getMethod("getRoute")` reflection guard with zero upfront cost. `LinkageError` covers `NoSuchMethodError`, `NoClassDefFoundError`, `IncompatibleClassChangeError`, etc.

- Fragment/Activity automatic tracking is unchanged
- `isAutomatedTrackingEnabled` behavior is unchanged (Compose tracking is independent of it, same as before)
- `registerNavController()` called multiple times with the same instance is still a no-op
- Apps that don't use Compose Navigation pay zero cost (tracker is never created)
- Module disabled (`isEnabled = false`) still fully cleans up and logs a warning on `registerNavController()`

### Checklist

### Generative AI usage

### How to Test These Changes

Ensure navigation module and automated tracking including jetpack compose still work

### DIFF
--- a/integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
+++ b/integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
@@ -19,12 +19,15 @@ package com.splunk.rum.integration.navigation
 import androidx.navigation.NavController
 import com.splunk.rum.common.logger.Logger
 import com.splunk.rum.integration.navigation.automatic.ComposeNavigationTracker
+import com.splunk.rum.integration.navigation.automatic.ScreenChangeDetector
 import io.opentelemetry.api.common.Attributes
 
 class Navigation internal constructor() {
 
     internal var listener: Listener? = null
     internal var composeTracker: ComposeNavigationTracker? = null
+    private var pendingDetector: ScreenChangeDetector? = null
+    private var pendingProcessor: NavigationEventProcessor? = null
 
     /**
      * Record a navigation to [screenName] with optional [attributes] (manual tracking).
@@ -44,12 +47,34 @@ class Navigation internal constructor() {
      * Call this after [NavController] is created (e.g. in your Activity's `setContent` block).
      */
     fun registerNavController(navController: NavController) {
-        val tracker = composeTracker
-        if (tracker == null) {
+        val tracker = composeTracker ?: createTrackerIfConfigured() ?: run {
             Logger.w(TAG, "Navigation module not installed. Cannot register NavController.")
             return
         }
         tracker.register(navController)
+    }
+
+    internal fun setTrackerConfig(detector: ScreenChangeDetector, processor: NavigationEventProcessor?) {
+        pendingDetector = detector
+        pendingProcessor = processor
+    }
+
+    internal fun clearTrackerConfig() {
+        pendingDetector = null
+        pendingProcessor = null
+    }
+
+    private fun createTrackerIfConfigured(): ComposeNavigationTracker? {
+        val detector = pendingDetector ?: return null
+        val tracker = ComposeNavigationTracker(
+            screenChangeDetector = detector,
+            processor = pendingProcessor
+        )
+        composeTracker = tracker
+        pendingDetector = null
+        pendingProcessor = null
+        Logger.d(TAG, "ComposeNavigationTracker initialized")
+        return tracker
     }
 
     /**
--- a/integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
+++ b/integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
@@ -67,7 +67,10 @@ class Navigation internal constructor() {
     private fun createTrackerIfConfigured(): ComposeNavigationTracker? {
         val detector = pendingDetector ?: return null
         if (!isRouteApiAvailable) {
-            Logger.d(TAG, "NavDestination.getRoute() not available (requires navigation 2.4.0+), Compose tracking disabled")
+            Logger.d(
+                TAG,
+                "NavDestination.getRoute() not available (requires navigation 2.4.0+), Compose tracking disabled"
+            )
             pendingDetector = null
             pendingProcessor = null
             return null
--- a/integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
+++ b/integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
@@ -66,15 +66,6 @@ class Navigation internal constructor() {
 
     private fun createTrackerIfConfigured(): ComposeNavigationTracker? {
         val detector = pendingDetector ?: return null
-        if (!isRouteApiAvailable) {
-            Logger.d(
-                TAG,
-                "NavDestination.getRoute() not available (requires navigation 2.4.0+), Compose tracking disabled"
-            )
-            pendingDetector = null
-            pendingProcessor = null
-            return null
-        }
         val tracker = ComposeNavigationTracker(
             screenChangeDetector = detector,
             processor = pendingProcessor
@@ -84,15 +75,6 @@ class Navigation internal constructor() {
         pendingProcessor = null
         Logger.d(TAG, "ComposeNavigationTracker initialized")
         return tracker
-    }
-
-    private val isRouteApiAvailable: Boolean by lazy {
-        try {
-            Class.forName("androidx.navigation.NavDestination").getMethod("getRoute")
-            true
-        } catch (_: Exception) {
-            false
-        }
     }
 
     /**
--- a/integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
+++ b/integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
@@ -71,8 +71,7 @@ class Navigation internal constructor() {
             processor = pendingProcessor
         )
         composeTracker = tracker
-        pendingDetector = null
-        pendingProcessor = null
+        clearTrackerConfig()
         Logger.d(TAG, "ComposeNavigationTracker initialized")
         return tracker
     }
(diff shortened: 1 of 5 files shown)


integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
```diff
@@ -44,14 +47,54 @@ class Navigation internal constructor() {
...
+        composeTracker = tracker
+        pendingDetector = null
+        pendingProcessor = null
+        Logger.d(TAG, "ComposeNavigationTracker initialized")
+        return tracker
+    }
+
+    private val isRouteApiAvailable: Boolean by lazy {
+        try {
+            Class.forName("androidx.navigation.NavDestination").getMethod("getRoute")
```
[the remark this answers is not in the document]
tonzhan2: it was removed in a commit that wasn't yet pushed

integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
```diff
@@ -44,14 +47,36 @@ class Navigation internal constructor() {
...
+    }
+
+    private fun createTrackerIfConfigured(): ComposeNavigationTracker? {
+        val detector = pendingDetector ?: return null
+        val tracker = ComposeNavigationTracker(
+            screenChangeDetector = detector,
+            processor = pendingProcessor
+        )
+        composeTracker = tracker
+        pendingDetector = null
```
surbhiia: nit: call `clearTrackerConfig()` method here instead.

integration/navigation/src/test/kotlin/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTrackerTest.kt
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Test
+    fun `LinkageError on route access disables tracking without crash`() {
+        val navController = mock(NavController::class.java)
+        tracker.register(navController)
+
+        val destination = mock(NavDestination::class.java)
+        `when`(destination.navigatorName).thenReturn("composable")
+        `when`(destination.route).thenThrow(NoSuchMethodError("getRoute"))
```
aditi-s3, line 105: I am not sure how valuable this test is, we are throwing a deliberate NoSuchMethodError from a runtime where getRoute() exists.

integration/navigation/src/test/kotlin/com/splunk/rum/integration/navigation/NavigationTest.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    @Test
+    fun `pending config is cleared after tracker creation`() {
+        navigation.setTrackerConfig(detector, null)
+        navigation.registerNavController(navController)
+
+        val secondNavigation = Navigation()
+        secondNavigation.composeTracker = null
+
+        navigation.clearTrackerConfig()
```
aditi-s3: Should we clearTrackerConfig here ? I am confused what this test is doing.

integration/navigation/src/test/kotlin/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTrackerTest.kt
```diff
@@ -0,0 +1,154 @@
...
+    private fun createDestination(route: String): NavDestination {
+        val destination = mock(NavDestination::class.java)
+        `when`(destination.navigatorName).thenReturn("composable")
+        `when`(destination.route).thenReturn(route)
+        `when`(destination.parent).thenReturn(null)
+        return destination
+    }
+
+    private fun simulateDestinationChanged(navController: NavController, destination: NavDestination) {
+        val listenerField = tracker.javaClass.getDeclaredField("activeListener")
```
aditi-s3: SHouln't we mockito capture the listener passed onto NavController.addOnDestinationChangedListener()  ?

integration/navigation/src/test/kotlin/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTrackerTest.kt
tonzhan2: We can't actually test against a pre-2.4.0 navigation library in unit tests, so Mockito is the only way to exercise the catch path. 

Removed the reflection approach and now calling handleDestinationChanged directly (made it @VisibleForTesting internal). Added a KDoc comment

integration/navigation/src/test/kotlin/com/splunk/rum/integration/navigation/NavigationTest.kt
tonzhan2: Removed this one, the other tests cover the behavior

integration/navigation/src/test/kotlin/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTrackerTest.kt
tonzhan2: ArgumentCaptor and any() both have Kotlin null safety interop issues with NavControllers nonnull parameters. 

Switched to calling handleDestinationChanged directly instead, which is cleaner anyway.

integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTracker.kt
```diff
@@ -72,14 +73,31 @@ internal class ComposeNavigationTracker(
...
     }
 
-    private fun handleDestinationChanged(destination: NavDestination, arguments: Bundle?) {
+    @VisibleForTesting
+    internal var routeApiAvailable = true
+        private set
+
+    @VisibleForTesting
+    internal fun handleDestinationChanged(destination: NavDestination, arguments: Bundle?) {
+        if (!routeApiAvailable) return
```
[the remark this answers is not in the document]
tonzhan2: This would never happen. Customer would not call registerNavController if they are using pre 2.4.0 navigation and run into LinkageError. Added a defensive check anyways though

integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/Navigation.kt
```diff
@@ -44,14 +47,35 @@ class Navigation internal constructor() {
...
+        val tracker = composeTracker ?: createTrackerIfConfigured() ?: run {
             Logger.w(TAG, "Navigation module not installed. Cannot register NavController.")
             return
         }
         tracker.register(navController)
     }
 
+    internal fun setTrackerConfig(detector: ScreenChangeDetector, processor: NavigationEventProcessor?) {
+        pendingDetector = detector
+        pendingProcessor = processor
```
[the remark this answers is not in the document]
tonzhan2: SplunkRum.install() runs on the main thread during Application.onCreate(), and registerNavController() runs on the main thread during Activity creation. 

Both calls are on the same thread, so the writes in setTrackerConfig() are fully visible before registerNavController() can execute. There can be no concurrent access.

integration/navigation/src/main/kotlin/com/splunk/rum/integration/navigation/automatic/ComposeNavigationTracker.kt
```diff
@@ -43,16 +44,17 @@ internal class ComposeNavigationTracker(
...
 
         clearRegistration()
 
         val listener = NavController.OnDestinationChangedListener { _, destination, arguments ->
             handleDestinationChanged(destination, arguments)
         }
-        navController.addOnDestinationChangedListener(listener)
         registeredController = WeakReference(navController)
         activeListener = listener
+        navController.addOnDestinationChangedListener(listener)
```
surbhiia, line 57: Should following be added after this line:

```
//  The current-destination callback is invoked before adding the
// listener internally. Remove it again after registration if that callback
// detected the missing route API.
if (!routeApiAvailable) {
    navController.removeOnDestinationChangedListener(listener)
    return
}
```

Reason:
On the first registration, routeApiAvailable is still true, the sequence is:

- register() sees routeApiAvailable == true.
- It calls addOnDestinationChangedListener(listener).
- Navigation immediately invokes the listener before storing it.
- destination.route throws NoSuchMethodError.
- The catch sets routeApiAvailable = false.
- clearRegistration() calls removeOnDestinationChangedListener(), but the controller has not stored the listener yet, so removal does nothing.
- The callback returns.
- Navigation then stores the listener.

The listener is therefore left attached. The opening guard (`if (!routeApiAvailable) return`) cannot help because it already ran before the failure occurred.

The suggested post-registration check handles that timing.

tonzhan2: Good call. Just added a post add check. After addOnDestinationChangedListener returns, if the synchronous callback already tripped routeApiAvailable = false, we immediately remove the listener and clear references. Handles the case regardless of whether NavController stores before or after dispatching.

surbhiia (approved):

PR #1681, opened 2026-08-18
DEMRUM-6380: Update server timing header regex to accept any two-digit hexadecimal flag value

surbhiia:
### Description

The Android RUM SDK’s Server-Timing parser requires traceparent values to end in 01. Newer Java and Python agents emit valid values ending in 03, causing the header to be rejected and breaking RUM-to-backend trace correlation.
Update the parser to accept any valid two-digit hexadecimal flags value, since Android only extracts the trace and span IDs and does not interpret the flags.

### Checklist

### Generative AI usage

### How to Test These Changes

- 01, 03, and other valid two-digit hex flags are accepted.
- Malformed flags remain rejected.
- Both OkHttp and HttpURLConnection are covered.

### DIFF
--- a/common/utils/src/main/kotlin/com/splunk/rum/agent/common/utils/ServerTimingHeaderParser.kt
+++ b/common/utils/src/main/kotlin/com/splunk/rum/agent/common/utils/ServerTimingHeaderParser.kt
@@ -29,14 +29,14 @@ data class ServerTraceContext(val traceId: String, val spanId: String)
 object ServerTimingHeaderParser {
 
     private val HEADER_PATTERN: Pattern = Pattern.compile(
-        """traceparent;desc=['"]00-([0-9a-f]{32})-([0-9a-f]{16})-01['"]"""
+        """traceparent;desc=['"]00-([0-9a-f]{32})-([0-9a-f]{16})-[0-9a-f]{2}['"]"""
     )
 
     /**
      * Parses the server-timing header to extract the trace ID and span ID.
      *
      * @param header A header string in the form:
-     *     traceparent;desc="00-9499195c502eb217c448a68bfe0f967c-fe16eca542cd5d86-01"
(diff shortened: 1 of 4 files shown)


common/utils/src/main/kotlin/com/splunk/rum/agent/common/utils/ServerTimingHeaderParser.kt
```diff
@@ -29,14 +29,14 @@ data class ServerTraceContext(val traceId: String, val spanId: String)
 object ServerTimingHeaderParser {
 
     private val HEADER_PATTERN: Pattern = Pattern.compile(
-        """traceparent;desc=['"]00-([0-9a-f]{32})-([0-9a-f]{16})-01['"]"""
+        """traceparent;desc=['"]00-([0-9a-f]{32})-([0-9a-f]{16})-[0-9a-f]{2}['"]"""
     )
```
scheler, line 33: A more tighter regex than {0-9a-f]{2} may be good since it is just 4 possibilities ["00", "01", "02", "03"]

surbhiia: Chose this because - W3C defines trace-flags structurally as an 8-bit field encoded using two lowercase hexadecimal digits. This parser only extracts the trace ID and span ID; it does not interpret the flags. 
Accepting the complete two-digit field keeps it compatible with future flags. Restricting it to 0[0-3] could recreate the same issue we have now when another reserved bit is assigned.

tonzhan2 (approved):

common/utils/src/main/kotlin/com/splunk/rum/agent/common/utils/ServerTimingHeaderParser.kt
scheler: Ok, sounds good.

